PR #20, opened 2019-01-29
Implementing stats.Min, stats.Max, stats.PeakToPeak and preprocessing.MinMaxScaler()

AdilZouitine:


creme/stats.py
```diff
@@ -130,3 +131,79 @@ def update(self, x):
...
+        max (stats.Max) 
+        min (stats.Min)
+        p2p (float): The running peak to peak.
+
+    """
+    def __init__(self):
+        super().__init__()
+        self.max = Max()
+        self.min = Min()
+        self.p2p = 0
```
MaxHalford: You don't have to store the `p2p` value. You can simply do the `max.get() - min.get()` computation in the `get` method.

creme/stats.py
```diff
@@ -130,3 +131,79 @@ def update(self, x):
...
+        return 'p2p'
+
+    def update(self,x):
+        new_max = self.max.update(x).get()
+        new_min = self.min.update(x).get()
+        self.p2p = new_max - new_min
+        return self 
+    
+    def get(self):
+        return self.p2p
```
MaxHalford: You can do `max.get() - min.get()` here.

creme/preprocessing/scale.py
```diff
@@ -84,3 +84,81 @@ def transform_one(self, x):
...
+               [1.        ],
+               [0.85519484],
+               [0.20199041],
+               [0.16984743]])
+
+    """
+    
+    def __init__(self):
+        self.min = collections.defaultdict(lambda : stats.Min())
+        self.p2p = collections.defaultdict(lambda : stats.PeakToPeak())
```
MaxHalford: I think you should use a `max`. The problem is that by using a `p2p` you're computing the `min` twice.

creme/preprocessing/scale.py
```diff
@@ -84,3 +84,81 @@ def transform_one(self, x):
...
+    def fit_one(self, x, y=None):
+
+        for i, xi in x.items():
+            self.min[i].update(xi)
+            self.p2p[i].update(xi)
+        return self.transform_one(x)
+
+    def transform_one(self, x):
+        return {
+            i: (xi - self.min[i].get()) / (self.p2p[i].get() + self.eps)
```
MaxHalford: If you use a `max` instead of a `p2p` you can divide by `self.max[i] - self.min[i] + self.eps` instead.

PR #25, opened 2019-01-31
Implementation L1 Loss and stats.Sum

AdilZouitine:


creme/stats.py
```diff
@@ -205,4 +205,26 @@ def update(self,x):
...
+    """
+    def __init__(self):
+        super().__init__()
+        self.sum = 0.0
+    
+    @property
+    def name(self):
+        return 'sum'
+    
+    def update(self,x):
```
MaxHalford: You just missed a space after the comma here :). Apart from that everything seems fine.

PR #28, opened 2019-02-01
Implementation optim.HingeLoss

AdilZouitine:


MaxHalford:
Looks good, but can you try and add some doctests? Maybe you can reproduce the examples from sklearn: [https://github.com/scikit-learn/scikit-learn/blob/7389dba/sklearn/metrics/classification.py#L1822][https://github.com/scikit-learn/scikit-learn/blob/7389dba/sklearn/metrics/classification.py#L1822].

creme/optim/losses.py
```diff
@@ -79,3 +79,32 @@ def __call__(self, y_true, y_pred):
...
+
+    """
+
+    def __call__(self, y_true, y_pred):
+        return max(0, 1 - y_pred * y_true)
+
+    def gradient(self, y_true, y_pred):
+        if y_pred * y_true < 1:
+            return - y_pred * y_true
+        else:
```
MaxHalford: You can remove this `else` and simply put the `return 0` instead.

creme/optim/losses.py
```diff
@@ -106,5 +130,4 @@ def __call__(self, y_true, y_pred):
     def gradient(self, y_true, y_pred):
         if y_pred * y_true < 1:
             return - y_pred * y_true
```
MaxHalford: I'm really not sure but isn't the gradient of the Hinge loss `-y`? I just looked at [https://stats.stackexchange.com/questions/4608/gradient-of-hinge-loss](https://stats.stackexchange.com/questions/4608/gradient-of-hinge-loss).

PR #29, opened 2019-02-01
Add CentralMoments, Kurtosis and Skew classes - new features

raphaelsty:
Test passed successfully. 

Kurtosis and Skew are the same as spicy version.

```
scipy.stats.skew()
scipy.stats.kurtosis()
```


MaxHalford:
Beautiful!

PR #55, opened 2019-03-27
New feature: standard error of the mean, incremental and rolling incremental sem

raphaelsty:
These two statistics have been validated. The results are the same as those of https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sem.html

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -367,6 +367,7 @@ Loss functions
    stats.NUnique
    stats.PeakToPeak
    stats.Quantile
+   stats.Sem
    stats.Skew
    stats.Sum
    stats.RollingMax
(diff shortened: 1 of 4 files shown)


MaxHalford:
Very nice! Can you just replace `Sem` by `SEM` in the class names? In Python the convention is to capitalize acronyms.

docs/api.rst
```diff
@@ -375,6 +376,7 @@ Loss functions
    stats.RollingMode
    stats.RollingPeakToPeak
    stats.RollingQuantile
+   RollingSem
```
MaxHalford: This is missing `stats.`

PR #92, opened 2019-05-16
Added AdaBound optimizer

tweakyllama:
Hi, this is my first pull request so I'm sorry if I've done something wrong, and apologies if I've made a pull request to the wrong branch, but I couldn't see the dev branch of the base repo.

Ive done my  best to implement the AdaBound optimizer as per [this](https://github.com/Luolc/AdaBound), but I wasn't sure how to do the learning rate decay. I noticed the `lr_schedule` classes in the optim module, so maybe the best way would be to use one of those? 

When running the code in the docstring the output is exactly the same as when running the code in the Adam optimizer, I think this is because I haven't actually implemented the learning rate decay and the paper mentions that without this the AdaBound optimizer behaves the same as the Adam optimizer.


MaxHalford:
Hey there,

Thanks for your PR. You're the first person outside of our group of friends to contribute, so cheers. I'm going on holiday in a few tomorrow but I will be back Tuesday. I'm not sure about the details but we will have to see if we implement the decay inside the optimizer or as a new learning rate scheduler. It all depends on what is the contribution of AdaBound. If it's just a new decay rate then we may as well code something generic which can be used for any optimizer.

I'll do a review when I get back, and assign you to the issue on GitHub.

tweakyllama:
No worries, I really like the look of this project, and am happy to help and contribute where I can. 

Sounds good.

MaxHalford:
So indeed it seems to me that the only thing you're missing is the learning rate decay. In the AdaBound repository this seems to be implemented as so:

```python
final_lr = group['final_lr'] * group['lr'] / base_lr
```

while you're just doing:

```python
final_lr = group['final_lr']
```

Correct? The thing is I'm not sure what each `lr` corresponds to... I guess `group['lr']` is just another parameter (as if there wasn't enough!) and `base_lr` is the current learning as specified by the learning rate scheduler (in our case this is `self.learning_rate`).

By the way you can use the `utils.clamp` function to clamp the learning rate instead of writing `max(lower_bound, min(step_size_bound, upper_bound))`. I know this seems silly but the clamping part is the main point of the algorithm and my feeling is that this should be made clear. But it isn't too important, I'm just being annoying.

MaxHalford:
Just a test.

@all-contributors please add @tweakyllama

tweakyllama:
I didn't see the `utils.clamp`, I will change that, makes it look neater. 

Yeah I got a bit confused by the three different learning rates. As far as I can tell, `lr` is the Adam learning rate, `base_lr` is initialised to this, and `final_lr` is the SGD final learning rate that we decay to. What I can't quite figure out is where `group['lr']` gets updated, or to what.

MaxHalford:
I believe that `group['lr']` is handled outside of the optimizer. It's value depends on the current iteration. In `creme` this is a `LRScheduler`. Using `self.learning_rate` will do the trick I believe. Take a look at the `Optimizer` class in `optim/base.py` if you haven't already :)

tweakyllama:
Looking at the base `Optimizer` class, it defaults to using a `ConstantLR` schedule, so in that case would we need to track our own value of `final_lr` in a property within the class, or use a new `lr_schedule`. Otherwise we'd just be using the same constant learning rate.

Or am I missing something here?

MaxHalford:
I believe you need to track your own `final_lr`.

tweakyllama:
Ok it properly tracks the `final_lr` now, though the value doesn't actually change when you run it, it now slightly outperforms the `Adam` optimiser, so I guess something went right.

MaxHalford:
Looks good to me. In the future I really want to add proper unit tests for each optimizer. Validating by looking at performance isn't enough. It would be interesting to see what Tensorflow and PyTorch do to validate their implementations. Thanks for your work :)

PR #96, opened 2019-05-22
Exponential Smoothing

koaning:
I'm adding this here now because the `dev` branch is missing? 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,7 @@ dmypy.json
 
 # IDE
 .vscode
+.idea
 
 # For convenience
 Untitled*.ipynb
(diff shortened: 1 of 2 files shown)


koaning:
Currently the only check I have in terms of testing is in the docstring. If more is needed I'll gladly hear it. 

Setting up the dev environment was a bit tricky due to packages, I've described the issue here: https://github.com/creme-ml/creme/issues/95 but the build looks green.

MaxHalford:
Really nice @koaning, kudos for going the extra mile. At some point we used a `dev` branch but it didn't feel convenient, I updated the `CONTRIBUTING.md` file accordingly.

PR #114, opened 2019-05-28
fix type hinting

flegac:


MaxHalford:
Thanks for the PR!

Indeed the return types are wrong for `Transformer` and `Clusterer`, they should be `'Transformer'` and `'Clusterer'`, respectively. Can you update your PR?

flegac:
:heavy_check_mark:

PR #128, opened 2019-07-02
fix error with utils.math.dot

Adam-Hoelscher:
fix dot product errors as demonstrated here: https://github.com/creme-ml/creme/issues/127

### DIFF
--- a/creme/utils/math.py
+++ b/creme/utils/math.py
@@ -38,7 +38,9 @@ def dot(x: dict, y: dict):
         42
 
     """
-    return sum(xi * y.get(i, 0) for i, xi in min(x, y, key=len).items())
+    l = min(x, y, key=len)
+    r = y if l is x else x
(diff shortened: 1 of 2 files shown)


creme/utils/math.py
```diff
@@ -38,7 +38,9 @@ def dot(x: dict, y: dict):
         42
 
     """
-    return sum(xi * y.get(i, 0) for i, xi in min(x, y, key=len).items())
+    l = min(x, y, key=len)
+    r = y if l is x else x
```
MaxHalford: I don't really like `if l is x` because it has to compare `dict`s. Can you instead do this:

```python
if len(l) < len(r):
    l, r = x, y
else:
    l, r = y, x
```

Or something like that?

MaxHalford:
Hey @Adam-Hoelscher, I left a comment :)

MaxHalford:
I fixed this myself as @Adam-Hoelscher didn't answer.

PR #144, opened 2019-09-04
Add IQR and RollingIQR

AdilZouitine:
I have implemented IQR and RollingIQR, don't hesitate to tell me the modifications to make.

### DIFF
--- a/creme/stats/iqr.py
+++ b/creme/stats/iqr.py
@@ -16,3 +16,22 @@ class IQR(base.Univariate):
 
     def get(self):
         return self.quantile_sup.get() - self.quantile_inf.get()
+
+
+class RollingIQR(base.RollingUnivariate):
+    
+    def __init__(self, window_size, q_inf=0.25, q_sup=0.75):
+        if q_inf >= q_sup:
+            raise ValueError('q_inf must be strictly less than q_sup')
+        self.quantile_inf = quantile.RollingQuantile(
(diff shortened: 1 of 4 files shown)


AdilZouitine:
I add some fix but I'm not sure about the `RollingIQR` inheritance and the` __str__` method uses `self.size` instead of `self.window_size`

creme/stats/iqr.py
```diff
@@ -34,13 +36,19 @@ class IQR(base.Univariate):
...
             raise ValueError('q_inf must be strictly less than q_sup')
-        self.quantile_inf = quantile.Quantile(quantile=q_inf)
-        self.quantile_sup = quantile.Quantile(quantile=q_sup)
+        self.q_inf = q_inf
+        self.q_sup = q_sup
+        self.quantile_inf = quantile.Quantile(quantile=self.q_inf)
+        self.quantile_sup = quantile.Quantile(quantile=self.q_sup)
+
+    def __str__(self):
+        return f'rolling_{self.__class__.__name__}_{self.q_inf}_{self.q_sup}'
```
MaxHalford: I think you're inversed the `__str__` method of both classes. Also, you type `iqr` instead of `self.__class__.__name__`.

MaxHalford: I'll merge the PR and take care of this, it's no biggie :)

PR #155, opened 2019-10-04
Adding iter_vaex, to stream data from vaex.dataframe

JovanVeljanoski:
Hello! 

First of all, great work on this project. It has a lot of potential! I really hope it will continue to grow and improve. 

This is just a small PR which adds a function to iterate over `vaex` DataFrames. 
Here is a minimal working example (screen-shot)
image

[Vaex](https://github.com/vaexio/vaex) is a DataFrame library (ala) `pandas`, but specialized for working with much larger datasets that do not fit onto the memory of a single computer. In a way, it shares similar concepts to `creme`, i.e. computing statistics in chunks. 

I can see these libraries working well together. I agree that for taking models to production, monitoring etc.. online learning makes a lot of sense, and is something I am considering for my upcoming models. But sometimes one does have plenty of data to experiment and train on, create features etc, before actually building the model and considering production. Synergy between these two libraries may solve both problems: exploration and analysis of "big" data locally, and creating a model and taking it to production, and not really needed to retrain it as time goes on.

### DIFF
--- a/creme/stream.py
+++ b/creme/stream.py
@@ -17,7 +17,11 @@ try:
 except ImportError:
     PANDAS_INSTALLED = False
 from sklearn import utils
-
+try:
+    from vaex.utils import _ensure_strings_from_expressions, _ensure_list
+    VAEX_INSTALLED = True
+except ImportError:
+    VAEX_INSTALLED = False
 
 __all__ = [
     'iter_csv',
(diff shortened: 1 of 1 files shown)


creme/stream.py
```diff
@@ -112,6 +116,42 @@ def iter_pandas(X, y=None, **kwargs):
...
+        X (vaex.DataFrame): A vaex DataFrame housing the training featuers.
+        y (string or vaex.Expression): The column or expression containing the target variable.
+        features (list of strings or vaex.Expressions): A list of features used for training.
+        If None, all columns in ``X`` will be used. Features specifying in ``y`` are ignored.
+
+    Yields:
+        tuple: A pair (``x``, ``y``) where ``x`` is a dict of features and ``y`` is the target.
+    """
+
+    if VAEX_INSTALLED is not True:
```
MaxHalford: There's many ways to do this, and indeed we used this same kind of check for `pandas`. What I would prefer here is to place the `from vaex.utils import _ensure_strings_from_expressions, _ensure_list` import inside the `iter_vaex` function. This way an `ImportError` will be raised if `vaex` isn't installed.

JovanVeljanoski: Hey! 
Sure, what you say makes sense, in fact that was my 1st version.. but than my local linter started to complain..  :) I will make adjustments per your suggestion.

JovanVeljanoski:
Yeah, few example notebooks would be great, even perhaps a couple of articles or blog posts (medium-like?). We have been working on analysis on "big" (100s of GB size) of datasets problems, including EDA and pre-processing, but we are stuck on the modelling part. Your library is a very good solution to this, and particle for the real world as well! I am happy that I came across your talk from PyData Amsterdam. Maybe if you have time you can [check out ours](https://www.youtube.com/watch?v=2Tt0i823-ec&t=833s), just to have a better sense of what we are trying to accomplish. But indeed I see a lot of possible synergies. Cheers.

MaxHalford:
I will check out your talk tomorrow during the day. As for a Medium post, we are waiting till we have a stable version of an online decision tree. Then we were thinking about posting an article on Towards Data Science. Cheers.

Issue #171, opened 2019-11-04
Check gradients with finite differences


MaxHalford:
I was reading [this](https://cilvr.cs.nyu.edu/diglib/lsml/bottou-sgd-tricks-2012.pdf) wonderful paper and in section 4.2 there is a paragraph about checking gradients via finite differences. I think this would be pretty nice to have as a unit test.

PR #172, opened 2019-11-05
Check gradients with finite differences - Work in Progress

raphaelsty:
Hi,

This draft pull request for issue #171, 📊

I used the parameter delta as a threshold for finite differences with slight updates in the weights.

I don't know yet if it's a smart choice. I need to analyze variations between finite differences for each optimizer and various datasets to select a smarter threshold.

Still in progress but your comments are welcome.

Raphaël


MaxHalford:
This is really good @raphaelsty! To push the envelope on this, I think it would be nice if tried it with different delta pertubations as detailed by Léon Bottou. Also, maybe we could add to this PR the possibility to initialize the weights of `LinearRegression` and `LogisticRegression`, following the work on weight initialization from @VaysseRobin.

creme/test_.py
```diff
@@ -130,3 +136,58 @@ def is_estimator(obj):
...
+    for x, y in X_y:
+        x = scaler.fit_one(x).transform_one(x)
+        utils.estimator_checks.check_gradient_finite_difference(model, x, y, delta = 0.001)
+        model.fit_one(x, y)
+        
+    # Test gradient finite difference with logistic regression
+    X_y = creme.datasets.fetch_electricity()
+    scaler = preprocessing.StandardScaler()
+    model = linear_model.LogisticRegression()
+    model.weights = initializer(shape = len(model.weights))
```
MaxHalford: This is not what I meant. I was thinking of replacing `self.weights = collections.defaultdict(float)` by `self.weights = collections.defaultdict(initializer)` in the `GLM` class. `initializer` would thus be a parameter specified in the `GLM` signature.

raphaelsty: @MaxHalford,

I updated GLM class, LinearRegression and LogisticRegression classes by adding a new parameter : initializer. This parameter is defined as optim.initializers.Zeros() by default in LinearRegression and LogisticRegression classes too. 

I updated the initializers by setting the default shape to 1 when it is called. It's easier for linear models and it makes sense to me. (Easier because of the defaultdict scheme with weights).

Did I understand what you wanted for the GLM class, the intializers and the gradient test? 

Thank you for your review.

MaxHalford: Yes it looks good! Are the unit tests passing?

raphaelsty:
That sounds good to me.✌️

Issue #76, opened 2019-04-19
L1 regularization for linear models


MaxHalford:
Shouldn't be too difficult to do but I'm not sure how exactly to make it work with L2 regularization at the same time. L1 + L2 is called ElasticNet but I'm not sure of the details.

References:

- [Stochastic Gradient Descent Training forL1-regularized Log-linear Models with Cumulative Penalty](https://www.aclweb.org/anthology/P09-1054)
- [Sparse Online Learning via Truncated Gradient](http://hunch.net/~jl/projects/interactive/sparse_online/paper_sparseonline.pdf)

gbolmier:
One more reference: [Efficient Learning using Forward-Backward Splitting](https://papers.nips.cc/paper/3793-efficient-learning-using-forward-backward-splitting)

gbolmier:
Ran some [experiments](https://nbviewer.jupyter.org/github/gbolmier/workbench/blob/master/induce-sparsity-in-online-learning-linear-models.ipynb) implementing methods proposed here:

> * [Stochastic Gradient Descent Training forL1-regularized Log-linear Models with Cumulative Penalty](https://www.aclweb.org/anthology/P09-1054)
> * [Sparse Online Learning via Truncated Gradient](http://hunch.net/~jl/projects/interactive/sparse_online/paper_sparseonline.pdf)

We have pretty good results on a synthetic dataset of 10 informative features, and 90 noisy features:

| Method                | Hyper-parmameters                     | # of features | MAE |
|:----------------------|:--------------------------------------|--------------:|:---:|
| No regularization     |                                       |           100 | 8.2 |
| L1 naive              | λ = 0.1                       |           100 | 8,7 |
| L1 cumulative         | λ = 0.1                       |            11 | 6.7 |
| Truncated gradient    | λ = 5 &nbsp; K=10 &nbsp; θ = 3 |             9 | 6.8 |

![download](https://user-images.githubusercontent.com/25319692/89712714-e2b22600-d992-11ea-8c33-4b2e2af4248c.png)

MaxHalford:
That looks really good Geoff! Let's wait till we've moved forward with the scikit-multiflow merge. The cost can probably be alleviated by writing some of it in Cython.

PR #187, opened 2019-11-27
Add FM and FFM models

gbolmier:


gbolmier:
`test_check_estimator` failed with `FFMClassifier` and `FFMRegressor` because of numeric feature names that are incompatible with field inference: `feature_name.split('_')[0]` leading to `AttributeError: 'int' object has no attribute 'split'`.
Should we just handle this case with `str(feature_name).split('_')[0]` or modify the test?

creme/linear_model/ffm.py
```diff
@@ -0,0 +1,364 @@
...
+        self.optimizer = optim.SGD(0.01) if optimizer is None else optimizer
+        self.intercept = intercept
+
+        self.intercept_lr = (
+            optim.schedulers.Constant(intercept_lr)
+            if isinstance(intercept_lr, numbers.Number) else
+            intercept_lr
+        )
+
+        if weight_initializer:
```
MaxHalford: A cleaner way here would be:

```python
if weight_initializer is None:
    weight_initializer = optim.initializers.Zeros()
self.weight_initializer = weight_initializer
```

But I'm probably nitpicking :)

creme/linear_model/ffm.py
```diff
@@ -0,0 +1,364 @@
...
+        )
+
+        field_latents_dict = functools.partial(
+            collections.defaultdict, random_latents
+        )
+
+        self.weights = collections.defaultdict(self.weight_initializer)
+        self.latents = collections.defaultdict(field_latents_dict)
+
+    def _field(self, j):
```
MaxHalford: I've thought about it a bit and I don't think I like this approach. In this case I'm assuming that you're expecting dicts like this:

```python
x = {
    'user_Pamela': 1,
    'item_Television': 1,
}
```

I guess this approach is good because it add flexibility. I would however prefer this way:

```python
x = {
    'user': 'Pamela',
    'item': 'Television',
    'hour': 16
}
```

By the way how would you handle extra features such as `'hour'`?

creme/linear_model/ffm.py
```diff
@@ -0,0 +1,364 @@
...
+        g_loss *= sample_weight
+
+        # Update the intercept
+        sign = lambda x: -1 if x < 0 else (1 if x > 0 else 0)
+
+        self.intercept -= w0_lr * (g_loss + l1 * sign(w0) + 2. * l2 * w0)
+
+        # Update the weights
+        gradient = {
+            j: g_loss * xj + l1 * sign(w[j]) + 2. * l2 * w[j]
```
MaxHalford: Do you have a reference for the L1 regularization method you're using? As we discussed it would preferable if we extracted the L1/L2 logic into separate classes. But I haven't thought it through, it might not be possible!

MaxHalford:
The code looks really good @gbolmier, you did a great job :). I just think we should talk about what the inputs look like. Before merging this, I would really like to have a comparison with at least one other established library. This could be added to the benchmarks directory. FM and FFM are really important models and I think it's important that we trust our models to be as accurate as other implementations, regardless of the speed differences.

creme/linear_model/ffm.py
gbolmier: Nice catch!

creme/linear_model/ffm.py
gbolmier: Working on it as discussed earlier, keep you in touch!

gbolmier: After taking a look at it, here is what I propose:

We can define the following method:
```python
def _transform(self, x):
        return dict((f'{k}_{v}', 1) if isinstance(v, str) else (k, v) for k, v in x.items())
```
And apply it at the beginning of `fit_one` and `predict_one`, after running an experiment it's actually faster (39 sec against 44 on movielens 100k) than using `creme.processing.one_hot_encoder`. The only caveat for the end user is to be careful to set such variables to strings (e.g. when using ids).

No problem with extra features as `hour`:
```python
def _field(self, j):
    return j.split('_')[0]

_field('hour')
>>> 'hour'
```

MaxHalford: That's nice to hear. It's nice that you have a benchmark to measure the speed impact! As for the strings, I think it's fine. The API is now really intuitive!

gbolmier: Great :)

creme/linear_model/ffm.py
gbolmier: As discussed earlier today, I'm just using the basic and naive method as a first step. Unfortunately it's not efficient in an online setting and results with a lot of non 0 features.

I would point to you your reference here: [L1 regularization for linear models #76](https://github.com/creme-ml/creme/issues/76).The first paper pointed explains well our problem and provide clues for improvement.

You are right about the regularization logic, that's something we need to think about. Let's do it as a next step

gbolmier:
Hey! Should be fine for the inputs regarding what we want now. Currently working on two notebooks:
- A Benchmark with `tffm` and `xLearn`, other ideas to compare with?
- **Factorization models for Recommender Systems**, a tutorial for the user guide (`SVD`, `FM` and `FFM`)

MaxHalford:
Good stuff!

creme/base.py
```diff
@@ -459,6 +459,10 @@ class Wrapper(abc.ABC):
     def _model(self):
         """Provides access to the wrapped model."""
 
+    @abc.abstractproperty
+    def _labelloc(self):
```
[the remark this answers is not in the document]
gbolmier: This attribute indicates if the wrapping utility should be drawn at the top or bottom of the wrapped model

creme/linear_model/ffm.py
```diff
@@ -0,0 +1,378 @@
...
+
+class FFM:
+    """Field-aware Factorization Machines.
+
+    Parameters:
+        n_factors (int): Dimensionality of the factorization or number of latent factors.
+        loss (optim.Loss): The loss function to optimize for.
+        optimizer (optim.Optimizer): The sequential optimizer used for updating the weights and
+            latent factors. Note that the intercept is handled separately.
+        intercept (float or `creme.stats.Univariate` instance): Initial intercept value.
```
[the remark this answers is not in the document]
gbolmier: If the user want to set the global mean as intercept for instance (like a biased matrix factorization). It just add an if statement in the code and doesn't slow down the algorithm

creme/linear_model/ffm.py
```diff
@@ -0,0 +1,378 @@
...
+
+        field_latents_dict = functools.partial(
+            collections.defaultdict, random_latents
+        )
+
+        self.weights = collections.defaultdict(self.weight_initializer)
+        self.latents = collections.defaultdict(field_latents_dict)
+
+    def _transform(self, x):
+        return dict((f'{j}_{xj}', 1) if isinstance(xj, str) else (j, xj) for j, xj in x.items())
```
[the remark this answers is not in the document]
gbolmier: Didn't know it was possible, sounds cool for me :)

creme/metrics/base.py
```diff
@@ -132,7 +134,7 @@ def bigger_is_better(self):
         return False
 
     def works_with(self, model) -> bool:
-        return isinstance(utils.estimator_checks.guess_model(model), base.Regressor)
+        return isinstance(utils.estimator_checks.guess_model(model), (base.Regressor, Recommender))
```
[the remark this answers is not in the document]
gbolmier: Indeed would be smarter!

creme/linear_model/fm.py
```diff
@@ -1,126 +1,386 @@
...
+
+            >>> model.predict_one({'Bob': 1, 'Harry Potter': 1})
+            5.320369...
+
+    Note:
+        Using a feature scaler such as `preprocessing.StandardScaler` on non-binary features helps
+        the optimizer to converge.
+
+    References:
+            1. `Factorization Machines <https://www.csie.ntu.edu.tw/~b97053/paper/Rendle2010FM.pdf>`_
```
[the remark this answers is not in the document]
gbolmier: Nice catch!

creme/base.py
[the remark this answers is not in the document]
MaxHalford: Oki doki. This feels a bit akward, and ideally it should be handled by the Graphviz rendering algorithm. For the moment let's leave it like this and we'll have another look later on.

creme/linear_model/ffm.py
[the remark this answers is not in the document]
gbolmier: At the end it uses less memory to format it as strings (one lighter object), for instance it reduces model memory usage from 1.1 MiB to 950 KiB on MovieLens 100K. That's something we may consider for large scale situations.

Any reasons to use tuples instead? If for readability, maybe we could change to something like `f'({j}, {xj})'`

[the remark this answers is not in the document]
MaxHalford: Nope, no particular reason to use tuples! Good to know they're heavier :). Use strings.

Issue #173, opened 2019-11-06
Implement Nadam optimizer


MaxHalford:
See [this](https://ruder.io/optimizing-gradient-descent/index.html#nadam).

MaxHalford:
Cheers @greatsharma! I'll add you to the list of contributors.

PR #219, opened 2020-01-17
Added Nadam

greatsharma:
Implemented nadam optimizer. I followed [this](https://ruder.io/optimizing-gradient-descent/index.html#nadam) paper mentioned in  #173


creme/optim/nadam.py
```diff
@@ -0,0 +1,39 @@
...
+        lr = self.learning_rate * \
+            (1 - self.beta_2 ** (self.n_iterations + 1)) ** .5
+        # Correct bias for `m`
+        lr /= (1 - self.beta_1 ** (self.n_iterations + 1))
+
+        for i, gi in g.items():
+            self.m[i] = self.beta_1 * self.m[i] + (1 - self.beta_1) * gi
+            self.v[i] = self.beta_2 * self.v[i] + (1 - self.beta_2) * gi ** 2
+
+            w[i] -= lr * (self.beta_1*self.m[i] + (1-self.beta_1)*gi*(1-np.power(
```
AdilZouitine: We try to be as least dependent on numpy as possible, I think it would be better to use the power operation available in the standard `math` library 😄

MaxHalford:
I'm having a hard-type verifying if the implementation is correct. @VaysseRobin do you have any thoughts?

greatsharma:
@MaxHalford where you think the implementation could be wrong?

Also in adam the default value for `lr` is `0.1` but in the mentioned original paper for adam the default value is `0.001`. Shouldn't we change it?

creme/optim/nadam.py
```diff
@@ -0,0 +1,68 @@
...
+            (1 - self.beta_2 ** (self.n_iterations + 1)) ** .5
+        # Correct bias for `m`
+        lr /= (1 - self.beta_1 ** (self.n_iterations + 1))
+
+        for i, gi in g.items():
+            self.m[i] = self.beta_1 * self.m[i] + (1 - self.beta_1) * gi
+            self.v[i] = self.beta_2 * self.v[i] + (1 - self.beta_2) * gi ** 2
+
+            w[i] -= lr * (self.beta_1*self.m[i] + (1-self.beta_1)*gi*(1-math.pow(
+                self.beta_1, self.n_iterations+1))) / (self.v[i] ** .5 + self.eps)
```
VaysseRobin: The implementation is almost correct but I see two problems (correct me if I am wrong), the first one is here `gi*(1-math.pow(
                self.beta_1, self.n_iterations+1))`, it should be a division instead of a multiplication between `gi` et the `(1-math.pow(...)`

The second one is a little bit tricky but I will try to explain you with some math.

VaysseRobin: The issue is because you do the bias correction in the learning rate which is usaually not a problem but in the case of NADAM, it is:

NADAM

So you should do the bias correction at each step.

greatsharma: > The implementation is almost correct but I see two problems (correct me if I am wrong), the first one is here `gi*(1-math.pow( self.beta_1, self.n_iterations+1))`, it should be a division instead of a multiplication between `gi` et the `(1-math.pow(...)`
> 
Thank you @VaysseRobin for this, it was a typo and lucky it worked well even with this. I changed it now

greatsharma: > The issue is because you integrate bias correction in the learning rate which is usaually not a problem but in the case of NADAM, it is:
> 
> NADAM
> 
> So you should do the bias correction at each step.

@VaysseRobin I changed it now. Thankyou :)

greatsharma:
> Also in adam the default value for `lr` is `0.1` but in the mentioned original paper for adam the default value is `0.001`. Shouldn't we change it?

@VaysseRobin what's your point on this?

creme/optim/nadam.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, lr=0.1, beta_1=0.9, beta_2=0.999, eps=1e-8):
+        super().__init__(lr)
+        self.beta_1 = beta_1
+        self.beta_2 = beta_2
+        self.eps = eps
+        self.m = collections.defaultdict(float)
+        self.v = collections.defaultdict(float)
+        self.m_hat = collections.defaultdict(float)
+        self.v_hat = collections.defaultdict(float)
```
VaysseRobin: You don't need to store history of `m_hat` and `v_hat` since the update of `m` and  `v` does not rely on them. So you should just use local variables in `_update_after_pred`.

VaysseRobin:
> > Also in adam the default value for `lr` is `0.1` but in the mentioned original paper for adam the default value is `0.001`. Shouldn't we change it?
> 
> @VaysseRobin what's your point on this?

I think that it does not really matter, but yes it could be good to match the paper.

creme/optim/nadam.py
greatsharma: I changed it now

VaysseRobin:
It seems good to me, nice job ! @MaxHalford I think you can merge it.

greatsharma:
> > > Also in adam the default value for `lr` is `0.1` but in the mentioned original paper for adam the default value is `0.001`. Shouldn't we change it?
> > 
> > 
> > @VaysseRobin what's your point on this?
> 
> I think that it does not really matter, but yes it could be good to match the paper.

@MaxHalford what you think?

MaxHalford:
@greatsharma I think it doesn't really matter. What we will do one day is to define a large set of benchmarks, we will then benchmark each optimizer and set the parameters that are the best overall.

PR #220, opened 2020-01-18
More Preprocessing

greatsharma:
Added `MaxAbsScaler` and `RobustScaler`. 

Also added support of `with_mean` and `with_std` to `StandardScaler` as there may be the case in which user only want to do one of the two things and also it makes the class more close and consistent with `sklearn.StandardScaler`.

Some of the changes are just code formatting because it is automatically done by my visual studio code. Although these formatting made the code more readable and good looking but if it's not suitable then I will try to undo them in some other text editor.


greatsharma:
I am looking to add some more preprocessing in this PR.

MaxHalford:
Hey @greatsharma, your 2 PR look good. I'll merge them once I get back home from my trip. Good work :)

greatsharma:
> Hey @greatsharma, your 2 PR look good. I'll merge them once I get back home from my trip. Good work :)

Thankyou @MaxHalford . Also I added `RobustScaler` but the results are not that close to `sklearn.RobustScaler` like other scalers are. On inspecting I found that our `online IQR` is not that close as `batch IQR` . So is there any hope to improve that in turns to improve the `stats.Quantile`

greatsharma:
@MaxHalford any suggestion/comment on this PR. Is there anything needs to be changed in this?

creme/stats/median.py
```diff
@@ -0,0 +1,107 @@
...
+from . import base
+
+
+__all__ = [
+    'Median',
+    'RollingMedian'
+]
+
+
+class Median(base.Univariate):
```
MaxHalford: I don't think I like this class... It sure works but it's going to keep growing as time goes on, which is not in the spirit of the library. You should be using `Quantile(0.5)` for the median.

creme/stats/median.py
```diff
@@ -0,0 +1,107 @@
...
+        return self
+
+    def get(self):
+        if self.len_ % 2:
+            return self.sorted_list[self.len_//2]
+        else:
+            return (self.sorted_list[self.len_ // 2 - 1] + self.sorted_list[self.len_ // 2]) / 2
+
+
+class RollingMedian(base.RollingUnivariate, utils.SortedWindow):
```
MaxHalford: Same here, we should be using `RollingQuantile`.

creme/preprocessing/scale.py
```diff
@@ -170,11 +185,184 @@ def fit_one(self, x, y=None):
...
+                [-0.72177808],
+                [-0.88142503],
+                [ 0.4361963 ],
+                [ 0.00400545],
+                [ 0.17840326],
+                [-0.94256856],
+                [ 0.60533751],
+                [ 0.38119272],
+                [-0.62990639],
+                [-0.6796607 ]])
```
MaxHalford: Performance is not that good but it'll get better if we improve the accuracy of the `Quantile` class.

greatsharma:
I am having some git issues with this branch and it seem to tedious to solve it in this branch. So I am closing this PR and reopening it.

creme/stats/median.py
greatsharma: @MaxHalford  I did the mentioned changes here #224 . I closed this PR because I was having lot of merge conflicts and rebasing problems. So I closed this PR and opened #224 . We can continue to discuss changes there. Sorry for inconvenience :(

Issue #182, opened 2019-11-24
Averaged SGD


MaxHalford:
Leon Bottou discusses it [here](https://leon.bottou.org/publications/pdf/compstat-2010.pdf) while PyTorch has implemented it [here](https://pytorch.org/docs/stable/_modules/torch/optim/asgd.html). Ideally I think we should implement an `Averager` class which takes as input any optimizer. This way we can do averaging on SGD by doing `Averager(SGD(0.01))` but also on Adam, AdaGrad, etc.

I tried to implement averaged stochastic gradient descent but I ran into some API issues. I'm too tired to solve it straightaway so I rather open an issue. Feel free to take a look at it.

PR #222, opened 2020-01-25
Averaged SGD

greatsharma:
An `Averager` class mentioned in #182 

I follwed this [https://leon.bottou.org/publications/pdf/compstat-2010.pdf](this)

I am not sure whether this current implementation of `Averager` class resembles the intended api.


greatsharma:
I have to make some changes. I am closing it for now

PR #223, opened 2020-01-26
Added SGD Averager and Cyclic scheduler

greatsharma:
An `Averager` class mentioned in #182.

I follwed this [https://arxiv.org/abs/1803.05407](this)

### DIFF
--- a/creme/optim/averager.py
+++ b/creme/optim/averager.py
@@ -1,0 +1,83 @@
+import collections
+
+from . import base
+from . import schedulers
+
+
+__all__ = ['Averager']
+
+
+class Averager(base.Optimizer):
+    """An Stochastic weight averaging wrapper around various sgd optimzers.
+
+    Example:
+
+        ::
+
+            >>> from creme import linear_model
+            >>> from creme import metrics
+            >>> from creme import model_selection
+            >>> from creme.optim.schedulers import Cyclic
+            >>> from creme.optim import Adam, Averager
+            >>> from creme import preprocessing
+            >>> from creme import stream
+            >>> from sklearn import datasets
+
+            >>> X_y = stream.iter_sklearn_dataset(
+            ...     dataset=datasets.load_breast_cancer(),
+            ...     shuffle=True,
+            ...     random_state=42
+            ... )
+
+            >>> adam = Adam(lr=Cyclic(1, 0.01, 100))
+            >>> optimizer = Averager(adam)
+            >>> model = (
+            ...     preprocessing.StandardScaler() |
+            ...     linear_model.LogisticRegression(optimizer)
+            ... )
+            >>> metric = metrics.F1()
+
+            >>> model_selection.progressive_val_score(X_y, model, metric)
+            F1: 0.959441
+
+    References:
+        1. `Averaging Weights Leads to Wider Optima and Better Generalization
+            <https://arxiv.org/abs/1803.05407>`_
+
+    """
+
+    def __init__(self, base_optim):
+        self.base_optim = base_optim
+        self.n_iterations = 0
+        self.c = (
+            base_optim.lr.c
+            if isinstance(base_optim.lr, schedulers.Cyclic) else
+            1
+        )
+        self.w_ = collections.defaultdict(float)
+        self.w_swa = collections.defaultdict(float)
+
+    def update_before_pred(self, w: dict) -> dict:
+        return self.base_optim.update_before_pred(w)
+
+    def _update_after_pred(self, w: dict, g: dict) -> dict:
+
+        # Update the weights
+        self.w_ = self.base_optim._update_after_pred(self.w_.copy(), g)
+
+        if self.n_iterations % self.c == 0:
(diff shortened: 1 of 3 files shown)


greatsharma:
While implementing this I struggled a bit majorly due to internals of api. And this code can be improved, currently this is what I made.

creme/optim/averager.py
```diff
@@ -0,0 +1,83 @@
...
+        self.w_ = collections.defaultdict(float)
+        self.w_swa = collections.defaultdict(float)
+
+    def update_before_pred(self, w: dict) -> dict:
+        return self.base_optim.update_before_pred(w)
+
+    def _update_after_pred(self, w: dict, g: dict) -> dict:
+
+        # Update the weights
+        self.w_ = self.base_optim._update_after_pred(self.w_.copy(), g)
```
MaxHalford: Shouldn't you be using `w`?

creme/optim/averager.py
```diff
@@ -0,0 +1,83 @@
...
+    References:
+        1. `Averaging Weights Leads to Wider Optima and Better Generalization
+            <https://arxiv.org/abs/1803.05407>`_
+
+    """
+
+    def __init__(self, base_optim):
+        self.base_optim = base_optim
+        self.n_iterations = 0
+        self.c = (
```
MaxHalford: What is `c`?

creme/optim/schedulers.py
```diff
@@ -33,6 +34,35 @@ def get(self, t):
...
+        c (int): Cycle length.
+
+    References:
+        1. `Averaging Weights Leads to Wider Optima and Better Generalization
+            <https://arxiv.org/abs/1803.05407>`_
+
+    """
+
+    def __init__(self, lr1, lr2, c):
+        assert lr1 >= lr2, "lr1 should be greater than or equal to lr2"
```
MaxHalford: There shouldn't be `assert`s in library code! They're reserved for test code :). You should use `if`s and raise `ValueError` exceptions.

creme/optim/averager.py
```diff
@@ -0,0 +1,83 @@
...
+
+from . import base
+from . import schedulers
+
+
+__all__ = ['Averager']
+
+
+class Averager(base.Optimizer):
+    """An Stochastic weight averaging wrapper around various sgd optimzers.
```
MaxHalford: I'm not sure you've implemented the correct algorithm. I was thinking about this one: https://arxiv.org/pdf/1107.2490v2.pdf, which not the same thing. It's the one implemented by `sklearn` in it's `SGDClassifier` class.

creme/optim/averager.py
greatsharma: Initially I used `w` but didn't get expected good results. Then on inspecting I found that in `_update_after_pred` of any `creme.optim.Optimizer` the param `w` is the same as whatever we returned in the previous iteration(like if in previous iteration `_update_after_pred` returned `[0.1, 0.2, 0.3]` than in next iteration param `w` will have this value). But in the case of `Averager` we don't want this `w` to train our `base_optim` because this is the average of previous `w's`. So I stored the immediate recent weights trained by the `base_optim` in `w_` and in the next iteration passed these `w_` to the `base_optim` rather than the averaged one. So it's like the `base_optim` has no idea it is going to averaged. And I think this is what we intended with `Averager`, to take an optimizer and average it's weight in regular intervals.

creme/optim/averager.py
greatsharma: `c` is the cycle length of the `Cyclic lr` scheduler. If the `base_optim` is using an `Cyclic lr` then `Averager` will average the weights in interval of length `c` and if `base_optim` is not using `Cyclic lr` than `c=1` which means `Averager` will average the weights after every iteration.

Actually in the mentioned paper and blogs I read on Averaging, they used a `Cyclic scheduler` for `lr`. That's why I add this.

creme/optim/averager.py
greatsharma: I will look at your proposed resources :)

I summarize what I did or atleast what I intended to do -

A `base_optim` which is the core optimizer around which we average the weights. `c` is the cycle length as I explained in above comment. The `_update_after_pred` first update the weights according to the `base_optim` and store them in `w_` for next iteration in this line,
`self.w_ = self.base_optim._update_after_pred(self.w_.copy(), g)`

Then `if not self.n_iterations % self.c` will check whether to take average (if `c` is 1 than this always evaluates to `True`) and if it's true than we update `w_swa`. Then we return `w_swa` if we averaged the weights atleast once otherwise return `w_` (the weights updated by `base_optim`).

The API may look complex because of the wrapping behaviour of `Averager`.

Also I can remove `c` to make it simple but in that case averaging will occur in every iteration.

creme/optim/schedulers.py
greatsharma: Fixed it.

MaxHalford:
I'm closing this because it isn't what we're looking for :(

PR #224, opened 2020-01-27
More Preprocessing

greatsharma:
Added `MaxAbsScaler`, `RobustScaler`, `AbsMax` and `RollingAbsMax`.

Also added support of `with_mean` and `with_std` to `StandardScaler` as there may be the case in which user only want to do one of the two things and also it makes the class more close and consistent with `sklearn.StandardScaler`.


creme/preprocessing/scale.py
```diff
@@ -25,6 +28,10 @@ class StandardScaler(base.Transformer):
     different than when using scikit-learn but this doesn't seem to have any impact on learning
     performance.
 
+    Parameters:
+        with_mean (boolean): Whether to centre the data before scaling. Defaults to ``True``.
```
MaxHalford: Should be `bool` not `boolean`.

MaxHalford:
Good stuff @greatsharma :)

creme/preprocessing/scale.py
greatsharma: I corrected it now

MaxHalford:
You now need to check out Travis and see why some of the tests failed.

greatsharma:
> You now need to check out Travis and see why some of the tests failed.

@MaxHalford I fixed them now. I think it would be nice if a code formatter is added to this library like `BLACK`. What's your thought on this?

MaxHalford:
Good work. Yeah we definitely have to set some rules in stone regarding code style. I don't really like `black` for many reasons. I think that a let invasive solution such as a `.pylintrc` file would be great :)

PR #227, opened 2020-01-30
Added Binarizer and minor doc change

greatsharma:
In `Binarizer`,  `self.on` contains the features which we need to binarize. Because by default it will binarize all numeric features but in many situations we only need to binarize selective features.


greatsharma:
I missed something

PR #232, opened 2020-02-05
Added Binarizer

greatsharma:


creme/preprocessing/kernel_approx.py
```diff
@@ -8,7 +8,7 @@
 class RBFSampler(base.Transformer):
     """Extracts random features which approximate an RBF kernel.
 
-    This is a powerful to give non-linear capacity to linear classifiers.
+    This is powerful to give non-linear capacity to linear classifiers.
```
MaxHalford: Actually the correction is "this is a powerful way to give" :)

greatsharma: I updated it :)

PR #233, opened 2020-02-05
minor AdaBound fix

greatsharma:
I was trying various optimizers in loop and accidentally run AdaBound with `schedulers` and got `TypeError: unsupported operand type(s) for /: 'float' and 'InverseScaling'` at line 70.

Because `AdaBound` is implicitly changing `lr` and don't require `scheduler`. So it is better to raise an error if in case user uses non-numeric `lr` because the default error is non-informatic.

### DIFF
--- a/creme/optim/ada_bound.py
+++ b/creme/optim/ada_bound.py
@@ -49,7 +49,11 @@ class AdaBound(base.Optimizer):
 
         if not isinstance(lr, numbers.Number):
             raise ValueError(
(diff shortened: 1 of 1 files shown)


MaxHalford:
Nice catch. Doesn't `final_lr` also have to be exclusively a number?

greatsharma:
> Nice catch. Doesn't `final_lr` also have to be exclusively a number?

I missed that !
Now I fixed it :)

PR #278, opened 2020-03-31
[PR] Stream utility for SQLAlchemy

3outeille:
Questions :
    - Do we have to handle the engine connection closure?
    - What if user inputs a wrong database string URI ?
    - Can a database file be stored online and not necessarily locally?


AdilZouitine:
The tests failed because it didn't find the function you created, you have to add it in this file `creme/stream/__init__.py`.

By importing it and adding it in the `_all__ `

creme/stream/modif-bouteille/sql.py
```diff
@@ -0,0 +1,25 @@
+from sqlalchemy import create_engine, MetaData, select
+import os
+
+def iter_sql(con):
+    """Yields rows from all tables in a database.
+    
+    Parameters:
+        con: SQLAlchemy connectable (engine/connection) or database str URI.
```
AdilZouitine: You should specify the type in the docsting :) 

```python
Parameters:
   param1 (type): description

```

MaxHalford:
Okay so some points to make things clearer:

- We want to stream of the result of a query, not over the data. In other words we want to implement the `read_sql` function from pandas, but in a streaming manner. However we could also in a later stage implement a `read_sql_table` which streams the data from a single relation. I don't much use in streaming all the data from the whole database.
- The tests you have provided are not good enough because they aren't checking anything. They don't demonstrate how to use the function, but they don't prove that the function works as intended.
- As @AdilZouitine hinted at there is some work to do before this PR can be accepted. Don't be frightened, we're here to help you and make you progress. Just keep asking questions and we'll get there.
- It's okay to provide a database file to run tests against, but it's going to have to be smaller than 50MB. An alternative would be to provide the SQL code to generate the database file, then we can run everytime `pytest` is called. I can help with this if you can provide the SQL code.

To answer your questions in particular:

> Do we have to handle the engine connection closure?

No we don't. The user should provide the engine and should close. In fact the user should provide the engine directly, not the connection string.

> What if user inputs a wrong database string URI ?

That's his/her fault. If the user provides the engine to us, then he/she will get that error before anyway.

> Can a database file be stored online and not necessarily locally?

Yes, definitely. This is all specified in the database URI. The database can even be stored in memory with SQLite.

creme/stream/modif-bouteille/sql.py
MaxHalford: The type of `con` should be [`sqlalchemy.engine.Engine`](https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.Engine).

3outeille:
Thanks for all your feedbacks. I'll get back to work !

3outeille:
I created a small database for testing purpose. It contains 2 tables (census, state_fact) with 15 rows each. Here is the result after the following query: "SELECT * FROM census"

image

3outeille:
Seems like the tests aren't passing because of the following error: "ModuleNotFoundError: No module named 'sqlalchemy'".

MaxHalford:
Yeah to make the tests pass you need to add SQLAlchemy as a dependency in the `dev_packages` list in `setup.py`. By default I think that we don't to require the user to install SQLAlchemy. Therefore, you can also include it as an optional dependency by adding to the `compat_packages` list in `setup.py`.

Your code is getting better and better. Could you please add the tests you've provide in a doctest or in a `test_sql.py` file, instead of in a `bouteille_modif` folder.

Keep up the good work!

3outeille:
Seems like mypy can't find a type hints for sqlalchemy. According to the doc,  we should add ``` add a # type: ignore```.  What do I have to do ?

Here is the following log:

```
creme/stream/test_sql.py:1: error: Cannot find implementation or library stub for module named 'sql'
creme/stream/test_sql.py:2: error: Skipping analyzing 'sqlalchemy': found module but no type hints or library stubs
creme/stream/create_database.py:1: error: Skipping analyzing 'sqlalchemy': found module but no type hints or library stubs
creme/stream/create_database.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

MaxHalford:
Oh yeah you need to an exception in the `mypy.ini` file.

creme/stream/test_sql.py
```diff
@@ -1,4 +1,5 @@
-from sql import iter_sql # type: ignore
+import sql
+from sql import iter_sql
```
MaxHalford, line 2: You need to do `from . import sql`. Then you can call `sql.iter_sql`.

creme/stream/test_sql.py
```diff
@@ -1,5 +1,4 @@
-import sql
-from sql import iter_sql 
+from . import sql
 from sqlalchemy import create_engine, MetaData
```
MaxHalford, line 2: Okay now you need to put your test code inside a function. I invite you check out other test files to get an idea of how to set this up. Feel free to ping me if you need help.

creme/stream/test_sql.py
3outeille: oh thanks god, the struggle was real

MaxHalford: Yeah importing modules when you're developping a library takes some practice. I suggest reading [this](http://www.blog.pythonlibrary.org/2016/03/01/python-101-all-about-imports/) article.

3outeille:
@MaxHalford  I have tried written some test. If you have time, can you take a quick look ?

Few issues I'm concerned about my tests:

- my ```@pytest.mark.paremetrize``` seems too long. Should I rather compare the lenght of keys/values instead of their contents ?
- isn't my ```test_sql()``` function going to be called for each ```query``` and thus, overwrite the database ? (because of the ```initialization()```)

MaxHalford:
Hey @3outeille. So I took the liberty to do some tidying up on your branch. I hope you don't mind. I added a zipped file with some pokemons from generation 1 as an example.

There is still some work to do though! Mainly we also want to add a `iter_sql_table` method which calls `iter_sql` and generates the appropriate SQL for querying a given table.

The other thing we need to work is a bit more complex. Basically I was expecting that SQLAlchemy streams the rows one-by-one when we loop over the result of a query. It does, but it also seems that it pre-fetches the entire results before streaming the results. We need to look into how to tell SQLAlchemy to yield the results on the fly, without pre-fetching the results. It's a bit technical but there seems to be [some stuff online](https://www.google.com/search?q=sqlalchemy+stream+results), after all this is a really useful thing to have.

Take the time to look how I set up the tests!

PR #279, opened 2020-04-01
Handle zero division error in SMAPE

raphaelsty:
Hi @MaxHalford,

Is this ok to return 0 if the denominator value in SMAPE is equal to 0? 

Raphaël :-)


raphaelsty:
Too slow, already updated few seconds ago :-)

PR #291, opened 2020-04-24
Shift values when using ``feature_extraction.agg``and ``feature_extraction.TargetAgg``

raphaelsty:
Being able to add lag to a pipeline is a useful feature. I added this feature to the transformers ``feature_extraction.agg``and ``feature_extraction.TargetAgg``. By default the shift is set to 0. I added to the transformers an additional parameter ``missing`` to handle missing values when shifting the input series. ``missing``is set to 0 by default.

```python
>>> from creme import feature_extraction
>>> from creme import stats

>>> X = [
...     {'place': 'Taco Bell', 'revenue': 42},
...     {'place': 'Burger King', 'revenue': 16},
...     {'place': 'Burger King', 'revenue': 24},
...     {'place': 'Taco Bell', 'revenue': 58},
...     {'place': 'Burger King', 'revenue': 20},
...     {'place': 'Taco Bell', 'revenue': 50}
... ]

>>> agg = feature_extraction.Agg(
...     on='revenue',
...     by='place',
...     how=stats.Mean(),
...     shift=1,
...     missing=0,
... )

>>> for x in X:
...     print(agg.fit_one(x).transform_one(x))
{'revenue_mean_shift_1_by_place': 0}
{'revenue_mean_shift_1_by_place': 0}
{'revenue_mean_shift_1_by_place': 16.0}
{'revenue_mean_shift_1_by_place': 42.0}
{'revenue_mean_shift_1_by_place': 20.0}
{'revenue_mean_shift_1_by_place': 50.0}
```

```python
>>> import creme

>>> X = [
...     {'place': 'Taco Bell', 'revenue': 42},
...     {'place': 'Burger King', 'revenue': 16},
...     {'place': 'Burger King', 'revenue': 24},
...     {'place': 'Taco Bell', 'revenue': 58},
...     {'place': 'Burger King', 'revenue': 20},
...     {'place': 'Taco Bell', 'revenue': 50}
... ]

# You can add lag to ``feature_extraction.TargetAgg`` by setting the parameter
# ``shift``. The feature extration pipeline will shift the input targets serie with
# respect to the aggregates.

>>> agg = creme.feature_extraction.TargetAgg(
...     by='place',
...     how=creme.stats.RollingMean(1),
...     shift=1,
...     missing=0,
... )

>>> for x in X:
...     print(agg.transform_one(x))
...     y = x.pop('revenue')
...     agg = agg.fit_one(x, y)
{'target_rolling_rollingmean_1_shift_1_by_place': 0}
{'target_rolling_rollingmean_1_shift_1_by_place': 0}
{'target_rolling_rollingmean_1_shift_1_by_place': 16.0}
{'target_rolling_rollingmean_1_shift_1_by_place': 42.0}
{'target_rolling_rollingmean_1_shift_1_by_place': 24.0}
{'target_rolling_rollingmean_1_shift_1_by_place': 58.0}
```

Raphaël


creme/feature_extraction/agg.py
```diff
@@ -18,10 +19,15 @@ class Agg(base.Transformer):
     aggregate statistics in one go. The keys of the `dict` containing the aggregates is
     automatically guessed from `how`, `by`, and `on`.
 
+    ``shift``parameters add a defined lag to the input stream. The running statistics returns
```
MaxHalford, line 22: Only one backtick is needed now, so `shift` instead of ``shift``. This goes for the rest of the inline stuff. 

The comment should rather be: "The `shift` parameter determines the lag applied to the input stream.".

creme/feature_extraction/agg.py
```diff
@@ -56,16 +62,43 @@ class Agg(base.Transformer):
         {'revenue_mean_by_place': 20.0}
         {'revenue_mean_by_place': 50.0}
 
+        # You can add lag to ``feature_extraction.Agg`` by setting the parameter ``shift``.
+        # The feature extration pipeline will shift the input serie with respect to the
+        # aggregates.
```
MaxHalford, line 67: No need for `#`

creme/feature_extraction/agg.py
```diff
@@ -56,16 +62,43 @@ class Agg(base.Transformer):
...
+        {'revenue_mean_shift_1_by_place': 20.0}
+        {'revenue_mean_shift_1_by_place': 50.0}
+
     References:
         1. [Streaming groupbys in pandas for big datasets](https://maxhalford.github.io/blog/streaming-groupbys-in-pandas-for-big-datasets/)
 
     """
 
-    def __init__(self, on: str, by: str, how: stats.Univariate):
+    def __init__(self, on: str, by: str, how: stats.Univariate, shift:int = 0, missing: float=0.):
```
MaxHalford, line 91: You don't to declare the type of `shift`, it's inferred from the default value AFAIK. Same goes for the missing parameter.

creme/feature_extraction/agg.py
```diff
@@ -56,16 +62,43 @@ class Agg(base.Transformer):
...
         self.on = on
         self.by = by if isinstance(by, (typing.List, typing.Tuple)) else [by]
-        self.how = how
-        self.groups = collections.defaultdict(functools.partial(copy.deepcopy, how))
+
+        if shift:
+            self.how = utils.Shift(period=shift, stats=how, missing=missing,
+                fitting_before_transform=True)
```
MaxHalford, line 97: I would prefer to call this parameter `fit_before_transform`.

creme/utils/shift.py
```diff
@@ -0,0 +1,100 @@
...
+        period (int): Number of periods to shift. Must be positive.
+        stats (stats.Univariate, None): The statistic to compute. When set to None, shift the input
+            series of values without updating any statistics.
+        missing: Handle missing values when there is'nt any reference for a given period.
+        fitting_before_transform (bool): Shift logic. Must be set to True when the user updates the
+            buffer before retrieving the observation with the offset selected.
+
+    Example:
+
+        ::
```
MaxHalford, line 41: We don't need to do this anymore. A single tab suffices. Check out the other docstrings :)

creme/utils/shift.py
```diff
@@ -0,0 +1,100 @@
...
+            >>> X = [
+            ...    1,
+            ...    2,
+            ...    3,
+            ...    4,
+            ...    5,
+            ... ]
+
+    """
+    def __init__(self, period: int, stats=DefaultStat(), missing: float=0.,
```
MaxHalford, line 79: Is there any point of having `DefaultStat` as a default?

Also now that I see the code I think we can do something better. As far as I can tell is that we want to compose `Shift` and `Mean`. Therefore, instead of passing a `stat` to `Shift`, we could instead implement a pipe operator to be able to chain statistics. Therefore we could write something like `Shift(1) | Mean()`. That would be cool because it means we don't have to modify the code of `Agg` and `TargetAgg`! Tell me what you think about this.

Issue #282, opened 2020-04-05
Implement AdPredictor algorithm


MaxHalford:
See [this](http://tullo.ch/articles/online-learning-with-adpredictor/) blog post.

dataJSA:
Hi, I would love to try this one, if you could assigne me the issue it would be perfect !

MaxHalford:
[Moved to the icebox](https://github.com/online-ml/river/projects/2#card-55044157).

PR #294, opened 2020-05-08
AdPredictor

dataJSA:
This is what I have implemented so far concerning  #282, I still have to do a few things : 

- [ ] Put the `NormalPrior` class in the `optim.initializers` module (if it is fine as it is) 
- [ ] Add comments
- [ ] Check formatting 
- [ ] Run the different tests

Let me know what you think !

### DIFF
--- a/=0.1
+++ b/=0.1
@@ -1,0 +1,2 @@
+Collecting package metadata (repodata.json): ...working... done
+Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
--- a/creme/linear_model/__init__.py
+++ b/creme/linear_model/__init__.py
@@ -5,6 +5,7 @@ from .glm import LogisticRegression
 from .pa import PAClassifier
 from .pa import PARegressor
 from .softmax import SoftmaxRegression
+from .adpredictor import AdPredictor
 
 
 __all__ = [
@@ -13,5 +14,6 @@ __all__ = [
     'LogisticRegression',
     'PAClassifier',
     'PARegressor',
-    'SoftmaxRegression'
+    'SoftmaxRegression',
+    'AdPredictor'
 ]
--- a/creme/linear_model/adpredictor.py
+++ b/creme/linear_model/adpredictor.py
@@ -1,12 +1,8 @@
 import collections
-import math 
-import numbers
-
 import numpy as np 
-import scipy as sp 
+from scipy import stats
 from .. import base 
 from .. import optim
-from .. import proba
 from .. import utils
 
 
@@ -29,11 +25,14 @@ class NormalPrior(optim.initializers.Initializer):
         self.prior_probability = prior_probability
         self.beta = beta
         self.n_features = n_features
-        self.mean = 0 if prior_probability is None else sp.stats.norm.ppf(prior_probability) * (beta ** 2 + n_features)
+        self.mean = 0 if prior_probability is None else stats.norm.ppf(prior_probability) * (beta ** 2 + n_features)
         self.variance = 1.0
 
     def __str__(self):
-        return f'Unbiased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})' if prior_probability is None else f'Biased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})'
+        return f'Unbiased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})' if self.prior_probability is None else f'Biased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})'
+
+    def __call__(self):
+        return self
 
 
 class BOPR:
@@ -57,7 +56,7 @@ class BOPR:
     def _gaussian_correction(self, t):
 
         t = utils.math.clamp(t, minimum=-self.surprise, maximum=self.surprise)
-        v = sp.stats.norm.pdf(t)/sp.stats.norm.cdf(t)
+        v = stats.norm.pdf(t)/stats.norm.cdf(t)
         w = v * (v + t)
         return (v, w)
 
@@ -70,7 +69,7 @@ class BOPR:
         return prior
 
 
-class AdPredictor(BOPR, base.Regressor):
+class AdPredictor(BOPR, base.BinaryClassifier):
     """AdPredictor.
     """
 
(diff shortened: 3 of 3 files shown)


=0.1
```diff
@@ -0,0 +1,2 @@
+Collecting package metadata (repodata.json): ...working... done
+Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
```
MaxHalford, line 2: You need to remove this.

creme/linear_model/__init__.py
```diff
@@ -5,6 +5,7 @@
 from .pa import PAClassifier
 from .pa import PARegressor
 from .softmax import SoftmaxRegression
+from .adpredictor import AdPredictor
```
MaxHalford, line 8: Can you rename the module to `ad_predictor`?

creme/linear_model/__init__.py
```diff
@@ -13,5 +14,6 @@
     'LogisticRegression',
     'PAClassifier',
     'PARegressor',
-    'SoftmaxRegression'
+    'SoftmaxRegression',
+    'AdPredictor'
```
MaxHalford, line 18: Please sort this list alphabetically.

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+from scipy import stats
+from .. import base 
+from .. import optim
+from .. import utils
+
+
+
+__all__ = [
+    'AdPredictor'
+]
```
MaxHalford, line 13: This should be:

```python
import collections
import numbers

import numpy as np 
from scipy import stats

from .. import base 
from .. import optim
from .. import utils


__all__ = [
    'AdPredictor'
]
```

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class NormalPrior(optim.initializers.Initializer):
```
MaxHalford, line 16: I'm not sure this has inherit from `Initializer`. Initializers are here to provide a way to generate random weights when initializing a model, but your case is a bit different.

I understand that there is some need for custom weights, in which case you could use something like a `collections.namedtuple`. You could also store the means and variances in two separate lists.

There's some other style stuff to change, but I'll add some comments once this is refactored :)

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+        self.variance = 1.0
+
+    def __str__(self):
+        return f'Unbiased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})' if self.prior_probability is None else f'Biased normal prior on weights 𝒩(μ={self.mean:.3f}, σ2={self.variance:.3f})'
+
+    def __call__(self):
+        return self
+
+
+class BOPR:
```
MaxHalford, line 61: Is there any reason why this is it's own particular class? If we're not going to reuse then we may as well incorporate it's logic in the `AdPredictor` class.

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+        return 1.0 if y == 1 else -1.0
+
+
+class AdPredictor(BOPR, base.BinaryClassifier):
+    """AdPredictor is a bayesian online probit regression model devised for search advertising click through rate (CTR) prediction.
+
+    AdPredictor
+
+    Parameters:
+        initializer (linear_model.adpredictor.NormalPrior): The initializer used to set beliefs over the weights.
```
MaxHalford, line 117: With the new documentation system, types are specified as type annotations in the `__init__` method, rather than in the docstring.

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+        surprise (float): Clipping value of the learning step function parameter ensure numerical stability.
+        epsilon (float): Noise parameter of the dynamic correction process.
+
+    Attributes:
+        weights (collections.defaultdict): The current gaussian belief over the weights.
+
+    Example:
+
+        ::
```
MaxHalford, line 126: No need for `::` anymore. Also you can move the example back one tab.

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+            >>> model = (
+            ...     compose.Select('age_of_domain','ip_in_url')|
+            ...     preprocessing.OneHotEncoder('age_of_domain')|
+            ...     linear_model.AdPredictor(initializer=NormalPrior(beta=0.1, n_features=X_y.n_features, prior_probability=0.6), surprise=5, epsilon=0.001)
+            ... )
+
+            >>> metric = metrics.ROCAUC()
+
+            >>> model_selection.progressive_val_score(X_y, model, metric)
+            ROCAUC: 0.540405
```
MaxHalford, line 139: That doesn't look too good! It might be nice to take Andrew Tulloch's example and to check whether you're able to reach the same performance as he does.

creme/linear_model/adpredictor.py
```diff
@@ -0,0 +1,175 @@
...
+        y = BOPR._target_encoding(y)
+        v, w = self._gaussian_correction(y*total_mean/total_variance)
+
+        for i, xi in x.items():
+            weight = self.weights[i]
+            mean_delta = y * weight.variance / np.sqrt(total_variance) * v
+            variance_multiplier = 1.0 - weight.variance / total_variance * w
+            weight.mean += mean_delta
+            weight.variance *= variance_multiplier
+            self.weights[i] = weight
```
MaxHalford, line 170: You need to return `self`.

creme/linear_model/__init__.py
dataJSA: Yes !

creme/linear_model/adpredictor.py
dataJSA: Thanks

creme/linear_model/adpredictor.py
dataJSA: Ahahah yes it's pretty bad I  benchmarked it against the logistic regression on same features `ROCAUC: 0.617649`. I will refactor the code and see if I didn't miss anything and benchmark it with Andrew Tulloch's implementation  !

creme/linear_model/adpredictor.py
dataJSA: Ok I will switch to that

creme/linear_model/adpredictor.py
dataJSA: Ok I will try that. Thanks for all your comments, it helps a lot !

=0.1
dataJSA: ok don't where it's coming from

Issue #267, opened 2020-03-20
Add partial_fit to scikit-learn wrappers


MaxHalford:
We should add `partial_fit` methods to the `compat.Creme2SKL*` wrapper classes. At the same time it would be good to test if converting an `sklearn` model to a `creme` model, and then back again to an `sklearn` model produces a class which passes the `sklearn.estimator_checks.check_estimator` test suite.

brcharron:
I gave this one a go, please have a look at the PR #312.

MaxHalford:
Thanks @brcharron for taking care of this.

PR #312, opened 2020-06-30
Add partial_fit to scikit-learn wrappers of creme estimators #267

brcharron:
Also fix the Creme2SKLClassifier so it passes scikit-learn's checks.
Ensure SKL2CremeRegressor -> Creme2SKLRegressor is still a proper
estimator. Not ensured for classifiers at the moment due to discrepancy
in how classes are discovered (at first partial_fit for scikit-learn and
at __init__ for creme if prediction classes are to be consistent).

### DIFF
--- a/creme/compat/creme_to_sklearn.py
+++ b/creme/compat/creme_to_sklearn.py
@@ -106,6 +106,27 @@ class Creme2SKLRegressor(Creme2SKLBase, sklearn_base.RegressorMixin):
 
         self.estimator = estimator
 
+    def _partial_fit(self, X, y):
+
+        # Check the inputs
+        X, y = utils.check_X_y(X, y, **SKLEARN_INPUT_X_PARAMS, **SKLEARN_INPUT_Y_PARAMS)
+
+        # Store the number of features so that future inputs can be checked
+        if hasattr(self, 'n_features_in_') and X.shape[1] != self.n_features_in_:
+            raise ValueError(f'Expected {self.n_features_in_} features, got {X.shape[1]}')
+        self.n_features_in_ = X.shape[1]
+
+        # scikit-learn's convention is that fit shouldn't mutate the input parameters; we have to
+        # deep copy the provided estimator in order to respect this convention
+        if not hasattr(self, 'instance_'):
+            self.instance_ = copy.deepcopy(self.estimator)
+
+        # Call fit_one for each observation
+        for x, yi in STREAM_METHODS[type(X)](X, y):
+            self.instance_.fit_one(x, yi)
+
+        return self
+
     def fit(self, X, y):
         """Fits to an entire dataset contained in memory.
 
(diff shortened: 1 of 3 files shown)


creme/compat/creme_to_sklearn.py
```diff
@@ -164,54 +190,104 @@ class Creme2SKLClassifier(Creme2SKLBase, sklearn_base.ClassifierMixin):
...
+        # scikit-learn's convention is that fit shouldn't mutate the input parameters; we have to
+        # deep copy the provided estimator in order to respect this convention
+        if not hasattr(self, 'instance_'):
+            self.instance_ = copy.deepcopy(self.estimator)
 
         # creme's BinaryClassifier expects bools or 0/1 values
         if not isinstance(self.estimator, base.MultiClassifier):
-            self.label_encoder_ = preprocessing.LabelEncoder().fit(y)
+            if not hasattr(self, 'label_encoder_'):
+                self.label_encoder_ = preprocessing.LabelEncoder().fit(classes)
```
MaxHalford, line 241: Very nice.

brcharron:
Thanks for the review.
My bad, I got it right for the classifier as it was a bit tricky but went too fast on the other 3. Actually `fit` was idempotent but so was `partial_fit`, which it shouldn't.

creme/compat/creme_to_sklearn.py
```diff
@@ -99,42 +99,73 @@ class Creme2SKLRegressor(Creme2SKLBase, sklearn_base.RegressorMixin):
...
+            y: array-like of shape n_samples.
 
-        return self
+        Returns:
+            self
+
+        """
+
+        # Fit with one pass of the dataset portion
+        return self._partial_fit(X, y)
```
MaxHalford, line 162: Is there any reason why need `_partial_fit`? Can't we just have `partial_fit`?

creme/compat/creme_to_sklearn.py
```diff
@@ -164,54 +198,102 @@ class Creme2SKLClassifier(Creme2SKLBase, sklearn_base.ClassifierMixin):
...
-            y (array-like of shape n_samples)
+        self.estimator = estimator
 
-        Returns:
-            self
+    def _more_tags(self):
 
-        """
+        # If converting a BinaryClassifier, notify that only binary classification is supported
+        return dict(binary_only=isinstance(self.estimator, base.BinaryClassifier))
```
MaxHalford, line 211: I'm not sure about this but you can leave it :). It's actually a good idea.

MaxHalford:
No worries @brcharron! I just want to check that `_partial_fit` is necessary instead of `partial_fit`. You've done good work :)

creme/compat/creme_to_sklearn.py
brcharron: It's similar to how it's [done](https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/linear_model/_stochastic_gradient.py#L449) it `sklearn`. `_partial_fit` contains the common denominator to `fit` and `partial_fit`. The current `Creme2SKL*` are simple enough that there is nothing else to put in `partial_fit` but theoretically there could be some specific checks like [that](https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/linear_model/_stochastic_gradient.py#L680). Seems a fair design although I don't love the extra code either. I can put all in `partial_fit` if you prefer, just let me know.

creme/compat/creme_to_sklearn.py
brcharron: That's to pass `sklearn`'s `check_estimator` tests. Otherwise it will pass in some example `y` with more than 2 classes, which may crash the underlying `creme` estimator. Actually it still does due to the bug I reference in the code but still it should be good to conform to `sklearn`'s [guidelines](https://scikit-learn.org/dev/developers/develop.html#estimator-tags).

MaxHalford:
Everything looks good to me! Thank you very much for doing this.

Issue #280, opened 2020-04-04
Implement Cython dict


MaxHalford:
We really want to give a shot at implementing a fast dictionary in Cython. scikit-learn has one [here](https://github.com/scikit-learn/scikit-learn/blob/5e7ff0d0cde7b997ee5bc1300c2dc5bdac4c2608/sklearn/utils/_fast_dict.pyx). This could potentially be very useful for linear models.

MaxHalford:
@brcharron has done a wonderful job at implementing `utils.VectorDict`, which is dictionary-like data structure that has methods implemented in Cython. For instance it implements the `@` operator for doing a dot product. Let's use in the `linear_model` module before closing this issue.

MaxHalford:
`VectorDict` has been successfully integrated into `linear_model`. We will now focus on the optimizers from the `optim` module. We will keep this issue open for potential discussions.

raphaelsty:
@brcharron Nice job, it's really cool to get good performance with dictionaries.

brcharron:
The [`WeightVector`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/_weight_vector.pyx) class in `sklearn` could be used as inspiration for further work, it implements custom functions like `w += c * u` or `w *= c`.

MaxHalford:
I took a go at converting the AdaDelta optimizer and ran into some issues. The current code is:

```
for i, gi in g.items():

    self.g2[i] = self.rho * self.g2[i] + (1 - self.rho) * gi ** 2
    step = - self._rms(self.s2[i]) / self._rms(self.g2[i]) * gi
    self.s2[i] = self.rho * self.s2[i] + (1 - self.rho) * step ** 2
    w[i] += step
```

I wrote the following vector notation:

```
if self.g2 is None:
    self.g2 = base.empty_like(g)
    self.s2 = base.empty_like(g)

self.g2 = self.rho * self.g2 + (1 - self.rho) * g ** 2
step = -self._rms(self.s2) / self._rms(self.g2) * g
self.s2 = self.rho * self.s2 + (1 - self.rho) * step ** 2
w += step
```

The `empty_like` function is defined as so:

```python
def empty_like(x):
    if isinstance(x, utils.VectorDict):
        return utils.VectorDict({})
    return np.empty_like(x)
```

[...]

PR #315, opened 2020-07-09
WIP Cython dict wrapper supporting vector-like operations

brcharron:
For #280.
Still WIP, implements a Cython `dict` wrapper that supports vector-like (as in element of a vector space) operations `+`, `-`, `*`, `/` and `@`. Should be useful in `glm` to store the weights, as a lot of dot products and arithmetic between weights and samples are made. Still need to add tests and integrate in code but maybe you can check if it fits the feature request.

Example of speedup and code readability improvement:
```python
>>> from creme.utils.math import dot
>>> x = {'a': 4, 'b': 2, 'c': 5}
>>> y = {'a': 3, 'b': 9, 'd': 4}
>>> %timeit dot(x, y)
993 ns ± 5.01 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

>>> from creme.utils import VectorDict
>>> vx = VectorDict(x)
>>> vy = VectorDict(y)
>>> assert vx @ vy == dot(x, y)
>>> %timeit vx @ vy
>>> %timeit vx @ y  # compatible with raw dict as well, same speed
151 ns ± 2.55 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
154 ns ± 1.8 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
Note that I started with a C++ `unordered_map` for speed, similar to scikit-learn's one but although it is faster for operations between Cython objects, operations between Cython objects and `dict` (such as `vx @ y` above) are slow. This is because `dict` keys will be typically `str` which need to be encoded into `bytes` to match / store keys in the C++ map as `string`, incurring overhead. Also it restricts keys to strings while the current implementation allows any Python object as key, making it seamless. As the external API of `creme` is based on `dict`, the proposed design seems better. scikit-learn's `IntFloatDict` with a C++ map is a better choice for them as they keep it internal and don't worry about string encoding.

Also note that there is a similar [implementation](https://pypi.org/project/VectorDict/) but it's in Python so too slow.

### DIFF
--- a/creme/utils/__init__.py
+++ b/creme/utils/__init__.py
@@ -5,6 +5,7 @@ from . import pretty
 from .histogram import Histogram
 from .sdft import SDFT
 from .skyline import Skyline
+from .vectordict import VectorDict
 from .window import Window
 from .window import SortedWindow
 
@@ -15,5 +16,6 @@ __all__ = [
     'SDFT',
     'Skyline',
     'SortedWindow',
+    'VectorDict',
     'Window'
 ]
--- a/creme/utils/vectordict.pyx
+++ b/creme/utils/vectordict.pyx
@@ -1,0 +1,256 @@
+cimport cython
+
+cdef VectorDict _add_const(VectorDict res, other):
+    # add the constant other to res then return res
+    for key, value in res._map.items():
+        res._map[key] = value + other
+    return res
+
+cdef VectorDict _add_dict(VectorDict res, dict other):
+    # add the dict other to res then return res
+    for key, value in other.items():
+        res._map[key] = res._map.get(key, 0) + value
+    return res
+
+cdef VectorDict _sub_const(VectorDict res, other, int rev):
+    # subtract the constant other to res then return res
+    if rev:
+        for key, value in res._map.items():
+            res._map[key] = other - value
+    else:
+        for key, value in res._map.items():
+            res._map[key] = value - other
+    return res
+
+cdef VectorDict _sub_dict(VectorDict res, dict other, int rev):
+    # subtract the dict other to res then return res
+    if rev:
+        for key, value in other.items():
+            res._map[key] = value - res._map.get(key, 0)
+    else:
+        for key, value in other.items():
+            res._map[key] = res._map.get(key, 0) - value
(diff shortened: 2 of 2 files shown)


MaxHalford:
Holy moly this is good stuff @brcharron. I don't have the time to take a look at your PR right now but I'm going to do it over the weekend. I actually did some similar stuff [here](https://github.com/creme-ml/workbench/blob/master/cython%20dict.ipynb), so maybe you can check it out. However I can already tell that what you've done is better. Really good work.

MaxHalford:
I gave at comparing your method against a pure Python approach and a version that uses a a C++ `cpp_map` (not `unordered_map`), all the time using integers:

benchmark

Basically your method is good enough. It allows using any type of keys and that's a good thing. I took a good look at the code and I don't see any issues. The code for adding/subtracting/multiplying is a tad complex, but it's fine. Good to go!

brcharron:
Thanks for the quick review! Actually was still WIP I'll make another one with tests :). The code is a bit complex as Cython didn't support reversed operators like `__rsub__` so you need to reverse-engineer the actual expression based on types... The in-place methods like `__isub__` are simpler. Seems to be [fixed](http://docs.cython.org/en/latest/src/userguide/special_methods.html#arithmetic-methods) in Cython 3.0 though but it's just pre-release.

MaxHalford:
Oups yeah tests would be nice. 

Also I was thinking, do you it would be worth using a `collections.defaultdict` instead of a `dict` as the underlying storage? Or maybe `__getitem__` could return 0 in case of a `KeyError`. We could also pass an initialisation function which is called when a `KeyError` occurs.

 There might also be the need for a subtraction which scales the right dict by a scalar, so then we can do something like `scaled_sub(weights, gradient, scale=learning_rate)`. 

Basically I tried using this in the `linear_model` module and I reached some road blocks.

brcharron:
I can make it have the behavior of `defaultdict` yes. I think `0` is a good default given the vector interpretation rather than allowing any constructor.
For the second point, could you give an example of an issue in `linear_model`?

PR #316, opened 2020-07-10
[MRG] Use VectorDict to store weights in GLM

brcharron:
For #280.
Added tests for `VectorDict` and features necessary to its use in `GLM`.

### DIFF
--- a/creme/linear_model/glm.py
+++ b/creme/linear_model/glm.py
@@ -84,7 +84,7 @@ class GLM:
         loss_gradient *= w
         loss_gradient = np.clip(loss_gradient, -self.clip_gradient, self.clip_gradient).item()
 
-        return (loss_gradient * utils.VectorDict(x)  + 2. * self.l2 * self._weights, loss_gradient)
+        return loss_gradient * utils.VectorDict(x)  + 2. * self.l2 * self._weights, loss_gradient
 
     def fit_one(self, x, y, w=1.):
         with self._fit_mode(x):
@@ -93,7 +93,7 @@ class GLM:
     # Mini-batch methods
 
     def _raw_dot_many(self, X: pd.DataFrame) -> np.ndarray:
-        return X.values @ self._weights.to_numpy(X) + self.intercept
+        return X.values @ self._weights.to_numpy(X.columns) + self.intercept
 
     def _eval_gradient_many(self,
                             X: pd.DataFrame,
--- a/creme/linear_model/glm.py
+++ b/creme/linear_model/glm.py
@@ -1,4 +1,4 @@
-import collections
+import contextlib
 import math
 import numbers
 import typing
@@ -36,17 +36,31 @@ class GLM:
             intercept_lr
         )
         self.clip_gradient = clip_gradient
-        self.weights = collections.defaultdict(initializer)
         self.initializer = initializer
+        self._weights = utils.VectorDict(None)
 
         # The predict_many functions are going to return pandas.Series. We can name the series with
         # the name given to the y series seen during the last fit_many call.
         self._y_name = None
 
+    @property
+    def weights(self):
+        return self._weights.to_dict()
+
+    @contextlib.contextmanager
+    def _fit_mode(self, mask=None):
+        weights = self._weights
+        try:
+            # enable the initializer and set a mask
+            self._weights = utils.VectorDict(weights, self.initializer, mask)
+            yield
+        finally:
+            self._weights = weights
+
     def _fit(self, x, y, w, get_grad):
 
         # Some optimizers need to do something before a prediction is made
-        self.weights = self.optimizer.update_before_pred(w=self.weights)
+        self.optimizer.update_before_pred(w=self._weights)
 
         # Calculate the gradient
         gradient, loss_gradient = get_grad(x, y, w)
@@ -55,37 +69,31 @@ class GLM:
         self.intercept -= self.intercept_lr.get(self.optimizer.n_iterations) * loss_gradient
 
         # Update the weights
-        self.weights = self.optimizer.update_after_pred(w=self.weights, g=gradient)
+        self.optimizer.update_after_pred(w=self._weights, g=gradient)
 
         return self
 
     # Single instance methods
 
     def _raw_dot_one(self, x: dict) -> float:
-        return utils.math.dot(self.weights, x) + self.intercept
+        return self._weights @ utils.VectorDict(x) + self.intercept
 
     def _eval_gradient_one(self, x: dict, y: float, w: float) -> (dict, float):
 
         loss_gradient = self.loss.gradient(y_true=y, y_pred=self._raw_dot_one(x))
         loss_gradient *= w
-        loss_gradient = np.clip(loss_gradient, -self.clip_gradient, self.clip_gradient)
+        loss_gradient = np.clip(loss_gradient, -self.clip_gradient, self.clip_gradient).item()
 
-        return (
-            {
-                i: xi * loss_gradient + 2. * self.l2 * self.weights.get(i, 0)
-                for i, xi in x.items()
-            },
-            loss_gradient
-        )
+        return (loss_gradient * utils.VectorDict(x)  + 2. * self.l2 * self._weights, loss_gradient)
 
     def fit_one(self, x, y, w=1.):
-        return self._fit(x, y, w, get_grad=self._eval_gradient_one)
+        with self._fit_mode(x):
+            return self._fit(x, y, w, get_grad=self._eval_gradient_one)
 
     # Mini-batch methods
 
     def _raw_dot_many(self, X: pd.DataFrame) -> np.ndarray:
-        weights = np.array([self.weights[c] for c in X.columns])
-        return X.values @ weights + self.intercept
+        return X.values @ self._weights.to_numpy(X) + self.intercept
 
     def _eval_gradient_many(self,
                             X: pd.DataFrame,
@@ -107,7 +115,8 @@ class GLM:
 
     def fit_many(self, X: pd.DataFrame, y: pd.Series, w: typing.Union[float, pd.Series] = 1):
         self._y_name = y.name
-        return self._fit(X, y, w, get_grad=self._eval_gradient_many)
+        with self._fit_mode(set(X)):
+            return self._fit(X, y, w, get_grad=self._eval_gradient_many)
 
 
 class LinearRegression(GLM, base.Regressor):
@@ -132,7 +141,7 @@ class LinearRegression(GLM, base.Regressor):
         initializer: Weights initialization scheme.
 
     Attributes:
-        weights (collections.defaultdict): The current weights.
+        weights (dict): The current weights.
 
     Example:
 
@@ -241,8 +250,8 @@ class LinearRegression(GLM, base.Regressor):
 
         names = list(map(str, x.keys())) + ['Intercept']
         values = list(map(fmt_float, list(x.values()) + [1]))
-        weights = list(map(fmt_float, [self.weights.get(i, 0) for i in x] + [self.intercept]))
-        contributions = [xi * self.weights.get(i, 0) for i, xi in x.items()] + [self.intercept]
+        weights = list(map(fmt_float, [self._weights.get(i, 0) for i in x] + [self.intercept]))
+        contributions = [xi * self._weights.get(i, 0) for i, xi in x.items()] + [self.intercept]
         order = reversed(np.argsort(contributions))
         contributions = list(map(fmt_float, contributions))
 
@@ -276,7 +285,7 @@ class LogisticRegression(GLM, base.Classifier, base.MiniBatchClassifier):
         initializer: Weights initialization scheme.
 
     Attributes:
-        weights (collections.defaultdict): The current weights.
+        weights (dict): The current weights.
 
     Example:
 
--- a/creme/linear_model/glm.py
+++ b/creme/linear_model/glm.py
@@ -82,7 +82,7 @@ class GLM:
 
         loss_gradient = self.loss.gradient(y_true=y, y_pred=self._raw_dot_one(x))
         loss_gradient *= w
-        loss_gradient = np.clip(loss_gradient, -self.clip_gradient, self.clip_gradient).item()
+        loss_gradient = float(utils.math.clamp(loss_gradient, -self.clip_gradient, self.clip_gradient))
 
         return loss_gradient * utils.VectorDict(x)  + 2. * self.l2 * self._weights, loss_gradient
 
--- a/creme/optim/sgd.py
+++ b/creme/optim/sgd.py
@@ -1,3 +1,5 @@
+from creme import utils
+
 from . import base
 
 
@@ -40,7 +42,10 @@ class SGD(base.Optimizer):
 
     def _update_after_pred(self, w, g):
 
-        for i, gi in g.items():
-            w[i] -= self.learning_rate * gi
+        if isinstance(w, utils.VectorDict) and isinstance(g, utils.VectorDict):
+            w -= self.learning_rate * g
+        else:
+            for i, gi in g.items():
+                w[i] -= self.learning_rate * gi
 
         return w
--- a/creme/utils/test_vectordict.py
+++ b/creme/utils/test_vectordict.py
@@ -1,0 +1,58 @@
+from creme.utils import VectorDict
+
+
+def test_vectordict():
+
+    x = {'a': 8, 'b': -1.2, 4: 2.7}
+    vx = VectorDict(x)
+    assert isinstance(vx, dict)
+    assert vx == x
+    assert vx['a'] == 8
+    assert vx[4] == 2.7
+    assert vx[()] == 0
+    vx[9] = 8.9
+    assert vx[9] == 8.9
+
+    y = dict()
+    vy = VectorDict()
+    assert vy == y
+
+    z1 = {'a': 1, 'b': -5, 'c': -3}
+    z2 = {'a': 2, 'b': 0.5, 'd': 4}
+    vz1 = VectorDict(z1)
+    vz2 = VectorDict(z2)
+    assert vz1 == vz1
+    assert +vz1 == vz1
+    assert -vz1 == {'a': -1, 'b': 5, 'c': 3}
+    assert vz1 + 2 == 2 + vz1 == {'a': 3, 'b': -3, 'c': -1}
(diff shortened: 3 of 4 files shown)


MaxHalford:
If `VectorDict` inherits from `dict`, we don't need to define pass-through methods, because they're inherited, right?

brcharron:
The `dict` inheritance is just to have `issubclass(VectorDict, dict)`, it doesn't really inherit anything from `dict`, everything is overridden, but it can be considered like a `dict`.
I also tried to just use the fact that by inheritance `VectorDict` already has everything and the code is simpler but the operations are slower, especially for small-ish sizes. That's because on the Cython side it is considered a Python object while operations on `dict` are optimized so you either go through the Python layer or do some casting back and forth between `VectorDict` and `dict` which both induce some overhead. Storing the content in `_map`, it's already a `dict` and can be used directly.
Not a huge difference though and it's an implementation detail that can be changed later if something faster comes up.

brcharron:
I pushed a quick commit using `VectorDict` for the weights in GLM. I removed the `initializer` option so it only supports `Zeros` now. I will put it back once I understand how it was meant.

`defaultdict` calls the initializer if `__getitem__` is called, like `d[3]`, not if `get` is called, like `d.get(3)`.
From what I understood of the code, `_raw_dot_one` does not induce weight initialization but `_raw_dot_many` does, so the intended behavior is not too clear. Then gradient calculation does not induce initialization but weight update does, at least in `SGD`.

When should the weight initialization occur? An option would be as soon as a new key is observed, except in prediction methods. The distinction makes things a bit more complicated. Or maybe initialization by zero is fine for `GLM`.

creme/linear_model/glm.py
```diff
@@ -62,21 +60,16 @@ def _fit(self, x, y, w, get_grad):
     # Single instance methods
 
     def _raw_dot_one(self, x: dict) -> float:
-        return utils.math.dot(self.weights, x) + self.intercept
+        return self.weights @ x + self.intercept
 
     def _eval_gradient_one(self, x: dict, y: float, w: float) -> (dict, float):
 
+        x = utils.VectorDict(x)
```
brcharron, line 67: Not awesome but just to show the limits of the current implementation.
The conversion is not needed for `_raw_dot_one(x)` but it is for `x * loss_gradient`.
Conversions induce a copy, same as any addition or multiplication that is not in-place.
On the other hand the dict comprehension does not make any copy except the creation of the final output.

brcharron: An alternative would be to make conversions and operations lazy, like a computation graph, but it would be quite a bit more complex. For example `(loss_gradient * VectorDict(x) + 2. * self.l2 * self.weights).eval()`.

MaxHalford:
The initializer is meant to be called whenever a non-existing weight is accessed. The fact that `_raw_dot_one` does not call the initializer **is a bug**, and therefore needs fixing. I think it's fine to initialize the weight during a prediction. Does that make things clearer for you?

creme/linear_model/glm.py
```diff
@@ -129,10 +122,9 @@ class LinearRegression(GLM, base.Regressor):
             will be not be updated.
         l2: Amount of L2 regularization used to push weights towards 0.
         clip_gradient: Clips the absolute value of each gradient value.
-        initializer: Weights initialization scheme.
 
     Attributes:
-        weights (collections.defaultdict): The current weights.
+        weights (utils.VectorDict): The current weights.
```
MaxHalford, line 127: I'm not sure, but maybe that `weights` should be a computed property, whilst the `VectorDict` should be stored in a `_weights` attribute. What do you thing?

creme/linear_model/glm.py
MaxHalford: I guess that we can do the conversion for the moment. The lazy computation graph is interesting, and could be considered as a separate project.

brcharron:
> The initializer is meant to be called whenever a non-existing weight is accessed. The fact that `_raw_dot_one` does not call the initializer **is a bug**, and therefore needs fixing. I think it's fine to initialize the weight during a prediction. Does that make things clearer for you?

It's clear for use in fit mode, a weight is initialized as soon as a sample is provided with a new dimension. For prediction I think that may not be intuitive as it would change the state of the estimator while one may think that a prediction leaves the estimator unchanged. In particular if a random number generator is used for initialization, it may lead to strange things.

Another clarification is about weights penalty, [currently](https://github.com/creme-ml/creme/blob/master/creme/linear_model/glm.py#L75) there is no gradient for the weights on dimensions not present in the sample. Is that intentional? A vector addition as implemented in `VectorDict` would put a non-zero gradient due to the penalty, which would shrink the weight.

creme/linear_model/glm.py
brcharron: Yes that seems better.

MaxHalford:
> For prediction I think that may not be intuitive as it would change the state of the estimator while one may think that a prediction leaves the estimator unchanged. In particular if a random number generator is used for initialization, it may lead to strange things.

Yes, you're right.

> Another clarification is about weights penalty, currently there is no gradient for the weights on dimensions not present in the sample. Is that intentional?

I'm pretty sure that's intentional.

> A vector addition as implemented in VectorDict would put a non-zero gradient due to the penalty, which would shrink the weight.

So we need to operate only on the keys that are in `x`, right? This is getting a bit tricky as it requires specialised functions. To be honest, I'm actually happy removing L2 regularisation for now. I didn't implement in the mini-batch version for lack of time. Also, with @gbolmier we want to design a friendly way to apply arbitrary regularisation in a principled manner. Therefore, maybe that we can remove L2 regularisation, and add it as a to-do issue.

brcharron:
I implemented the lazy version [here](https://colab.research.google.com/drive/1uRwz42PG-WmgAzKDLmrEnuhYxgCTZW1d?usp=sharing). After a bunch of testing I am not convinced that a class-based approach like that or `VectorDict` would beat (in speed) a few custom Cython functions for the usual expressions, like a dot product or `a + b * u + c * v` with `u` and `v` `dict`s and the rest `float`s, or `u += a + b * v`, etc.
Besides pure speed though, the case may be made that it improves code readability, and as you mention it could be used to share the same code for `ndarray`s and `dict`.

brcharron:
> So we need to operate only on the keys that are in `x`, right?

Yes if that's the desired behavior.
Not sure how you see it but doing it like now is like projecting the model on the dimensions of the sample and taking a gradient step there. Using the vector addition is like first expanding the model to the union of the existing dimensions and those of the sample, with initialization if needed, then taking a gradient step. That union now becomes the domain of the estimator.

MaxHalford:
> Not sure how you see it but doing it like now is like projecting the model on the dimensions of the sample and taking a gradient step there. Using the vector addition is like first expanding the model to the union of the existing dimensions and those of the sample, with initialization if needed, then taking a gradient step. That union now becomes the domain of the estimator.

Yes, I understand. However, I'm pretty sure that we only want to operate on `x`'s keys.

Your lazy evaluation PoC is quite impressive! I like the class based approach of `VectorDict`; it allows writing code that works both for 1D numpy arrays and dicts. We can also imagine having a small `*.pyx` file in the `linear_model` module that does certain things in particular, such as `a + b * u + c * v`. It's fine to not make everything generic.

brcharron:
Got it, seems like `VectorDict` would be useful then.
I am working on a more featureful version that can support things like `defaultdict`-style factories or quick fill by zero, masking by another `dict` to project on a sub-space and lazy init for speed, would look like (if all goes well):
```python
cdef class VectorDict:
    def __init__(self, data=None, default_value=0, default_factory=None,
                 mask=None, copy=False):
        """A dictionary-like object that supports vector-like operations.

        Supports addition and subtraction with a VectorDict or a constant.
        Supports multiplication and division by a constant.
        Supports dot product with a VectorDict through the @ operator.

        If default_factory is not None, it is called whenever a missing key is
        accessed, either directly or in an operation such as addition, and the
        result is inserted as the value for that key.
        If default_factory is None but not default_value, default_value is used
        whenever a missing key is accessed, but it is not inserted for that key.
        If default_value and default_factory are None, a KeyError exception is
        raised whenever a missing key is accessed.
        If mask is not None, the behavior above only applies to keys which are
        present in mask. For other keys, the behavior is the same as a dict.

        Parameters:
            data: a VectorDict or dict to initialize key-values from, or None
            default_value: an object that supports the arithmetic operations,
                or None
            default_factory: a callable returning an object that supports the
                arithmetic operations, or None
            mask: a VectorDict or dict such that keys not in mask will
                not be considered in operations and will always result in a
                KeyError if accessed by __getitem__, or None
            copy: if data and/or mask are specified, whether to store a copy of
                the underlying dictionaries or references at initialization
        """
```

MaxHalford:
That sounds promising! Eventually you could consider that `default_value` is always equal to 0. I can't think of any cases where it would be different.

brcharron:
I was thinking it can be `None` to force an exception when doing operations on vectors in different spaces, which could be useful in some cases. Also could be `0` or `0.0` or `Fraction(0)`, etc depending on the vector components types but actually `0` would work fine in all cases. Could make it a boolean then.

brcharron:
I integrated `VectorDict` in `GLM` based on the previous discussion. Update of the optimizers for the new inline operators should probably be in a separate issue as it may impact a lot of the code.

creme/linear_model/glm.py
```diff
@@ -55,37 +69,31 @@ def _fit(self, x, y, w, get_grad):
...
-        return self._fit(x, y, w, get_grad=self._eval_gradient_one)
+        with self._fit_mode(x):
+            return self._fit(x, y, w, get_grad=self._eval_gradient_one)
 
     # Mini-batch methods
 
     def _raw_dot_many(self, X: pd.DataFrame) -> np.ndarray:
-        weights = np.array([self.weights[c] for c in X.columns])
-        return X.values @ weights + self.intercept
+        return X.values @ self._weights.to_numpy(X) + self.intercept
```
MaxHalford, line 96: I think `to_numpy(X.columns)` would be more explicit, but whatever.

creme/linear_model/glm.py
```diff
@@ -55,37 +69,31 @@ def _fit(self, x, y, w, get_grad):
...
+        loss_gradient = np.clip(loss_gradient, -self.clip_gradient, self.clip_gradient).item()
 
-        return (
-            {
-                i: xi * loss_gradient + 2. * self.l2 * self.weights.get(i, 0)
-                for i, xi in x.items()
-            },
-            loss_gradient
-        )
+        return (loss_gradient * utils.VectorDict(x)  + 2. * self.l2 * self._weights, loss_gradient)
```
MaxHalford, line 87: No need for parenthesis. 

I assume that we could write a bespoke Cython function to do this in one single loop?

brcharron:
Good catch on the context manager, I didn't know it was so slow.

First here are some overall timings in `master`
```python
>>> from creme.linear_model import LinearRegression
>>> lr = LinearRegression()
>>> x = {i: i for i in range(n)}
>>> %timeit lr.fit_one(x, 1)
# with n = 10
38 µs ± 417 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
# with n = 100
185 µs ± 271 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
and on this branch
```python
# with n = 10
34.6 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
# with n = 100
109 µs ± 105 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
so not much for small spaces but some decent speed up for larger ones.
Note that I didn't particularly try to squeeze every nanosecond as it was already not so easy to implement, can do another pass for performance later.

Now the context manager overhead
```python
>>> %timeit with lr._fit_mode(x): pass
2.18 µs ± 9.26 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>>  %timeit utils.VectorDict(lr._weights, lr.initializer, x)
198 ns ± 0.562 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
is quite significant compared to the actual operation involved (which is fast since there is no copy involved) but I would say not a big issue compared to the overall runtime of the function call and it is quite convenient I think.

creme/linear_model/glm.py
brcharron: Yes could do that, should I add to this PR? As mentioned in another comment there may be several points to optimize, would need to profile to check where to focus.

brcharron:
Committed some small fixes after profiling, here are the timings now in this branch
```python
>>> from creme.linear_model import LinearRegression
>>> lr = LinearRegression()
>>> x = {i: i for i in range(n)}
>>> %timeit lr.fit_one(x, 1)
# with n = 10
14.8 µs ± 37.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# with n = 100
69.7 µs ± 118 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
so about 2.6x speedups in both cases compared to `master`.
Maybe another pass of profiling can be done after optimizers are updated and a better idea of the performance impact of `VectorDict` is known.

brcharron:
And some timings for prediction.
In `master`:
```python
>>> from creme.linear_model import LinearRegression
>>> lr = LinearRegression()
>>> x = {i: i for i in range(n)}
>>> lr.fit_one(x, 1)
>>> %timeit lr.predict_one(x)
# with n = 10
5.89 µs ± 38.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# with n = 100
43.2 µs ± 417 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
and in this branch
```python
>>> from creme.linear_model import LinearRegression
>>> lr = LinearRegression()
>>> x = {i: i for i in range(n)}
>>> lr.fit_one(x, 1)
>>> %timeit lr.predict_one(x)
# with n = 10
2.05 µs ± 14.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# with n = 100
11.1 µs ± 50.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
so 2.9x and 3.9x speedups respectively.

MaxHalford:
Good job! Sorry for the delayed answer, it's the summer and I'm not available as during the rest of the year. In my opinion the speedups are good enough for justifying a merge. For the optimisers, I think we should assume that the inputs are always `VectorDict`s and not `dict`s. This might take some work but it's well worth it, and as we said opens up the path to using numpy arrays as well as `VectorDict`s.

Do you think this is ready to be merged or do you want to do other stuff?

brcharron:
No problem, same here :).
I think it's ready to be merged now, it seems to do the job for `GLM`. Once it's integrated in the rest of the code, let's see if some design or performance bottleneck emerges and we can revisit it.

gbolmier:
Thx for the nice work @brcharron! I started working on vectorizing the optimizers and I noticed that the perfomance boost you observed doesn't come from `VectorDict`, but from other optimizations you did in `linear_model.GLM`.

Mostly from not using numpy to clip gradients:

https://github.com/creme-ml/creme/blob/3b397d6aafdfad48dbb36813c90ed3fd1844fc30/creme/linear_model/glm.py#L84

Here is a quick benchmark of `VectorDict` against standard python dicts:

```python
from creme import utils

xx = {i: i for i in range(1_000)}
y = utils.VectorDict(x)
yy = utils.VectorDict(xx)

%timeit utils.math.dot(x, x)
# 10.6 µs ± 109 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit utils.math.dot(xx, x)
# 11 µs ± 125 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit y @ y
# 11.6 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit yy @ y
# 83.8 µs ± 17.2 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

brcharron:
@gbolmier Indeed as mentioned above I only profiled the `GLM` part but there is definitely still room for performance improvements. `VectorDict` became a bit bloated since the first version to accommodate the different use cases (`default_factory`, `mask`, `copy`) so not every pathway is optimized. For example `yy @ y` is slow because it uses the same code whether `y` has a default factory or not so it loops on the union of keys while `utils.math.dot` only considers the intersection. If this is a bottleneck somewhere, I can write a faster pathway for when `y` has no default factory. And similar optimizations for other operations.

gbolmier:
Thx for the prompt reply Bruno, makes more sense to me.

> there is definitely still room for performance improvements

You mean by getting rid of the extra overhead or doing better than your initial implementation? :yum:

> benchmark

brcharron:
Depends on the situation. Did you identify a use case where a `VectorDict` operation is the bottleneck?

gbolmier:
Models dealing with a fraction of the features at each iteration — factorization models for instance — suffer a lot from looping on the union of the keys instead of the intersection.

With the current `VectorDict` genericity, a standard dot product isn't faster than with python dicts. That's something we need to optimize as much as possible.

Also, I'm not sure anyone pointed it, but `VectorDict` data structure significantly reduce memory storage, which is rather useful for large parameterized linear models.

brcharron:
@gbolmier Thanks for additional details and sorry for the delay.
I created #349 that should help for the dot product.
I'm not quite familiar with the operations used in factorization models, is there some concern for other operations like addition, etc? I can update the PR with those.
And good find for the memory storage, didn't know that.

MaxHalford:
@brcharron thanks for that. Just popping by to say to nobody that commits to an opensource has to say sorry, even though you're just being polite :). I want everyone to have a good time contributing to the project and feel no pressure whatsoever.

brcharron:
Thanks for the kind words @MaxHalford :)

gbolmier:
> @gbolmier Thanks for additional details and sorry for the delay.
> I created #349 that should help for the dot product.

Thank you for the awesome work you did! I commented your work in the concerned thread. No need to say that I completely agree with what Max said above.

> I'm not quite familiar with the operations used in factorization models, is there some concern for other operations like addition, etc? I can update the PR with those.

No concerns about separated operations, one about single statements including multiple operations. It could slow down gradient computations, for instance in `glm`:
 https://github.com/creme-ml/creme/blob/3b397d6aafdfad48dbb36813c90ed3fd1844fc30/creme/linear_model/glm.py#L86

Not sure that would make a difference though

brcharron:
@gbolmier Got it, I will implement the lazy scaling and see if there is a significant difference in this kind of expressions.

brcharron:
Implemented in #353 but the speedups did not seem to be worth the added complexity.

PR #318, opened 2020-07-28
Add several operations to VectorDict

brcharron:
`*` and `/` between two VectorDict
`abs`, `min`, `max`, `minimum`, `maximum`

Uses union of of keys for now in `*` and `/`, see discussion in #280.

### DIFF
--- a/creme/utils/test_vectordict.py
+++ b/creme/utils/test_vectordict.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 
 from creme.utils import VectorDict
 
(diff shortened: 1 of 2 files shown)


MaxHalford:
Good job, that was quick. Can you do `pow` too? There's some quirk about the optional modulo argument, I didn't understand it fully.

brcharron:
`pow` is implemented in `__pow__` and `__ipow__`, there are tests also with `**`.
For the modulo I didn't implement, I don't think it's needed here, see [here](https://stackoverflow.com/a/58518404/1061133).

PR #319, opened 2020-07-29
Use Cython to boost performance of metrics

jacobmontiel:
Performance metrics calculation is an expensive operation when performed on single samples instead of batches. A known bottleneck is the extensive usage of `for` loops.

This PR focuses on optimizing parts of the metrics calculations by using Cython implementations.

## Changes required

- [x] Implement a confusion matrix that supports both insertion and deletion (binary and multi-class)
- [x] Implement muti-label confusion matrix
- [ ] ~Refactor the "classification report" so a single instance of the confusion matrix is updated incrementally and used to calculate multiple metrics on demand.~
- [x] Replace python implementation where possible

## Metrics

- [x] accuracy
- [x] kappa
- [x] kappa-t
- [x] kappa-m
- [x] precision (micro, macro, weighted, example-based)
- [x] recall (micro, macro, weighted, example-based)
- [x] f1 (micro, macro, weighted, example-based)
- [x] fbeta (micro, macro, weighted, example-based)
- [x] geometric mean
- [x] roc auc
---
- [x] Hamming score 
- [x] exact match
- [x] Jaccard score


jacobmontiel:
@MaxHalford @smastelini This is an ongoing effort. So far I focused on:

- Fix installation as the subtree uses a different approach to process Cython files
- Modify the existing Cython implementation of the confusion matrix for binary and multi-class to be closer to the python implementation
  - Support strings
  - Use `dict` of `dict` to access via `[][]`
  - Provide a similar interface including the `revert` method
  - I did not extend `MultiClassMetric` as I am not familiar with it. This is relevant depending on how we intend to use the confusion matrix, especially within the report.
- I also temporarily brought the `ClassificationPerformanceEvaluator` so we have all the code in a single place. The new "classification report" will likely be a combination of it and the existing code.
- If this looks good then I will proceed to do the modifications in the multi-label version.

setup.py
```diff
@@ -104,6 +112,18 @@
     ],
     ext_modules=cythonize(
         module_list=[
+            setuptools.Extension(
```
MaxHalford, line 115: Is this necessary? Doesn't the `sources=['**/*.pyx']` underneath include everything?

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
...
+                width=width
+            )
+            for y_true in classes
+        ))
+
+        return table
+
+
+cdef class MultiLabelConfusionMatrix:
+    """ Multi-label Confusion Matrix.
```
MaxHalford, line 113: Extra space.

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
...
+                *[f'{self.data[y_true][y_pred]:{self._fmt}}' for y_pred in classes],
+                width=width
+            )
+            for y_true in classes
+        ))
+
+        return table
+
+
+cdef class MultiLabelConfusionMatrix:
```
MaxHalford, line 112: Just thinking out loud, isn't `MultiLabelConfusionMatrix` essentially a `dict` of `ConfusionMatrix`?

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
+import warnings
+
+import numpy as np
+cimport numpy as np
+
+from collections import defaultdict
+
+cdef bint boolean_variable = True
```
MaxHalford, line 8: I can't this being used anywhere :)

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
...
+        y_true, y_pred = entry
+        self.update(y_true, y_pred, sample_weight)
+
+    def __getitem__(self, key):
+        """Syntactic sugar for accessing the counts directly."""
+        return self.data[key]
+
+    def update(self, y_true, y_pred, sample_weight=1.):
+        if sample_weight is None:
+            # Since we ca not set a default value in the signature
```
MaxHalford, line 53: cannot

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
...
+
+    def __init__(self, classes=None):
+        self._init_classes = set(classes) if classes is not None else set()
+        self.classes = self._init_classes
+        self.sum_diag = 0.0
+        self.sum_row = defaultdict(float)
+        self.sum_col = defaultdict(float)
+        self.data = defaultdict(lambda: defaultdict(float))
+
+    def __setitem__(self, entry, double sample_weight):
```
MaxHalford, line 41: I don't mind having this, but I can't think of a usecase for it.

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,232 @@
+import warnings
+
+import numpy as np
+cimport numpy as np
+
+from collections import defaultdict
+
+cdef bint boolean_variable = True
+
+cdef class ConfusionMatrix:
```
MaxHalford, line 10: How about if we move the methods from `ClassificationPerformanceEvaluator` into this class?

setup.py
jacobmontiel: This is required as the files use numpy arrays. In this case, we must indicate the path to `numpy` which is done via the `get_include` method.

Alternatively, we can define the numpy path as a generic option, but I am not sure about the impact (if any) on cython code that does not use `numpy`.

creme/metrics/_confusion_matrix.pyx
jacobmontiel: Definetly, I will explore this option.

creme/metrics/_confusion_matrix.pyx
jacobmontiel: It is possible to see it that way. However, many operations can be easily performed by using a 3D array:

```python
sum_tp = data[:,1,1]
```

setup.py
jacobmontiel: I simplified this in a61a9597c360687d911e4a90eaded8f8088b06cd by setting the path to numpy by default for all cython modules.

creme/metrics/_confusion_matrix.pyx
jacobmontiel: It is used here :)

https://github.com/creme-ml/creme/blob/a61a9597c360687d911e4a90eaded8f8088b06cd/creme/metrics/_confusion_matrix.pyx#L168

MaxHalford: Lol sorry I don't understand: `boolean_variable` is the name of the boolean variable that is declared, but I don't see any reference to `boolean_variable` in the rest of the file.

creme/metrics/report.py
```diff
@@ -16,109 +17,71 @@ class ClassificationReport(base.MultiClassMetric):
...
-           Micro       0.400    0.400   0.400
-        Weighted       0.600    0.400   0.480
-        <BLANKLINE>
-                         40.0% accuracy
+    >>> print(report)
+               Precision   Recall   F1      Support
+    <BLANKLINE>
+       apple       0.000    0.000   0.000       1.0
+      banana       0.000    0.000   0.000       3.0
+        pear       0.000    0.000   0.000       1.0
```
smastelini, line 43: Isn't something odd with those results?

MaxHalford: Yes indeed... I'll look into it!

creme/metrics/_confusion_matrix.pyx
jacobmontiel: Removed in a61a9597c360687d911e4a90eaded8f8088b06cd

creme/metrics/_confusion_matrix.pyx
jacobmontiel: My bad, you are right, that was completely unrelated and unused.

creme/metrics/_confusion_matrix.pyx
jacobmontiel: Fixed typo in a61a9597c360687d911e4a90eaded8f8088b06cd

creme/metrics/_confusion_matrix.pyx
```diff
@@ -0,0 +1,286 @@
...
+            )
+            for y_true in classes
+        ))
+
+        return table
+
+    @property
+    def majority_class(self):
+        try:
+            return self._class_counter.most_common(1)[0][0]
```
MaxHalford, line 146: We can also use the `sum_row` attribute, right? That way we don't need to create and update `_class_counter`.

jacobmontiel: That was the original idea. I used the counter to avoid the loop to get the majority class. However, since we are using Cython this shouldn't be a big issue. Now we are using `sum_row`

jacobmontiel:
> Wouldn't it be better to put this logic in the `revert` method? Then we wouldn't need to check if `sample_weight` is negative.

> I would return None :)

I am not sure about what these comments refer to

MaxHalford:
Sorry @jacobmontiel my comments got lost in the changes :)

jacobmontiel:
No problem @MaxHalford 

I addressed the comment regarding the negative `sample_weight`.

Regarding the usage of `NoChangeClassifier` and `PriorClassifier` seems an overhead as we still need to track the observed weight. The code is a pseudo-classifier only implementing the core elements with the information already available in the confusion matrix. But I am biased toward the current approach. What I like in the current implementation is that any update to the confusion matrix will keep statistics updated for all related metrics. We can also open a follow-up PR to investigate how to improve in this aspect.

PR #321, opened 2020-08-04
Add data conversion utilities

smastelini:
Add two simple data conversion utility functions:

* `numpy2dict`: convert an one-dimensional numpy array to a dictionary
* `dict2numpy`: convert a dictionary containing only numerical data as its values to an one-dimensional numpy array

Note that the building process is being fixed in #319. Once this PR is merged, I should do a rebase to make sure everything is working as intended.

### DIFF
--- a/creme/utils/data_conversion.py
+++ b/creme/utils/data_conversion.py
@@ -1,0 +1,53 @@
+import numpy as np
+
+
+def dict2numpy(data):
+    """Convert a dictionary containing data to a numpy.array
+
+    Parameters
+    ----------
+    data : dict
+        A dictionary whose keys represent input attributes and the values
+        represent their observed contents.
+
+    Returns
+    -------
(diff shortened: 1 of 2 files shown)


creme/utils/data_conversion.py
```diff
@@ -40,8 +40,8 @@ def numpy2dict(data):
     Returns
     -------
         dict
-            A dictionary where keys are integers :math: `k \in \{0, 1, ..., |\mathtt{data}| - 1\}`,
-            and the values are each one of the :math: `k` entries in `data`.
+            A dictionary where keys are integers $k \in \{0, 1, ..., |\text{data}| - 1\}$,
```
MaxHalford, line 43: Much more fun, right? :)

smastelini: For sure! The dream come true :D

PR #323, opened 2020-08-05
[skmultiflow->creme] Merge k-NN algorithms

smastelini:
This is the beginning of the efforts to merge the `lazy` module of `skmultiflow` to `creme.neighbors`.

This PR depends on #319 and #321, so I am making it a draft. The idea is to get some feedback and plan forward our next steps.

Here is the checklist:

- [x] Update basic documentation
- [x] Use creme's base classes in the estimators
- [x] Make algorithms work in a single-instance basis
- [x] Handle data conversion
- [x] Test the algorithms
- [x] Update the examples in the docstring

The kNN algorithms rely on some utility functions from `skmultiflow.utils.utils.py`. That is why I brought them to `creme.utils.skmultiflow_utils.py`


creme/neighbors/base_neighbors.py
```diff
@@ -0,0 +1,156 @@
+from collections import deque
+
+import numpy as np
+from sklearn.neighbors import KDTree
+
+
+from creme.utils.skmultiflow_utils import get_dimensions
+
+
+class SlidingWindow:
```
MaxHalford, line 10: There's `utils.Window` in `creme` that is slightly more generic, maybe you can look into using that? Instead of storing two queues you would store pairs of `(x, y)`.

creme/neighbors/base_neighbors.py
```diff
@@ -0,0 +1,156 @@
+from collections import deque
+
+import numpy as np
+from sklearn.neighbors import KDTree
```
MaxHalford, line 4: Would it be possible to use the KD-tree from `scipy`? I would prefer not having `sklearn` as a dependency, if possible :)

creme/neighbors/base_neighbors.py
smastelini: @jacobmontiel also suggested me to use `utils.Window` to keep track of `y`. I just did not do it yet because the structure does not currently implement some operations I need (e.g., `popleft`), but I can easily extend it. What do you think @MaxHalford?

We are keeping separate structures for `X` and `y` to avoid creating a `np.ndarray` from a queue every time, though the performed array stacking operations are also expensive. This, indeed, is a temporary solution. In the long run we should explore online learning-friendly approaches, such as Locality Sensitive Hashing.

creme/neighbors/base_neighbors.py
smastelini: That's a good point. Until yesterday I didn't know scipy also had its own implementation. I'll check that. I agree with you about depending on `sklearn`: as far as I am concerned we use it in too few locations to make it a dependency.

Any suggestions here @jacobmontiel?

jacobmontiel: Also worth considering, there is a Cython implementation in `scipy` (the implementation in `sklearn` is also in Cython):

https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html#scipy.spatial.cKDTree

Switching implementations seems trivial as the interface is quite similar. However, the `sklearn` implementation seems to provide more features, for example it supports multiple distance metrics which is nice for the user.

creme/neighbors/base_neighbors.py
smastelini: In the latest commit I changed the deque to `utils.Window`.

smastelini:
Hi @MaxHalford and @jacobmontiel. I changed the usage of `collections.deque` to `utils.Window`.

Besides that, I created a circular buffer logic that eliminates the need to perform (the costly) `np.delete` and `np.vstack` operations. I still have to do some testing with it. But the basic idea is to manually circle in the matrix an index that indicates where new observations should be inserted and also keep a binary instance mask to filter/select the filled slots.

I believe we should get some nice speedups since numpy does not need to keep reallocating a matrix `(window_size, n_features)`  every time an insertion takes place.

creme/neighbors/base_neighbors.py
smastelini: I did some reading earlier today on this subject. The `cKDTree` seems to only support the *Minkowski* distance and it seems good enough to me as we can get *Manhattan*, *Euclidean* and so on.

But one might want to experiment with different distances. What do you guys think?

creme/neighbors/base_neighbors.py
smastelini: I ended up using a simple `list()` to keep track of `y`, so that I can use the same circular insertion strategy.

creme/neighbors/base_neighbors.py
smastelini: Now the algorithms are using the scipy version of KDTree (cKDTree).

MaxHalford:
Good progress @smastelini. When you're done you can remove `KNeighborsClassifier` and `KNeighborsRegressor`. We'll have to benchmark what's the fastest way to proceed when `dict`s are used natively. Note that we could also use `utils.VectorDict` to speed things up.

smastelini:
Hi @MaxHalford and @jacobmontiel.

I recently updated the KNNRegressor. My next step is going to check the requirements to bring the ADWIN-based k-NN, which is implemented in skmultiflow. It might be worth checking if we want to make more cKDTree's parameters accessible for the user too.

jacobmontiel:
> My next step is going to check the requirements to bring the ADWIN-based k-NN, which is implemented in skmultiflow. 

This should be done in another PR as I am in the middle of porting ADWIN and the other drift detection methods. As long as we have the basic classes we can move forward.

> It might be worth checking if we want to make more cKDTree's parameters accessible for the user too.

+1 on the idea of giving the user this flexibility.

creme/neighbors/__init__.py
```diff
@@ -1,9 +1,13 @@
 """Neighbors-based learning."""
 from .knn import KNeighborsClassifier
 from .knn import KNeighborsRegressor
```
MaxHalford, line 3: I guess you could remove these classes in your PR :)

creme/neighbors/base_neighbors.py
```diff
@@ -0,0 +1,187 @@
...
+        self._next_insert = 0
+        self._oldest = 0
+        self._imask = None
+        self._X = None
+        self._y = None
+        self._is_initialized = False
+
+        return self
+
+    def add_one(self, x, y):
```
MaxHalford, line 68: I would call this `append` to follow the `collections.deque` terminology.

creme/neighbors/base_neighbors.py
```diff
@@ -0,0 +1,187 @@
...
+        self._imask[self._next_insert] = True  # Mark slot as filled
+        self._next_insert = self._next_insert + 1 if self._next_insert < self.window_size - 1 \
+            else 0
+
+        if slot_replaced:  # The oldest sample was replaced (complete cycle in the buffer)
+            self._oldest = self._next_insert
+        else:  # Actual buffer increased
+            self._size += 1
+
+    def remove_one(self, rem_oldest=True):
```
MaxHalford, line 111: I would call this `popleft` to follow the `collections.deque` terminology.

creme/neighbors/knn_regressor.py
```diff
@@ -0,0 +1,158 @@
...
+                         p=p)
+        if aggregation_method not in {self._MEAN, self._MEDIAN, self._WEIGHTED_MEAN}:
+            raise ValueError('Invalid aggregation_method: {}.\n'
+                             'Valid options are: {}'.format(aggregation_method,
+                                                            {self._MEAN, self._MEDIAN,
+                                                             self.f_WEIGHTED_MEAN}))
+        self.aggregation_method = aggregation_method
+
+    def learn_one(self, x: dict, y: base.typing.RegTarget) -> 'Regressor':
+        """Fits to a set of features ``x`` and a real-valued target ``y``.
```
MaxHalford, line 90: You can write this as ``x`` and ``y``, no need for two back ticks.

creme/neighbors/__init__.py
smastelini: I'll do that as we progress with the PR. Currently I am using them for comparison and reference :D

I created a notebook comparing the `dict`-based implementation against the numpy+ckDTree one. So far we are getting a speed up of about 9x, depending on the window size!

creme/neighbors/knn_regressor.py
smastelini: Old habits die hard :P
I will fix that!

smastelini: Fixed in [1a91c21](https://github.com/creme-ml/creme/pull/323/commits/1a91c2109076c5b00b947242115ed4f346681c42).

creme/neighbors/base_neighbors.py
smastelini: Changed in [1a91c21](https://github.com/creme-ml/creme/pull/323/commits/1a91c2109076c5b00b947242115ed4f346681c42).

creme/neighbors/base_neighbors.py
smastelini: Nice catch! I changed it in [1a91c21](https://github.com/creme-ml/creme/pull/323/commits/1a91c2109076c5b00b947242115ed4f346681c42). Maybe the current usage is confusing, since we use `append` and pass separate `x` and `y` objects.

smastelini:
> This should be done in another PR as I am in the middle of porting ADWIN and the other drift detection methods. As long as we have the basic classes we can move forward.

Nice! I added the `pop` and `popleft` methods in `KNeighborsBuffer` just to be ready for the future. After the drift detection algorithms are merged, we can easily implement the ADWIN-kNN variants.

> +1 on the idea of giving the user this flexibility.

Same here!

creme/neighbors/base_neighbors.py
smastelini: I added that possibility in the last commit. These hyperparameters are related to the KDTree construction. There is yet another set of useful parameters we could consider. They are used to make queries in the KDTree.

creme/neighbors/base_neighbors.py
```diff
@@ -0,0 +1,208 @@
...
+            x, y = self._X[self._oldest], self._y[self._oldest]
+            self._imask[self._oldest] = False  # Mark slot as free
+            self._oldest = self._oldest + 1 if self._oldest < self.window_size - 1 else 0
+            if self._oldest == self._next_insert:
+                # Shift circular buffer and make its starting point be the index 0
+                self._oldest = self._next_insert = 0
+            self._size -= 1
+
+            return x, y
+        else:
```
MaxHalford, line 138: No need for the `else`.

creme/neighbors/knn_classifier.py
```diff
@@ -0,0 +1,151 @@
...
+    `window_size` training samples. The predicted class-label for a
+    given query sample is obtained in two steps:
+
+    1. Find the closest `n_neighbors` to the query sample in the data window.
+    2. Aggregate the class-labels of the `n_neighbors` to define the predicted
+       class for the query sample.
+
+    Parameters
+    ----------
+    n_neighbors : The number of nearest neighbors to search for.
```
MaxHalford, line 23: The format should be:

```py
n_neighbors
    The number of nearest neighbors to search for.
```

And no need to add an empty line between parameters.

creme/neighbors/knn_classifier.py
```diff
@@ -0,0 +1,151 @@
...
+
+from creme import base
+from creme.utils import dict2numpy
+from creme.utils.math import softmax
+
+from .base_neighbors import BaseNeighbors
+
+
+class KNNClassifier(BaseNeighbors, base.Classifier):
+    """ k-Nearest Neighbors classifier.
```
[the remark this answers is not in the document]
smastelini: Thanks for pointing that out! I changed the documentation in commit [64f91a8](https://github.com/creme-ml/creme/pull/323/commits/64f91a803c1bd8871fdb4e01694f6d1be14d8ad8).

creme/neighbors/knn_classifier.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Notes
+    -----
+    This estimator is not optimal for a mixture of categorical and numerical
+    features. This implementation treats all features from a given stream as
+    numerical.
+
+    Examples
+    --------
+        >>> from creme import datasets
```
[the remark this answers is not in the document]
smastelini: Changed in commit [64f91a8](https://github.com/creme-ml/creme/pull/323/commits/64f91a803c1bd8871fdb4e01694f6d1be14d8ad8).

creme/neighbors/base_neighbors.py
smastelini: Changed in commit [64f91a8](https://github.com/creme-ml/creme/pull/323/commits/64f91a803c1bd8871fdb4e01694f6d1be14d8ad8).

creme/neighbors/knn_classifier.py
smastelini: Changed in commit [64f91a8](https://github.com/creme-ml/creme/pull/323/commits/64f91a803c1bd8871fdb4e01694f6d1be14d8ad8).

PR #327, opened 2020-08-09
Replace fit_one convention by learn_one

gbolmier:
Sorry about commiting that ahead of #325, I forgot to checkout at the right place when I created a `learn_one` branch. Shouldn't matter much though.


MaxHalford:
I just realised that this was targeted to `master` and not `the-merge`, so I reverted the merge. Can you fix the target @gbolmier? :)

gbolmier:
Sure! Sorry for the inconvenience :)

PR #336, opened 2020-08-29
Classification evaluator

jacobmontiel:
This PR is intended to bring back functionality in the form of the "classification evaluator" which tracks incrementally the performance of a classifier and provides multiple performance metrics on demand.

The evaluator is used by some learning methods such as Adaptive Random Forest, Streaming Random Patches, etc.

**Note:** This implementation is intended to recover functionality and will likely change in the future. The intention is to simplify the porting of models and test scripts.

Changes:
- Add evaluator for binary, multi-class, and multi-label classification.
- Add window (rolling) evaluator for binary, multi-class, and multi-label classification.

Subtree changes:
- Update clone method to support `Estimator` type objects
- Update test cases (non-functional changes)
- Disable SAMkNN which was breaking the setup script


MaxHalford:
Hey, sorry I completely forgot about this PR as I took a break from things. Just a question: would it not be worth to keep these evaluators private?

jacobmontiel:
> would it not be worth to keep these evaluators private?

Indeed, I have made all the report classes private for the time being and updated the names for consistency. They will become public once we define the approach to follow.

I also added the regression reports. The single-target classes are based on the existing metrics. The multi-target classes are based on their scikit-multiflow implementation (array-based). This is temporary, all reports will be dictionary-based.

MaxHalford:
Good job! I just have to add a condition in the docs to ignore modules and classes that begin with `_`.

PR #338, opened 2020-09-09
[WIP] Vectorize linear models

gbolmier:
This PR intends to vectorize linear models and optimizers thanks to `utils.VectorDict` implemented by @brcharron (cf. #315, #316, #318). We can merge this PR with master once we are satisfied with the performance.

Related to #280 

To do:
- [x] `facto/fm.py`
- [ ] `facto/ffm.py`
- [ ] `facto/fwfm.py`
- [ ] `facto/hofm.py`
- [x] `linear_model/glm.py`
- [ ] `linear_model/softmax.py`
- [ ] `optim/ada_bound.py`
- [ ] `optim/ada_delta.py`
- [ ] `optim/ada_grad.py`
- [ ] `optim/ada_max.py`
- [ ] `optim/adam.py`
- [ ] `optim/ams_grad.py`
- [ ] `optim/average.py`
- [ ] `optim/ftrl.py`
- [ ] `optim/momentum.py`
- [ ] `optim/nadam.py`
- [x] `optim/nesterov.py`
- [ ] `optim/newton.py`
- [x] `optim/rms_prop.py`
- [x] `optim/sgd.py`

### DIFF
--- a/creme/facto/base.py
+++ b/creme/facto/base.py
@@ -1,5 +1,6 @@
 import abc
 import collections
+import contextlib
 import numbers
 
 from .. import optim
@@ -32,7 +33,7 @@ class BaseFM:
         if weight_initializer is None:
             weight_initializer = optim.initializers.Zeros()
         self.weight_initializer = weight_initializer
-        self.weights = collections.defaultdict(weight_initializer)
+        self._weights = utils.VectorDict(None, weight_initializer)
 
         if latent_initializer is None:
             latent_initializer = optim.initializers.Normal(sigma=.1, seed=seed)
@@ -42,18 +43,35 @@ class BaseFM:
         self.clip_gradient = clip_gradient
         self.seed = seed
 
+    @property
+    def weights(self):
+        return self._weights.to_dict()
+
     @abc.abstractmethod
     def _init_latents(self) -> collections.defaultdict:
         """Initializes latent weights dict."""
 
+    @contextlib.contextmanager
+    def _fit_mode(self, mask=None):
+        weights = self._weights
+        try:
+            # enable the initializer and set a mask
+            self._weights = utils.VectorDict(weights, self.weight_initializer, mask)
+            yield
+        finally:
+            self._weights = weights
+
     def fit_one(self, x, y, sample_weight=1.):
         x = self._ohe_cat_features(x)
 
-        if self.sample_normalization:
-            x_l2_norm = sum((xj ** 2 for xj in x.values())) ** 0.5
-            x = {j: xj / x_l2_norm for j, xj in x.items()}
+        with self._fit_mode(x):
+            x = utils.VectorDict(x)
 
-        return self._fit_one(x, y, sample_weight=sample_weight)
+            if self.sample_normalization:
+                x_l2_norm = (x ** 2).sum() ** 0.5
+                x = x / x_l2_norm
+
+            return self._fit_one(x, y, sample_weight=sample_weight)
 
     def _ohe_cat_features(self, x):
         """One hot encodes string features considering them as categorical."""
@@ -76,7 +94,7 @@ class BaseFM:
 
         # Update the weights
         weights_gradient = self._calculate_weights_gradients(x, g_loss)
-        self.weights = self.weight_optimizer.update_after_pred(w=self.weights, g=weights_gradient)
+        self._weights = self.weight_optimizer.update_after_pred(w=self._weights, g=weights_gradient)
 
         # Update the latent weights
         self._update_latents(x, g_loss)
@@ -89,7 +107,7 @@ class BaseFM:
         y_pred = self.intercept
 
         # Add the unary interactions
-        y_pred += utils.math.dot(x, self.weights)
+        y_pred += self._weights @ x
 
         # Add greater than unary interactions
         y_pred += self._calculate_interactions(x)
--- a/creme/facto/fm.py
+++ b/creme/facto/fm.py
@@ -77,7 +77,7 @@ class FM(BaseFM):
         for j, xj in x.items():
             gradients[j] = {
                 f: g_loss * (xj * precomputed_sum[f] - v[j][f] * xj ** 2) +
-                   l1 * sign(v[j][f]) + l2 * v[j][f]
+                l1 * sign(v[j][f]) + l2 * v[j][f]
                 for f in range(self.n_factors)
             }
 
--- a/creme/facto/fm.py
+++ b/creme/facto/fm.py
@@ -57,14 +57,9 @@ class FM(BaseFM):
         )
 
     def _calculate_weights_gradients(self, x, g_loss):
-
         # For notational convenience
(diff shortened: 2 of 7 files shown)


gbolmier:
Currently, factorization models — including those from `reco` module — store latent factors in numpy arrays and perform latent vectors dot product with `np.dot`.

As latent vectors have fixed sizes and don't need names, I was wondering if `VectorDict` could exploit it to perform faster dot products than numpy. Any opinion @brcharron, @MaxHalford?

MaxHalford:
If the sizes of the weight vectors are known and that they don't have any names, then I see no reason why not to use numpy. I think we should only use VectorDicts for dynamically sized weight vectors.

gbolmier:
Sure Max, I'm just wondering if knowing the size of the weight vectors would allow substantial optimizations.

gbolmier:
Here is a quick benchmark of FM with only 2 variables:

```python
from creme import compose
from creme import datasets
from creme import facto

X_y = [(x, y) for x, y in datasets.MovieLens100K()][:100]

model = compose.Select('user', 'item')
model |= facto.FMRegressor()

%%timeit
for x, y in X_y:
    model.fit_one(x, y)

# On master branch
> 9.44 ms ± 408 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# On vectorize_linear_model branch
> 13.1 ms ± 639 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

gbolmier:
Another one for different operations:

| Experiments   | Python dicts         | VectorDict            | NumPy                |
|:------------------:|:-----------------------:|:-----------------------:|:-----------------------:|
| `x ** 2 / 2 + 1` | 11.9 µs ± 172 ns  | 15.1 µs ± 403 ns  | 2.6 µs ± 107 ns   |
| `<x, x>`           | 8.65 µs ± 69.4 ns | 8.71 µs ± 208 ns  | 953 ns ± 33.6 ns |
| `<xx, x>`         | 8.68 µs ± 81.1 ns | 68.6 µs ± 1.05 µs | 7.24 µs ± 250 ns |

Using the following code:

```python
x = {i: float(i) for i in range(100)}
xx = {i: float(i) for i in range(1_000)}

%timeit {j: xj ** 2 / 2 + 1 for j, xj in x.items()}
%timeit utils.math.dot(x, x)
%timeit utils.math.dot(xx, x)

from creme import utils

x = utils.VectorDict(x)
xx = utils.VectorDict(xx)

%timeit x ** 2 / 2 + 1
%timeit x @ x
%timeit xx @ x

import numpy as np
import random

x = np.array(range(100))
xx = np.array(range(1_000))

mask = random.sample(range(1_000), 100)

%timeit x ** 2 / 2 + 1
%timeit np.dot(x, x)
%timeit np.dot(xx[mask], x)
```

brcharron:
Regarding going faster than `numpy` I would be surprised if a significant improvement was possible.
I think supporting unnamed coefficients in `VectorDict` would make the interface a bit weird.
There could be another abstraction for that though, like `Vector`, if `numpy` does not have all the necessary features but probably using `numpy` under the hood.
Can have a look at scikit-learn's [`WeightVector`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/_weight_vector.pyx#L23)

On the benchmark please have a look at #349 for the dot products.
For `x ** 2 / 2 + 1` the issue is that it's treated as 3 operations in `VectorDict`.
Could have lazy scaling like the above mentioned `WeightVector` and lazy constant shift if needed although it will make things a bit complicated.

gbolmier:
> Regarding going faster than `numpy` I would be surprised if a significant improvement was possible.

If we look at the table above and the benchmark run on your original `VectorDict` implementation (https://github.com/creme-ml/creme/pull/315#issuecomment-656549424), it seems pretty close to numpy.

> I think supporting unnamed coefficients in `VectorDict` would make the interface a bit weird.
> There could be another abstraction for that though, like `Vector`, if `numpy` does not have all the necessary features but probably using `numpy` under the hood.

Agree, if we can't go faster than numpy maybe that's something we can consider later if we need to reduce memory usage of factorization models. To quickly illustrate:

```python
import creme
import numpy as np
import sys

vd = utils.VectorDict({i: 1. for i in range(1_000)})
arr = np.array([1.] * 1_000)

sys.getsizeof(vd)   #   96 bytes
sys.getsizeof(arr)  # 8096 bytes
```

> Can have a look at scikit-learn's [`WeightVector`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/_weight_vector.pyx#L23)

Indeed, would be nice if someone could give a try and report the results.

> For `x ** 2 / 2 + 1` the issue is that it's treated as 3 operations in `VectorDict`.

Makes sense, thx for the explanation. Is there a way to treat the whole statement together?

> Could have lazy scaling like the above mentioned `WeightVector` and lazy constant shift if needed although it will make things a bit complicated.

Yeah, not sure it's worth it too

MaxHalford:
Regarding `sys.getsizeof`, I'm pretty doubtful of those figures. I think `sys.getsizeof` looks at the memory in too shallow way. Maybe that some Python magic method has to overridden to make it correct.

gbolmier:
Good catch Max! I found this interesting [article](https://goshippo.com/blog/measure-real-size-any-python-object/). Basically, `sys.getsizeof` does not recursively add the size of sub-attributes. The article provides code to do so, here is what I get:

```python
get_sizeof(vd)   # 28092 bytes
get_sizeof(arr)  # 40096 bytes
```

smastelini:
Hi @gbolmier. Quite some time ago I implemented an iterative version of the code in the mentioned article to measure the size of objects. This is the utility used till this day to measure memory in skmultiflow. I believe it is already in the merge branch. Perhaps in the future we can adapt the existing function to perform memory measurements in river.

gbolmier:
Hey Saulo! Do you mean something that updates itself every `k` observations? The code shared in the mentioned [article](https://goshippo.com/blog/measure-real-size-any-python-object/) is the one implemented by Max in the `Estimator` base class:

https://github.com/creme-ml/creme/blob/f4563611a641ddf9bb46485fd2ca7f0284715f16/creme/base/estimator.py#L118-L145

smastelini:
> Do you mean something that updates itself every k observations?

Actually, it is an iterative algorithm rather than a recursive one. I changed the recursive calls to a queue of objects to measure.

gbolmier:
> Actually, it is an iterative algorithm rather than a recursive one. I changed the recursive calls to a queue of objects to measure.

I see, nice! Any reason to choose a specific algorithm?

smastelini:
> I see, nice! Any reason to choose a specific algorithm?

It might be overkill to opt for an iterative algorithm. But theoretically, a (very) large object could result in stack overflows due to too many recursive calls. To be fair, I never saw into this situation in my tests. I am just following a conservative/risk-proof approach hahaha.

gbolmier:
Thx for the clear explanation Saulo. Indeed we shouldn't worry too much about that, but as you already implemented it I don't see any reason not to use the iterative algorithm, it's just safer

PR #348, opened 2020-09-15
Add iter_dataset

MaxHalford:
Hey @creme-ml/developers. We agreed that in `river` we would rather a single point of access to datasets, instead of one class per dataset. This is akin to what HuggingFace [is doing](https://github.com/huggingface/datasets), so it's probably a good direction. 

I have added an `iter_dataset` to the `stream` module. This function returns one of the dataset from the `datasets` module. I was thinking that we could simply move the `datasets` module to the `stream` module and make it private.

Because `iter_dataset` returns a dataset, it means you can inspect it, as so:

```py
>>> from creme import stream

>>> dataset = stream.iter_dataset('Bikes')
>>> dataset
Bikes dataset

              Task  Regression                                                    
 Number of samples  182,470                                                       
Number of features  8                                                             
            Sparse  False                                                         
              Path  /Users/mhalford/creme_data/Bikes/toulouse_bikes.csv           
               URL  https://maxhalford.github.io/files/datasets/toulouse_bikes.zip
              Size  12.52 MB                                                      
        Downloaded  True  
```

You can iterate over it as so:

```py
>>> for x, y in dataset:
...     break
```

You can see the available datasets like this:

```py
>>> stream.available_datasets()
['AirlinePassengers',
 'Bananas',
 'Bikes',
 'ChickWeights',
 'CreditCard',
 'Elec2',
 'HTTP',
 'Higgs',
 'ImageSegments',
 'Insects',
 'MaliciousURL',
 'MovieLens100K',
 'Music',
 'Phishing',
 'Restaurants',
 'SMSSpam',
 'SMTP',
 'SolarFlare',
 'TREC07',
 'Taxis',
 'TrumpApproval']
```

One special case is the `Insects` dataset, which has different variants. Basically it has a `variant` parameter. You can pass it as an extra keywork argument:

```py
>>> dataset = stream.iter_dataset('Insects', variant='abrupt_balanced')
```

Tell me what you guys think :). I'm happy to explore other options.

@jacobmontiel @smastelini: what do you guys think about the name `stream` instead of `data`? We agreed on the latter but now this seems a bit too generic to me. Happy to discuss.


raphaelsty:
Sounds good to me :-)

smastelini:
Hi @MaxHalford, it's looking pretty good!

The idea of separating the submodules `stream` and `data` was to make their goals clear. But to be fair, if we are OK with having a lot of tools  in `river.stream`, from the user standpoint the usage would be simplified.

Personally, I like this idea: everything related to data (streams) in the same place. :D

MaxHalford:
> Personally, I like this idea: everything related to data (streams) in the same place. :D 

As far as I recall, we actually agreed on this with Jacob! But we agreed on calling the module `data` instead of `stream`. I'm just saying that I like `stream` better 👍

gbolmier:
Good idea Max. I think we should add something to the documentation so that users can still access information about every datasets there and not just the name of the available ones

> what do you guys think about the name `stream` instead of `data`? We agreed on the latter but now this seems a bit too generic to me

Agree on that too

MaxHalford:
> I think we should add something to the documentation so that users can still access information about every datasets there and not just the name of the available ones

Yeah I was thinking of adding this to the "Reading data" user guide. I still have to think about this a little.

MaxHalford:
@gbolmier I added an [entry to the user guide](https://github.com/creme-ml/creme/blob/dataset-class/docs/user-guide/available-datasets.ipynb). Tell me what you think about it! I like it because it's hackable and it gives the user some insights on how to inspect things by herself.

gbolmier:
Nice I really like it! `stream.iter_dataset` documentation page should point to the user guide entry. I think we should also print a little description of every datasets, if I'm not wrong we have that within docstrings. The idea is to enable users to easily access datasets information, and quickly get a glimpse of what a dataset is about.

PR #349, opened 2020-09-15
Performance improvements for VectorDict

brcharron:
Implements a fast track for dot products when only the intersection of the keys is needed.
Also a bug fix for cases with mask.
This addresses concerns of [low performance](https://github.com/creme-ml/creme/pull/316#issuecomment-684126028) shared by @gbolmier, below is updated benchmarks in this branch.

```py
from creme import utils

xx = {i: i for i in range(1_000)}
y = utils.VectorDict(x)
yy = utils.VectorDict(xx)

%timeit utils.math.dot(x, x)
# 12.2 µs ± 120 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit utils.math.dot(xx, x)
# 12.1 µs ± 95.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit y @ y
# 8.93 µs ± 126 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit yy @ y
# 8.93 µs ± 102 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

### DIFF
--- a/creme/utils/test_vectordict.py
+++ b/creme/utils/test_vectordict.py
@@ -89,18 +89,18 @@ def test_vectordict():
     assert x['f'] == 2
 
     # test mask
-    x = {'a': 1, 'b': -5}
+    x = {'a': 1, 'b': -5, 'e': 2}
     y = {'b': 0.5, 'd': 4, 'e': 3, 'f': 8}
     z = {'b': 4, 'd': 2, 'g': -1}
     vx = VectorDict(x)
     vy = VectorDict(y)
-    assert vx + vy == {'a': 1, 'b': -4.5, 'd': 4, 'e': 3, 'f': 8}
+    assert vx + vy == vy + vx == {'a': 1, 'b': -4.5, 'd': 4, 'e': 5, 'f': 8}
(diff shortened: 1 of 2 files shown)


gbolmier:
Thx for your work! Do you think there is still room for improvement? Your original implementation reported ≈ 10x improvement for 100 features (https://github.com/creme-ml/creme/pull/315#issuecomment-656549424). Or does `VectorDict` genericity prevent better performance?

MaxHalford:
I'll merge this because it's an improvement. You can keep discussing here or elsewhere :)

MaxHalford:
@brcharron I believe some of the unit tests in `test_vectordict.py` are not working, can you check?

brcharron:
@MaxHalford Thanks for merging, the tests pass on my side, did you rebuild Cython maybe?
Which assert broke? I added some to catch a bug that was in `master`.

MaxHalford:
Lol it's all good, I'd forgotten to rebuild. Got to get better at that 👍

PR #353, opened 2020-09-18
[WIP] Implement lazy scaling and shifting for VectorDict

brcharron:
Draft PR related to [this](https://github.com/creme-ml/creme/pull/338#issuecomment-689853117) and [that](https://github.com/creme-ml/creme/pull/316#issuecomment-692907210) comment, implements lazy scaling and shifting for `VectorDict`.
First 2 commits are preparation, just look at the 3rd one for the implementation.

Benchmarks below use `.keys()` to force the evaluation of the lazy part so it's comparable to the non-lazy implementation (copy at each operation).

```py
from creme import utils
y = utils.VectorDict(x)

# this PR
%timeit y.keys()
106 ns ± 0.0774 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
%timeit y + 1
139 ns ± 0.337 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
%timeit (y + 1).keys()
6 µs ± 32.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit (y ** 2 / 2 + 1).keys()
31.1 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
%timeit (2 * y + 3 * y).keys()
14.2 µs ± 42.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

# previous PR #351, should be similar to master
%timeit y.keys()
108 ns ± 0.132 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
%timeit y + 1
5.08 µs ± 20.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit (y + 1).keys()
5.13 µs ± 12.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit (y ** 2 / 2 + 1).keys()
35.6 µs ± 106 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
%timeit (2 * y + 3 * y).keys()
20.5 µs ± 86.5 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

We can see it's a bit faster for complex operations but it's not huge and it makes other operations a bit slower.
Also it hurts maintainability.

Maybe some custom methods as mentioned [here](https://github.com/creme-ml/creme/pull/316#issuecomment-658676077) would be a better option although a wrapper would be needed to get the same expressions with `numpy` arrays.

### DIFF
--- a/creme/utils/vectordict.pyx
+++ b/creme/utils/vectordict.pyx
@@ -57,6 +57,9 @@ cdef class VectorDict:
     cdef bint _lazy_mask
     cdef bint _use_factory
     cdef object _default_factory
+    cdef object _scale
+    cdef object _shift
+    cdef bint _evaluated
 
     def __init__(self, data=None, default_factory=None, mask=None, copy=False):
         """A dictionary-like object that supports vector-like operations.
@@ -109,6 +112,9 @@ cdef class VectorDict:
                 if data_._lazy_mask and mask is not data_._mask:
                     raise ValueError(
                         "Cannot mask a masked VectorDict without copy")
+                if not data_._evaluated:
+                    raise ValueError(
+                        "Cannot wrap an unevaluated VectorDict without copy")
                 data = data_._data
         elif not isinstance(data, dict):
             raise ValueError(f"Unsupported type for data: {type(data)}")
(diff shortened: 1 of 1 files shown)


MaxHalford:
Thanks Bruno. 

As you hinted at, I don't think that there's more cons than pros. It is an interesting topic nonetheless, and certainly one where I should spend more time looking at. In a way, numpy sort of has the same problem, right? That's why [numexpr](https://github.com/pydata/numexpr) was created. Maybe that if we grow bigger, there might some justification to spend more time here and dedicate effort towards a tidy module that is dedicated to `VectorDict`. I don't know.

But anyway, things are very good right now, no need to throw the kitchen sink at the problem, at least that's how things seem to me.

brcharron:
Indeed `numpy` has the same problem although since it's pretty fast already it's a bit less of an issue. I will close this one then and maybe this can be revisited in the future with an option that can deal with both arrays and dictionaries, potentially on the lines of `numexpr`.

PR #356, opened 2020-09-23
Data stream generators

jacobmontiel:
This PR is to port the existing synthetic stream generators:

- [x] `Agrawal`
- [x] `AnomalySine`
- [x] `ConceptDriftStream`
- [x] `Friedman`
- [x] `Hyperplane`
- [x] `LED`
- [x] `LEDDrift`
- [x] `Logical`
- [x] `Mixed`
- [x] `RandomRBF`
- [x] `RandomRBFDrift`
- [x] `RandomTree`
- [x] `SEA`
- [x] `Sine`
- [x] `STAGGER`
- [x] `Waveform`

Not ported generators are:
- `DataStream` and `FileStream`. Functionality is already available via `iter_csv` and `iter_array`.
- `MultilabelGenerator` and `RegressionGenerator`. They rely on `scikit-learn` data generators. The same functionality can be achieved using the `scikit-learn` generator and `iter_array`.

Notes: Some methods have (minor) changes that result in (expected) changes in usage/data values.
- `LEDDrift`. Setting the seed value was (unnecessarily) set twice. The new implementation fixes this.
- `RandomTree`. The new implementation returns the nominal features in single columns. Previously, these features were one-hot encoded.
- `SEA`. This is a different implementation that uses Python's random number generator.


creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
+import numpy as np
+
+from . import base
+from creme.utils.skmultiflow_utils import check_random_state
+
+
+class AGRAWALGenerator(base.SyntheticDataset):
```
MaxHalford, line 7: How about calling this simply `Agrawal`? `AGRAWAL` isn't an acronym so I don't think that it should be capitalized. I also don't think that `Generator` brings anything meaningful.

If we want to make a distinction between finite datasets and infinite datasets (aka. generators) then I could add an `iter_generator` function in addition to `iter_dataset`. What do you think? :)

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+    """Agrawal stream generator.
+
+    The generator was introduced by Agrawal et al., and was common source
+    of data for early work on scaling up decision tree learners.
+    The generator produces a stream containing nine features, six numeric and
+    three categorical.
+    There are ten functions defined for generating binary class labels from the
+    features. Presumably these determine whether the loan should be approved.
+    The features and functions are listed in the original paper [1]_.
```
MaxHalford, line 16: The reference has to be `[^1]` instead of `[1]_`, as explained [here](https://squidfunk.github.io/mkdocs-material/reference/footnotes/).

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
+import numpy as np
+
+from . import base
+from creme.utils.skmultiflow_utils import check_random_state
+
+
+class AGRAWALGenerator(base.SyntheticDataset):
+    """Agrawal stream generator.
+
+    The generator was introduced by Agrawal et al., and was common source
```
MaxHalford, line 10: was **a** common

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+    | elevel       | the education level  | uniformly chosen from 0 to 4                                        |
+    | car          | car maker            | uniformly chosen from 1 to 20                                       |
+    | zipcode      | zip code of the town | uniformly chosen from 0 to 8                                        |
+    | hvalue       | value of the house   | uniformly distributed from 50k x zipcode to 100k x zipcode          |
+    | hyears       | years house owned    | uniformly distributed from 1 to 30                                  |
+    | loan         | total loan amount    | uniformly distributed from 0 to 500k                                |
+
+    Parameters
+    ----------
+    classification_function: int (Default=0)
```
MaxHalford, line 32: You don't have to specify types and default values. They are extracted from the `__init__` signature.

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+        Which of the four classification functions to use for the generation.
+        The value can vary from 0 to 9.
+
```
MaxHalford, line 35: I've been using as a convention not to include spaces between parameters. I don't think it matters that much, just saying. I think that without the types and default values there's less clutter, so I don't find there's a need to add an extra space. To each their own :)

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+    random_state: int, RandomState instance or None, optional (default=None)
```
MaxHalford, line 36: I've been using `seed` as a convention instead of `random_state`. I get that `random_state` is what `sklearn` does, I just find that in most cases users specify an integer and not a `numpy.random.RandomState`.

Also, I've been using the `random` module as a convention instead of `numpy.random`. The reason why is that `random` is usually faster for generating a single value, which happens a lot in a streaming context. Here's a quick benchmark on my laptop:

```py
import random
import numpy as np

seed = 42

rng = random.Random(seed)
np_rng = np.random.RandomState(seed)

%timeit rng.random()
%timeit np_rng.random()
```

```
66.5 ns ± 0.383 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
298 ns ± 2.72 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Therefore, I really encourage us to use `random` instead of `numpy.random` when possible. For instance, this is what I've done in [`SEA`](https://github.com/creme-ml/creme/blob/the-merge/creme/stream/datasets/sea.py).

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+        self.current_sample_y = None
+        self.sample_idx = 0
+
+        self._prepare_for_use()
+
+    def _prepare_for_use(self):
+        self._random_state = check_random_state(self.random_state)
+        self._next_class_should_be_zero = False
+
+    def restart(self):
```
MaxHalford, line 122: I'm not sure about having a `restart` function. If we do have it, then do we need to implement for each generator?

I believe that an alternative approach would be to have to be `__iter__` be idempotent. That is, if you set a seed, then looping over the dataset a second time should produce the same data as when iterating over it a first time. In the case of this generator, this simply requires to set everything up at the start of `__iter__`.

What do you think? If we agree on this convention I'll have to modify the two existing generators, which are `SEA` and `Friedman`. I can even write a unit test to check that `__iter__` is idempotent.

creme/stream/datasets/agrawal.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        new_function = self._random_state.randint(10)
+        while new_function == self.classification_function:
+            new_function = self._random_state.randint(10)
+        self.classification_function = new_function
+
+    @staticmethod
+    def _classification_function_0(salary, commission, age, elevel, car, zipcode, hvalue,
+                                   hyears, loan):
+        return 0 if ((age < 40) or (60 <= age)) else 1
```
MaxHalford, line 214: This could be rewritten as `int(age < 40 or 60 <= age)`. The same goes for the rest of the code. It's not very important though. I don't mind cleaning things up once this is merged. I actually enjoy doing that kind of work :)

creme/stream/datasets/agrawal.py
jacobmontiel: > How about calling this simply Agrawal? AGRAWAL isn't an acronym so I don't think that it should be capitalized. I also don't think that Generator brings anything meaningful.

Agree and done.

> If we want to make a distinction between finite datasets and infinite datasets (aka. generators) then I could add an iter_generator function in addition to iter_dataset. What do you think? :)

+1 for `iter_generator`

creme/stream/datasets/agrawal.py
jacobmontiel: Fixed.

creme/stream/datasets/agrawal.py
jacobmontiel: Fixed.

creme/stream/datasets/agrawal.py
jacobmontiel: Old habits die hard :-) 
I removed types and default values.

creme/stream/datasets/agrawal.py
jacobmontiel: I agree, it looks better. Edited.

creme/stream/datasets/agrawal.py
jacobmontiel: Agree, in this case using `random` works well. Replaced.

creme/stream/datasets/agrawal.py
jacobmontiel: Agree. This is a good time to revisit the code and consider cosmetic changes that can improve the code base.

creme/stream/datasets/agrawal.py
jacobmontiel: This approach has been agreed upon.

Setup is performed in the `__iter__` method so a new  `for` loop will generate the same data (is the random number generator is seeded).

For the corner case where we need to continue from where we stop the loop we can create a generator as follows:

```python
stream = iter(synt_dataset)

# Generate and print 3 samples
for i, (x,y) in enumerate(stream):
    print(x,y)
    if i == 3:
        break

# Calling again the "stream" generator will continue from it left
for i, (x,y) in enumerate(stream):
    print(x,y)
    if i == 3:
        break
```

creme/datasets/synth/concept_drift_stream.py
```diff
@@ -0,0 +1,128 @@
...
+    {0: 8.9217, 1: 0.8693, 2: 4.2192} True
+    {0: 2.2321, 1: 7.3647, 2: 6.7669} True
+    {0: 8.9217, 1: 0.8693, 2: 4.2192} True
+    {0: 0.2979, 1: 2.1863, 2: 5.0535} False
+    {0: 0.2653, 1: 1.9883, 2: 6.4988} False
+    {0: 5.4494, 1: 2.2044, 2: 5.8926} False
+    {0: 8.0943, 1: 0.0649, 2: 8.0581} False
+    """
+
+    def __init__(self, stream: base.SyntheticDataset = Agrawal(seed=112),
```
MaxHalford, line 79: It's not good practice to set default values with mutable values. See [this](http://effbot.org/zone/default-values.htm) for an explanation why. The correct way to proceed is either to not specify a default value, which in my opinion makes sense in this case, or either to specify `None` and check `if is None` before setting the `Agrawal` default. Note that if you use the latter option you can reuse the same `seed` parameter you're passing to `ConceptDriftStream`.

@smastelini this applies to you too, I've seen you doing it with the trees during the last Zoom call :)

creme/datasets/synth/led.py
```diff
@@ -0,0 +1,192 @@
...
+    irrelevant_features
+        Adds 17 non-relevant attributes to the stream.
+
+    Notes
+    -----
+    An instance is generated based on the parameters passed. If `has_noise`
+    is set then the total number of attributes will be 24, otherwise there will
+    be 7 attributes.
+
+    References
```
MaxHalford, line 38: Normally the references go after the example. This won't affect the docs in any case.

creme/datasets/synth/led.py
```diff
@@ -0,0 +1,192 @@
...
+
+    [^2]: A. Asuncion and D. J. Newman. UCI Machine Learning Repository
+          [http://www.ics.uci.edu/∼mlearn/mlrepository.html].
+          University of California, Irvine, School of Information and
+          Computer Sciences,2007.
+
+    Examples
+    --------
+    >>> from creme import synth
+    >>>
```
MaxHalford, line 53: No need for empty `>>>`s.

MaxHalford (commented):
Ok I'm done reviewing! I haven't checked the logic because there's too much of it. As long as it works it's fine. There's probably opportunities to streamline the code but we can come back to that later. Good job!

creme/stream/datasets/agrawal.py
jacobmontiel: I rolled back the change of the random number generator in order to use NumPy's. This is to ensure replicability with the previous implementation when using the same seed values. The variable/attribute name was kept as `seed`.

creme/datasets/synth/led.py
jacobmontiel: I was following the Numpy docstring format which places the example at the end.

https://numpydoc.readthedocs.io/en/latest/format.html

I don't mind changing the location. It might be worth documenting the docstring sections order for consistency.

creme/datasets/synth/anomaly_sine.py
```diff
@@ -0,0 +1,126 @@
...
+    from attribute 2 with values from a sine function different to the one
+    used in attribute 1. The ``contextual`` flag can be used to introduce
+    contextual anomalies which are values in the normal global range,
+    but abnormal compared to the seasonal pattern. Contextual attributes
+    are introduced by replacing values in attribute 2 with values from
+    attribute 1.
+
+    """
+
+    def __init__(self, n_samples: int = 10000, n_anomalies: int = 2500, contextual: bool = False,
```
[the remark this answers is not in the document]
jacobmontiel: I see your point. Personally, I find it easier the consistency of type hinting all attributes. `bool` seems like an exception.

creme/datasets/synth/__init__.py
```diff
@@ -3,11 +3,39 @@
...
     'Friedman',
-    'SEA'
+    'Hyperplane',
+    'LED',
+    'LEDDrift',
+    'Logical',
+    "Mixed",
+    "RandomRBF",
+    "RandomRBFDrift",
+    "RandomTree",
```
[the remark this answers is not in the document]
jacobmontiel: Good catch. Fixed.

creme/datasets/synth/agrawal.py
```diff
@@ -0,0 +1,286 @@
...
+        If int, `seed` is used to seed the random number generator;
+        If RandomState instance, `seed` is the random number generator;
+        If None, the random number generator is the `RandomState` instance used
+        by `np.random`.
+    balance_classes
+        Whether to balance classes or not. If balanced, the class
+        distribution will converge to a uniform distribution.
+    perturbation
+        The probability that noise will happen in the generation. At each
+        new sample generated, the sample with will perturbed by the amount of
```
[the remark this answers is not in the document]
jacobmontiel: Fixed.

creme/datasets/synth/agrawal.py
```diff
@@ -0,0 +1,286 @@
...
+    | car          | car maker            | uniformly chosen from 1 to 20                                       |
+    | zipcode      | zip code of the town | uniformly chosen from 0 to 8                                        |
+    | hvalue       | value of the house   | uniformly distributed from 50k x zipcode to 100k x zipcode          |
+    | hyears       | years house owned    | uniformly distributed from 1 to 30                                  |
+    | loan         | total loan amount    | uniformly distributed from 0 to 500k                                |
+
+    Parameters
+    ----------
+    classification_function
+        Which of the four classification functions to use for the generation.
```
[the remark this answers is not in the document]
jacobmontiel: Fixed.

creme/datasets/synth/led.py
jacobmontiel: I have reordered the sections, now `Notes` and `References` are placed after the examples. This is since arguably the examples are more relevant to the user.

creme/datasets/synth/led.py
jacobmontiel: Removed.

creme/datasets/synth/concept_drift_stream.py
jacobmontiel: Agree on the best practice for mutable attributes. However, I am missing its application in this case. Is the `Agrawal` generator mutable when the `seed` is set?

MaxHalford: Probably not! Just wanted to make you aware. Don't worry about it in this case :)

smastelini: > @smastelini this applies to you too, I've seen you doing it with the trees during the last Zoom call :)

Hi @MaxHalford, I do instantiate a regressor as the default attribute. In practice, I clone this model to replicate it to the leaves. Should I change it to `None`? No modifications are performed in the passed object. That's why I didn't bother much about it at the time.

jacobmontiel:
I think the PR is ready. Let me know if there is something else to do @MaxHalford :-)

creme/datasets/synth/concept_drift_stream.py
MaxHalford: @smastelini it's a code smell to not use `None`. Users always find a way to get bitten by this :)

PR #357, opened 2020-09-23
Make streaming datasets idempotent

MaxHalford:
Here's an example with the `SEA` dataset. @smastelini as you can see the parameters are shown in `__repr__`. Do you think that this is good enough?

```python
>>> from creme import stream

>>> gen = stream.iter_dataset('SEA', seed=42)
>>> gen
SEA synthetic dataset.

Each observation is composed of 3 features. Only the first two features are relevant. The
target is binary, and is positive if the sum of the features exceeds a certain threshold. There
are 4 thresholds to choose from. Concept drift can be introduced by switching the threshold
anytime during the stream.

    Name  SEA                  
    Task  Binary classification
 Samples  ∞                    
Features  3                    
  Sparse  False                
 Variant  0                    

Parameters:

    variant
        Determines the classification function to use. Possible choices are 0, 1, 2, 3.
    noise
        Determines the amount of observations for which the target sign will be flipped.
    seed
        Random seed number used for reproducibility.
```

With this change, the generator is idempotent, meaning that iterating over it twice gives the same results.

```py
>>> for i, (x, y) in enumerate(gen.take(5)):
...     print(i, x, y)
0 {0: 6.394267984578837, 1: 0.25010755222666936, 2: 2.7502931836911926} False
1 {0: 2.2321073814882277, 1: 7.364712141640124, 2: 6.766994874229113} True
2 {0: 8.921795677048454, 1: 0.8693883262941615, 2: 4.2192181968527045} True
3 {0: 0.29797219438070344, 1: 2.1863797480360336, 2: 5.053552881033624} False
4 {0: 0.26535969683863625, 1: 1.988376506866485, 2: 6.498844377795232} False

>>> for i, (x, y) in enumerate(gen.take(5)):
...     print(i, x, y)
0 {0: 6.394267984578837, 1: 0.25010755222666936, 2: 2.7502931836911926} False
1 {0: 2.2321073814882277, 1: 7.364712141640124, 2: 6.766994874229113} True
2 {0: 8.921795677048454, 1: 0.8693883262941615, 2: 4.2192181968527045} True
3 {0: 0.29797219438070344, 1: 2.1863797480360336, 2: 5.053552881033624} False
4 {0: 0.26535969683863625, 1: 1.988376506866485, 2: 6.498844377795232} False
```

We can add a "pausing" mechanism by using `iter`, as so:

```py
>>> gen = iter(stream.iter_dataset('SEA', seed=42))
>>> type(gen)
generator
```

Using `iter` means that `gen` is a generator and not a `SyntheticDataset`, which means that we can't use `.take(5)`. We can however use `islice`:

```py
>>> import itertools

>>> for i, (x, y) in enumerate(itertools.islice(gen, 2)):
...     print(i, x, y)
0 {0: 6.394267984578837, 1: 0.25010755222666936, 2: 2.7502931836911926} False
1 {0: 2.2321073814882277, 1: 7.364712141640124, 2: 6.766994874229113} True
```

We can pick up from where we started:

```py
>>> for i, (x, y) in enumerate(itertools.islice(gen, 3), start=i + 1):
...     print(i, x, y)
2 {0: 8.921795677048454, 1: 0.8693883262941615, 2: 4.2192181968527045} True
3 {0: 0.29797219438070344, 1: 2.1863797480360336, 2: 5.053552881033624} False
4 {0: 0.26535969683863625, 1: 1.988376506866485, 2: 6.498844377795232} False
```


smastelini:
> Here's an example with the SEA dataset. @smastelini as you can see the parameters are shown in __repr__. Do you think that this is good enough?

My main concern is how the user is going to customize the synthetic datasets. Will they have to rely on `set_params`?

That issue aside, the `__repr__` information is more than enough :)

MaxHalford:
The users would just have to pass parameters to `iter_dataset`, as so:

```py
gen = stream.iter_dataset('SEA', variant=2, noise=3.1, seed=42)
```

smastelini:
Will the datasets be accessible standalone? I might be confusing things here, but my concern is that the user will have to firstly instantiate a dataset to be able to access its description. Was I clear in my explanation? Do not hesitate to ask me for clarification.

smastelini:
XD no worries!

As I understand it, the only access point to synthetic dataset will be by using `iter_dataset`, right?

So, the user can customize the generators with `**kwargs`. So far, so good :D

My point is, where does one find information about the synthetic datasets? Well, they can find it in the `__repr__` method of the dataset, since we are not possibly going to put all the descriptions in the documentation of `iter_dataset`.

But if that is true, then one has to get the dataset (using the default parameters) and customize it using `set_params` or perform another call to `iter_dataset` and use the desired parameters (after reading about them in the documentation).

Am I confusing things too much? Hahahaha

MaxHalford:
Yeah I know what you mean. That's the problem of having a single point of entry to the datasets, it makes the documentation less clear. I've added a [notebook](https://github.com/creme-ml/creme/blob/the-merge/docs/user-guide/available-datasets.ipynb) that enumerates all the datasets. It's nice because it gives a bird's eye view of the datasets. Do you think that's enough?

smastelini:
What about a method that does something like that? Given the dataset key, returns its description. Or if we come up with a way to put the datasets' docstrings in the API reference page, that would be nice too.

MaxHalford:
Yeah I can definitely generate a page and put it in the API reference.

PR #365, opened 2020-10-18
Add Robust Soft Learning Vector Quantization method

jacobmontiel:


MaxHalford:
I don't know this method so it's hard for me to comment on the implementation.

Obviously, in an ideal world, this would/should be using dicts instead of arrays. Additionally, there's things like the optimizer and the learning rate scheduler that could make use of what already exists in the `optim` module. This definitely needs to be heavily refactored. However, if you feel that this is an essential addition for the release, I won't veto the merge.

creme/proto/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Prototype- based methods."""
```
MaxHalford, line 1: nit: no need for whitespace after the dash.

More important: could you possible add a short paragraph to give some intuition on what are prototype-based methods? I don't think that a lot of people are aware of them.

creme/proto/robust_soft_learning_vector_quantization.py
```diff
@@ -0,0 +1,388 @@
...
+from sklearn.utils import validation
+from sklearn.metrics import euclidean_distances
+from sklearn.utils.multiclass import unique_labels
+
+from creme.base import Classifier
+from creme.utils.skmultiflow_utils import check_random_state
+from creme.utils import dict2numpy
+
+
+class RobustSoftLearningVectorQuantization(Classifier):
```
MaxHalford, line 14: TBH I would call this `RSLVQClassifier`, or `RobustSoftLVQClassifier` considering that `LVQ` is an acronym that is used in research papers. `RobustSoftLearningVectorQuantization` is too big a mouthful IMO :)

I wouldn't worry about using acronyms. Indeed, in the documentation we always show the first line of the docstring to explain the acronym anyway. There's no need to stutter.

creme/proto/robust_soft_learning_vector_quantization.py
```diff
@@ -0,0 +1,388 @@
...
+    """Robust Soft Learning Vector Quantization for Streaming and Non-Streaming Data.
+
+    By choosing another gradient descent method the Robust Soft Learning Vector Quantization
+    (RSLVQ) method can be used as an adaptive version.
+
+    Parameters
+    ----------
+    prototypes_per_class
+        Number of prototypes per class.
+    initial_prototypes: array-like, shape =  , optional
```
MaxHalford, line 24: No need for the type hint.

scikit-multiflow/src/skmultiflow/prototype/robust_soft_learning_vector_quantization.py
```diff
@@ -1,407 +0,0 @@
-import math
```
MaxHalford, line 1: You guys don't have to remove the files from the `skmultiflow` directory, we'll remove it entirely once we merge anyway :)

creme/proto/__init__.py
jacobmontiel: Fixed the typo and added a short description.

creme/proto/robust_soft_learning_vector_quantization.py
jacobmontiel: I renamed the class as `RSLVQClassifier`. It is a good compromise. 

We went from `RSLVQ` -> `RobustSoftLearningVectorQuantization` in an attempt to achieve clarity, but we overdid it :-)

creme/proto/robust_soft_learning_vector_quantization.py
jacobmontiel: My bad, this was not supposed to be there. I went back and fixed other minor issues with the documentation.

scikit-multiflow/src/skmultiflow/prototype/robust_soft_learning_vector_quantization.py
jacobmontiel: This is to keep the history of the file (`git mv`). Also helps to clearly highlight what we are modifying. Makes things easier to track down issues from the required changes and (minor) maintenance that we are performing.

jacobmontiel:
> Obviously, in an ideal world, this would/should be using dicts instead of arrays. Additionally, there's things like the optimizer and the learning rate scheduler that could make use of what already exists in the `optim` module. This definitely needs to be heavily refactored.

I completely agree. There are multiple "areas of opportunity" in this implementation. I am not a big fan of using extensively `sklearn`, especially if there are in-house alternatives. However, refactoring will take some effort and at this point, I rather focus on getting the functionality in place. We could even contact the authors of this method to see if they are willing to help. I am thinking that if something ends in the release, even when refactoring is needed, it means that we should prioritize it. Methods that will be out of the release are less critical and can be considered later as resources are available. Of course, I am open to discussing this approach ;-)

> However, if you feel that this is an essential addition for the release, I won't veto the merge.

Prototype-based methods are nice for its white-box philosophy. I hope we will see more methods like this in the future.

MaxHalford:
Thanks for all the answers!

PR #366, opened 2020-10-20
ADWINBagging and LeveragingBagging

jacobmontiel:
This PR includes two bagging methods:

- `ADWINBagging`
- `LeveragingBagging`


MaxHalford:
LGTM! The methods are a bit heavy but that's fine for now. Maybe in the future we can include some variant where we don't check for drift and whatnot every `k` iterations instead of every iteration.

PR #376, opened 2020-11-01
Trees: tweaks or treats

smastelini:
This PR adds the following capabilities to the trees:

- debug_one
- draw

It also modifies the current implementation by:
- making all the classification trees' nodes returning the class probabilities directly in their `predict_one`, rather than counts.
- rename `predict_one` to `leaf_prediction` in the trees' leaves to avoid confusion between classifier nodes and regression nodes
.
Besides that, the trees use the standard `class_counts / sum_class_counts` normalization for the probabilities instead of softmax.


river/tree/_base_tree.py
```diff
@@ -311,3 +322,244 @@ def __find_learning_nodes(self, node, parent, parent_branch, found):
...
+    https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/_export.py
+    """
+    colors = []
+
+    # Initialize saturation & value; calculate chroma & value shift
+    s, v = .75, .9
+    c = s * v
+    m = v - c
+
+    for h in [i for i in range(25, 385, int(360 / n))]:
```
jacobmontiel, line 536: The `Hoeffding Adaptive Tree classifier` triggers an error:

```
~/Documents/workspace/git/river/river/tree/_base_tree.py in _color_brew(n)
    534     m = v - c
    535 
--> 536     for h in [i for i in range(25, 385, int(360 / n))]:
    537 
    538         # Calculate some intermediate values

ZeroDivisionError: division by zero
```

MWE:
```python
from river import synth
from river import evaluate
from river import metrics
from river import tree
gen = synth.Agrawal(classification_function=0, seed=42)
# Take 1000 instances from the infinite data generator
dataset = iter(gen.take(1000))
model = tree.HoeffdingAdaptiveTreeClassifier(
    grace_period=100,
    split_confidence=1e-5,
    nominal_attributes=['elevel', 'car', 'zipcode']
)
metric = metrics.Accuracy()
evaluate.progressive_val_score(dataset, model, metric)

model.draw()
```

river/tree/_nodes/base.py
```diff
@@ -50,6 +50,24 @@ def filter_instance_to_leaf(self, x: dict, parent: 'Node', parent_branch: int) -
         """
         return FoundNode(self, parent, parent_branch)
 
+    def path(self, x) -> Iterator['Node']:
```
jacobmontiel, line 53: Interesting. Did you notice any difference in the training/testing time due to this new way to traverse the tree?

river/tree/_nodes/hatc_nodes.py
```diff
@@ -287,8 +293,13 @@ def learn_one(self, x, y, *, sample_weight=1., tree=None, parent=None, parent_br
 
     def predict_one(self, x, *, tree=None):
         # In case split nodes end up being used (if emerging categorical feature appears,
-        # for instance)
-        return self.stats  # Use the MC (majority class) prediction strategy
+        # for instance) use the MC (majority class) prediction strategy
+        dist = {c: 0. for c in self.stats}
+        votes_sum = sum(self.stats.values())
+        if votes_sum > 0:
```
jacobmontiel, line 299: We can take the chance to improve this method. It is not optimal to protect it externally and resulted in multiple issues opened in the past due to the error throwing

```python
def normalize_values_in_dict(dictionary, factor=None, inplace=True, raise_error=False):
    """Normalize the values in a dictionary using the given factor.

    For each element in the dictionary, applies `value/factor`.

    Parameters
    ----------
    dictionary
        Dictionary to normalize.
    factor
        Normalization factor value. If not set, use the sum of values.
    inplace
        If True, perform operation in-place
    raise_error
        If `True`, will raise an error if the normalization factor is `0` or `None`. If `False`,
        will return gracefully if the normalization factor is `0` or `None`.

    """
    if factor is None:
        factor = sum(dictionary.values())
    if raise_error and (factor == 0 or math.isnan(factor)):
        raise ValueError(f'Can not normalize, normalization factor is {factor}')

    if not inplace:
        dictionary = copy.deepcopy(dictionary)

    if math.isnan(factor) or factor == 0:
        # Can not normalize, return gracefully
        return dictionary

    for key, value in dictionary.items():  # loop over the keys, values in the dictionary
        dictionary[key] = value / factor

    return dictionary
```

river/utils/skmultiflow_utils.py
```diff
@@ -264,3 +264,26 @@ def check_random_state(seed):
     if isinstance(seed, np.random.RandomState):
         return seed
     raise ValueError(f'{seed} cannot be used to seed a numpy.random.RandomState instance')
+
+
+def round_sig_fig(x, significant_digits=2):
```
jacobmontiel, line 269: What is the purpose of this?

river/ensemble/adaptive_random_forest.py
```diff
@@ -399,7 +399,7 @@ class AdaptiveRandomForestClassifier(BaseForest, base.Classifier):
     >>> metric = metrics.Accuracy()
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
-    Accuracy: 68.17%
+    Accuracy: 72.87%
```
jacobmontiel, line 402: I assume this comes from the change in the normalization.

river/ensemble/boosting.py
```diff
@@ -63,7 +63,7 @@ class AdaBoostClassifier(base.WrapperMixin, base.EnsembleMixin, base.Classifier)
     ... )
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
-    LogLoss: 0.559289
+    LogLoss: 0.37111
```
jacobmontiel, line 66: Very interesting to see the impact of the normalization on boosting. Nice.

jacobmontiel (changes requested):
Thanks for the changes @smastelini 

These are very useful features for the trees. In general, this PR is looking good. I just left a couple of comments to address.

river/ensemble/boosting.py
smastelini: :)

river/ensemble/adaptive_random_forest.py
smastelini: I think so! Good news :D

river/utils/skmultiflow_utils.py
smastelini: I changed this function to return the rounded value directly. Let me explain its functioning:

This function helps us select only the significant digits in a long float (given the desired number of digits). 

For instance: 

- `round_sig_fig(0.0002345, significant_digits=2) -> 0.0002`
- `round(0.0002345, ndigits=2) -> 0.0`

On the other hand,

- `round_sig_fig(0.255, significant_digits=2) -> 0.26`
- `round(0.255, 2) -> 0.26`

It's a nice trick to keep significant decimal places and avoid creating non-meaningful rounding while drawing/printing the trees' structures. 

I always used a fixed number of decimal places to exhibit numbers in tables and similar stuff. That was until I performed a project in partnership with researchers of vitreous materials. They asked the guys from my lab and me to round the numbers to their significant figures. At first glance, the result seemed odd to me (as the resulting number of floating points is not the same for every number), but after second thought I realized the representation becomes more compact and concise.

river/tree/_base_tree.py
smastelini: What a fortunate MWE! It was a bug. HATC was not updating the set of seen classes. I fixed that in commit [d878086](https://github.com/creme-ml/creme/pull/376/commits/d878086d72b3d364c71d36075caeff11b49a602a).

river/tree/_nodes/base.py
smastelini: I did not run any benchmarks per se. This is the way creme traverses a path in the tree. In skmultiflow, we retrieve the leaf rather than using a generator. Perhaps @MaxHalford can shed some light on that matter. 

We might consider changing the strategy in the future. Note that in river we update only the statistics of the leaves upon sorting a new instance down the tree, whereas creme updates all the nodes in the path.

river/tree/_nodes/hatc_nodes.py
smastelini: You are totally right! And thank you so much for this comment. It helped spot some bugs in the Hoeffding Adaptive Tree that were not covered by our tests. I also updated this function as requested in commit [f2c6fad](https://github.com/creme-ml/creme/pull/376/commits/f2c6fad7fee9fa382de554c78bb20aa09a1e5033).

river/utils/skmultiflow_utils.py
jacobmontiel: Nice! And thanks for the explanation.

jacobmontiel (approved):
Implementation is correct and all comments have been properly addressed. The PR is ready.

PR #379, opened 2020-11-04
[tree] Make AO selection dynamic and expose internal adjustable parameters to the user

smastelini:
This PR makes the attribute observer (AO) selection configurable by the user in the trees.

In summary:
- The user can now select between three AOs: `bst`, `gaussian`, `histogram` (new option adapted from `creme`) in the classification trees (adaptive random forest included). They can also adjust the hyperparameters of these AOs, when applicable.
- The user can pass the minimum number of samples that each branch must have to make a split valid (regression trees).

Minor changes:
- Remove unnecessary usage of NumPy functions (applied in scalar data).
- Make some methods private.


smastelini:
This is my last PR before the release (probably hahaha).

IMHO there are only two missing pieces. So here go my questions:

@MaxHalford, I added a histogram-based attribute observer by leveraging the `NumericSplitEnum` implementation from creme. What do you say about the approximation of the `pdf` function I used in this AO? Does that sound reasonable at all?

Following, the specific portion of the code I mentioned:

https://github.com/smastelini/creme/blob/a2469806dca0e8245789eae3403111c6961f1122/river/tree/_attribute_observer/numeric_attribute_class_observer_histogram.py#L37-L48

Any suggestions are more than welcomed!

@jacobmontiel, the AO based on a binary search tree cannot calculate pdf function for a given value:

https://github.com/smastelini/creme/blob/a2469806dca0e8245789eae3403111c6961f1122/river/tree/_attribute_observer/numeric_attribute_class_observer_binary_tree.py#L65-L66

MOA also does not implement such a functionality. Hence, this AO cannot be used in conjunction with naive bayes leaves. I intend to refactor this AO if needed, but I think we need to decide on this portion before moving forward. Worst case scenario: we state in the documentation that this AO cannot be used when `leaf_prediction` is `nb` or `nba`. Ideally, I would like to provide (at least) a rough approximation, as I did with the histogram AO.

smastelini:
**Spoiler alert:** the `n_splits` parameter used when `attribute_observer='gaussian'` affects the performance **a lot!** In my tests the accuracy of the tree increased by almost 5% by simply changing it from `10 -> 64`.

MaxHalford:
@smastelini nice work!

It seems that you're using linear search to find the bin where a value belongs in a histogram. Surely it would be faster to use binary with the `bisect` module.

How's the performance of histogram vs. gaussian?

jacobmontiel:
> Worst case scenario: we state in the documentation that this AO cannot be used when leaf_prediction is nb or nba. Ideally, I would like to provide (at least) a rough approximation, as I did with the histogram AO.

TBH I have no issue if we limit the usage of this AO. If the other options are faster then go with that. The combination of a binary search and a "simplistic" leaf type(`MC`) can serve as a baseline.

My secondary concern is the investment to make to bring this functionality and test it vs its benefits. This can become an issue for the next cycle ;-)

jacobmontiel:
> **Spoiler alert:** the `n_splits` parameter used when `attribute_observer='gaussian'` affects the performance **a lot!** In my tests the accuracy of the tree increased by almost 5% by simply changing it from `10 -> 64`.

This is consistent with the expectations that for some data sets having more granularity will leave to better accuracy. On the other hand, it might be a compromise in processing time.

river/ensemble/adaptive_random_forest.py
```diff
@@ -372,6 +400,32 @@ class AdaptiveRandomForestClassifier(BaseForest, base.Classifier):
     nominal_attributes
         [*Tree parameter*] List of Nominal attributes. If empty, then assume that
         all attributes are numerical.
+    attribute_observer
+        [*Tree parameter*] The attribute observer (AO) algorithm used to monitor the class
```
jacobmontiel, line 404: This description should be in the `notes`section (now being rendered in the documentation). You can add a note here: "See notes for more information regarding this attribute."

river/ensemble/adaptive_random_forest.py
```diff
@@ -600,6 +660,21 @@ class AdaptiveRandomForestRegressor(BaseForest, base.Regressor):
     nominal_attributes
         [*Tree parameter*] List of Nominal attributes. If empty, then assume that
         all attributes are numerical.
+    attribute_observer
```
jacobmontiel, line 663: Same comment as before.

river/tree/_attribute_observer/numeric_attribute_class_observer_gaussian.py
```diff
@@ -11,14 +11,20 @@
...
+
+
+    Parameters
+    ----------
+    n_splits
+        The number of partitions to consider when querying for split candidates.
     """
 
-    def __init__(self, n_bins: int = 10):
+    def __init__(self, n_splits: int = 10):
```
jacobmontiel, line 22: `n_bins` is the standard way to refer to this parameter as it makes reference to "bins" where observations will be allocated. This is the way I have seen it named in multiple ml packages. However, I am open to discussing this.

river/tree/_attribute_observer/numeric_attribute_class_observer_histogram.py
```diff
@@ -0,0 +1,109 @@
...
+        # If only one single value has been observed, then no split can be proposed
+        if low >= high:
+            return
+
+        n_thresholds = min(
+            self.n_splits,
+            max(map(len, self.hists.values())) - 1
+        )
+
+        thresholds = list(decimal_range(start=low, stop=high, num=n_thresholds))
```
jacobmontiel, line 65: Why not return the list directly? Why implement it as a generator?

river/tree/extremely_fast_decision_tree.py
```diff
@@ -43,6 +43,32 @@ class ExtremelyFastDecisionTreeClassifier(HoeffdingTreeClassifier):
     nominal_attributes
         List of Nominal attributes identifiers. If empty, then assume that all numeric attributes
         should be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 46: Same comment as before.

river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -37,6 +37,32 @@ class HoeffdingAdaptiveTreeClassifier(HoeffdingTreeClassifier):
     nominal_attributes
         List of Nominal attributes. If empty, then assume that all numeric attributes should
         be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 40: Same here.

river/tree/hoeffding_adaptive_tree_regressor.py
```diff
@@ -45,6 +45,21 @@ class HoeffdingAdaptiveTreeRegressor(HoeffdingTreeRegressor):
     nominal_attributes
         List of Nominal attributes. If empty, then assume that all numeric attributes should
         be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 48: And here.

river/tree/hoeffding_tree_classifier.py
```diff
@@ -40,8 +40,34 @@ class HoeffdingTreeClassifier(BaseHoeffdingTree, base.Classifier):
     nb_threshold
         Number of instances a leaf should observe before allowing Naive Bayes.
     nominal_attributes
-        List of Nominal attributes identifiers. If empty, then assume that all numeric attributes
-        should be treated as continuous.
+        List of Nominal attributes identifiers. If empty, then assume that all numeric
```
jacobmontiel, line 43: Once more.

river/tree/hoeffding_tree_regressor.py
```diff
@@ -44,6 +44,21 @@ class HoeffdingTreeRegressor(BaseHoeffdingTree, base.Regressor):
     nominal_attributes
         List of Nominal attributes identifiers. If empty, then assume that all numeric attributes
         should be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 47: :-)

river/tree/isoup_tree_regressor.py
```diff
@@ -50,6 +50,21 @@ class iSOUPTreeRegressor(HoeffdingTreeRegressor, base.MultiOutputMixin):
     nominal_attributes
         List of Nominal attributes identifiers. If empty, then assume that all numeric attributes
         should be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 53: Just marking them to review later.

river/tree/label_combination_hoeffding_tree.py
```diff
@@ -41,6 +41,32 @@ class LabelCombinationHoeffdingTreeClassifier(HoeffdingTreeClassifier, base.Mult
     nominal_attributes
         List of Nominal attributes identifiers. If empty, then assume that all numeric attributes
         should be treated as continuous.
+    attribute_observer
```
jacobmontiel, line 44: To change.

jacobmontiel (commented):
No official review as this is a draft. I left a couple of comments.

river/tree/_attribute_observer/numeric_attribute_class_observer_gaussian.py
smastelini: I am not against using bins for the name. It makes sense to keep the current standards adopted by the research community.

I changed its name because it might confuse the user when comparing this AO with the Histogram one. The newly added Histogram AO has a parameter `n_bins` with a different purpose.

river/tree/_attribute_observer/numeric_attribute_class_observer_histogram.py
smastelini: That's a good question that I'll reserve for @MaxHalford, hahaha.

I simply leveraged the implementation from creme ;-)

smastelini:
> It seems that you're using linear search to find the bin where a value belongs in a histogram. Surely it would be faster to use binary with the bisect module.

Good catch! I'll change that.

> How's the performance of histogram vs. gaussian?

I didn't run many tests to be confident in my conclusions. But so far I've seen that:
- when using the majority class, the results approximate the ones of the Gaussian estimator. With the right combination of bins and n_splits we might get better results. That should be specially the case if the feature distribution is multimodal or skewed.
- when naive bayes is used, the results seem to drop. As you discuss [here](https://github.com/creme-ml/creme/projects/2#card-40799448), the current approximation of the `pdf` could to be a problem. It might be worth investigating in the future (and I have interest in that)  incremental KDE or GMM algorithms to bring to river's collection.

river/tree/_attribute_observer/numeric_attribute_class_observer_histogram.py
MaxHalford: No reason in particular, apart from the fact that Python's `range` function is a generator.

smastelini:
> TBH I have no issue if we limit the usage of this AO. If the other options are faster then go with that. The combination of a binary search and a "simplistic" leaf type(MC) can serve as a baseline.

I agree. I will add notes in the documentation to cover this possibility. The BST observer also needs some minor fixes. I'll address that later today. Since the pdf of any value is `0` with BST, the naive bayes leaves when coupled with this observer will rely only on the prior (`P(y)`) to compute the predictions. Hence, the trees will end up using some sort of majority class classifier in the end.

> My secondary concern is the investment to make to bring this functionality and test it vs its benefits. This can become an issue for the next cycle ;-)

You are right. I will limit the scope make the AO work minimally. Just ensuring that it can be used, even if as a baseline.

river/ensemble/adaptive_random_forest.py
smastelini: Updated in commit [d247eb7](https://github.com/creme-ml/creme/pull/379/commits/d247eb7b873f17442e35e1b1df8054ad05611a4b).

river/tree/extremely_fast_decision_tree.py
smastelini: Updated in commit [d247eb7](https://github.com/creme-ml/creme/pull/379/commits/d247eb7b873f17442e35e1b1df8054ad05611a4b).

river/tree/hoeffding_adaptive_tree_classifier.py
smastelini: Updated in commit [d247eb7](https://github.com/creme-ml/creme/pull/379/commits/d247eb7b873f17442e35e1b1df8054ad05611a4b).

river/tree/hoeffding_tree_classifier.py
smastelini: Updated in commit [d247eb7](https://github.com/creme-ml/creme/pull/379/commits/d247eb7b873f17442e35e1b1df8054ad05611a4b).

river/tree/label_combination_hoeffding_tree.py
smastelini: Updated in commit [d247eb7](https://github.com/creme-ml/creme/pull/379/commits/d247eb7b873f17442e35e1b1df8054ad05611a4b).

river/ensemble/adaptive_random_forest.py
smastelini: Updated in commit [f1eab02](https://github.com/creme-ml/creme/pull/379/commits/f1eab0274d21fedd3f599f8597e6a2f0230cbe26).

river/tree/hoeffding_adaptive_tree_regressor.py
smastelini: Updated in commit [f1eab02](https://github.com/creme-ml/creme/pull/379/commits/f1eab0274d21fedd3f599f8597e6a2f0230cbe26).

river/tree/hoeffding_tree_regressor.py
smastelini: Updated in commit [f1eab02](https://github.com/creme-ml/creme/pull/379/commits/f1eab0274d21fedd3f599f8597e6a2f0230cbe26).

river/tree/isoup_tree_regressor.py
smastelini: Updated in commit [f1eab02](https://github.com/creme-ml/creme/pull/379/commits/f1eab0274d21fedd3f599f8597e6a2f0230cbe26).

river/tree/_attribute_observer/numeric_attribute_class_observer_gaussian.py
jacobmontiel: To some extent, the meaning of "bins" is similar in both AO. I don't think this would be an issue.

jacobmontiel (approved):
Changes are correct. Thanks @smastelini for the improvements.

jacobmontiel:
Waiting for the conflicts to be resolved.

PR #385, opened 2020-11-12
Update notebooks

jacobmontiel:
- Solve compatibility issues
- Update text to reflect latest changes
- Cleanup references to obsolete methods/modules


docs/examples/batch-to-online.ipynb
```diff
@@ -114,7 +114,13 @@
    "outputs": [
     {
      "data": {
-      "text/plain": "array([7.760e+00, 2.454e+01, 4.792e+01, 1.810e+02, 5.263e-02, 4.362e-02,\n       0.000e+00, 0.000e+00, 1.587e-01, 5.884e-02, 3.857e-01, 1.428e+00,\n       2.548e+00, 1.915e+01, 7.189e-03, 4.660e-03, 0.000e+00, 0.000e+00,\n       2.676e-02, 2.783e-03, 9.456e+00, 3.037e+01, 5.916e+01, 2.686e+02,\n       8.996e-02, 6.444e-02, 0.000e+00, 0.000e+00, 2.871e-01, 7.039e-02])"
+      "text/plain": [
```
smastelini, line 117: It's a little bit difficult to review notebooks, hahaha.

But here we go. I took some time to read the introductory material, and I must say I am impressed by its quality! Wonderful work, guys!

I found, perhaps, a small caveat in this one.

At some point, this operation is used:

`xi_scaled = scaler.learn_one(xi).transform_one(xi)`

But previously, the scaler was used differently (as stated in the text):


> In river each call to a Transformer's `learn_one` will return the transformed output. Meanwhile, calling `learn_one` with a Classifier or a Regressor will return the updated model with its internal state modified by the given set of features. The twist is that a prediction is made before looking at the true target `y`.

and, then:

```python

scaler = preprocessing.StandardScaler()

for xi, yi in stream.iter_sklearn_dataset(datasets.load_breast_cancer()):
    xi = scaler.learn_one(xi)
```

Or this last statement is correct, or `xi` would point to `scaler` (if we follow the logic of the first code strip).

MaxHalford: Actually that paragraph:

> In river each call to a Transformer's learn_one will return the transformed output. Meanwhile, calling learn_one with a Classifier or a Regressor will return the updated model with its internal state modified by the given set of features. The twist is that a prediction is made before looking at the true target y.

is wrong! I guess I never changed it.

docs/examples/bike-sharing-forecasting.ipynb
```diff
@@ -460,9 +460,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The score we now have is much more realistic, as it is comparable with [related data science competitions](https://maxhalford.github.io/blog/a-short-introduction-and-conclusion-to-the-openbikes-2016-challenge/). Moreover, we can see that the model gets better with time, which feels better than the previous situations. The point is that `online_qa_score` method can be used to simulate a production scenario, and is thus extremely valuable.\n",
+    "The score we now have is much more realistic, as it is comparable with [related data science competitions](https://maxhalford.github.io/blog/a-short-introduction-and-conclusion-to-the-openbikes-2016-challenge/). Moreover, we can see that the model gets better with time, which feels better than the previous situations. The point is that `progressive_val_score` method can be used to simulate a production scenario, and is thus extremely valuable.\n",
```
smastelini, line 463: This link seems to be broken

MaxHalford: Yes, it's actually https://maxhalford.github.io/blog/openbikes-challenge/

docs/examples/batch-to-online.ipynb
jacobmontiel: What is the fix for this issue?

MaxHalford: Don't worry I took care of it, I need to update the example and the text. This intro is really old, I wrote roughly 3 months after I created creme. It might be worth revamping it.

jacobmontiel: Cool, thanks for the help :-)

PR #387, opened 2020-11-13
Regression generators

smastelini:
Adds the following new regression synthetic data generators:

- FriedmanDrift, as described in this [paper](https://link.springer.com/article/10.1007/s10618-010-0201-y) (the one that introduces the well-known FIMT-DD tree regressor). It implements three variants of concept drift
- Planes2D (originally 2DPlanes)
- Mv

The last two generators are described in the [Luís Torgo](https://www.dcc.fc.up.pt/~ltorgo/Regression/DataSets.html) repository of regression datasets and were used in multiple research papers in the last years.

Moreover, this PR also modifies `river.synth.Friedman` by adding a gaussian noise to the label, as originally proposed in the [original paper](https://www.jstor.org/stable/2241837?seq=1).

### DIFF
--- a/river/datasets/synth/friedman.py
+++ b/river/datasets/synth/friedman.py
@@ -10,7 +10,7 @@ class Friedman(base.SyntheticDataset):
     Each observation is composed of 10 features. Each feature value is sampled uniformly in [0, 1].
     The target is defined by the following function:
 
-    $$y = 10 sin(\\pi x_0 x_1) + 20 (x_2 - .5)^2 + 10 x_3 + 5 x_4 + \\epsilon$$
+    $$y = 10 sin(\\pi x_0 x_1) + 20 (x_2 - 0.5)^2 + 10 x_3 + 5 x_4 + \\epsilon$$
 
     In the last expression, $\\epsilon \\sim \\mathcal{N}(0, 1)$, is the noise. Therefore,
     only the first 5 features are relevant.
@@ -71,14 +71,14 @@ class FriedmanDrift(base.SyntheticDataset):
     ----------
     drift_type
         The variant of concept drift.</br>
-        - 'lea': Local Expanding Abrupt drift. The concept drift appears in two distinct
+        - `'lea'`: Local Expanding Abrupt drift. The concept drift appears in two distinct
         regions of the instance space, while the remaining regions are left unaltered.
         There are three points of abrupt change in the training dataset.
         At every consecutive change the regions of drift are expanded.</br>
-        - 'gra': Global Recurring Abrupt drift. The concept drift appears over the whole
+        - `'gra'`: Global Recurring Abrupt drift. The concept drift appears over the whole
         instance space. There are two points of concept drift. At the second point of drift
         the old concept reoccurs.</br>
-        - 'gsg': Global and Slow Gradual drift. The concept drift affects all the instance
+        - `'gsg'`: Global and Slow Gradual drift. The concept drift affects all the instance
         space. However, the change is gradual and not abrupt. After each one of the two change
         points covered by this variant, and during a window of length `transition_window`,
         examples from both old and the new concepts are generated with equal probability.
--- a/river/datasets/synth/friedman.py
+++ b/river/datasets/synth/friedman.py
@@ -204,7 +204,7 @@ class FriedmanDrift(base.SyntheticDataset):
     def __lea_in_r1(self, x, index):
         if index < self.change_point1:
             return False
-        if self.change_point1 <= index < self.change_point2:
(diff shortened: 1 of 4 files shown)


river/datasets/synth/friedman.py
```diff
@@ -51,6 +52,230 @@ def __iter__(self):
...
+        At every consecutive change the regions of drift are expanded.</br>
+        - `'gra'`: Global Recurring Abrupt drift. The concept drift appears over the whole
+        instance space. There are two points of concept drift. At the second point of drift
+        the old concept reoccurs.</br>
+        - `'gsg'`: Global and Slow Gradual drift. The concept drift affects all the instance
+        space. However, the change is gradual and not abrupt. After each one of the two change
+        points covered by this variant, and during a window of length `transition_window`,
+        examples from both old and the new concepts are generated with equal probability.
+        After the transition period, only the examples from the new concept are generated.
+    change_point1
```
jacobmontiel, line 86: This can be a single parameter:
```python
position: tuple = (point_1, point_2, point_3)
```

river/datasets/synth/friedman.py
```diff
@@ -51,6 +52,230 @@ def __iter__(self):
...
 
             x = {i: rng.uniform(a=0, b=1) for i in range(10)}
-            y = 10 * math.sin(math.pi * x[0] * x[1]) + 20 * (x[2] - .5) ** 2 + 10 * x[3] + 5 * x[4]
+            y = 10 * math.sin(math.pi * x[0] * x[1]) + 20 * (x[2] - .5) ** 2 + 10 * x[3] \
+                + 5 * x[4] + rng.gauss(mu=0, sigma=1)
 
             yield x, y
+
+
+class FriedmanDrift(base.SyntheticDataset):
```
jacobmontiel, line 61: Should this class extend the `Friedman` class?

river/datasets/synth/planes_2d.py
```diff
@@ -0,0 +1,73 @@
...
+    """2D Planes synthetic dataset.
+
+    This dataset is described in [^1] and was adapted from [^2]. The features are generated
+    using the following probabilities:
+
+    $$P(x_1 = -1) = P(x_1 = 1) = \\frac{1}{2}$$
+
+    $$P(x_m = -1) = P(x_m = 0) = P(x_m = 1) = \\frac{1}{3}, m=2,\\ldots, 10$$
+
+    The target value is defined by using rule:
```
jacobmontiel, line 16: Suggested rewording: "defined by the following rule"

jacobmontiel (commented):
The implementation looks good to me.

There are just minor comments. Let me know if you need further information.

jacobmontiel:
It might be worth exploring the possibility of expanding the `ConceptDriftStream` generator to account for the kinds of drift in the `FriedmandDrift` generator. This is outside of the scope of this PR but we could create a new issue for this purpose. Comments @smastelini ?

river/datasets/synth/friedman.py
smastelini: This simplifies the usage for the user, indeed. I changed how the concept drift locations are set in commit [afa307d](https://github.com/online-ml/river/pull/387/commits/afa307da973ab1519b2d06ae3852ba566f5ee147), as suggested.

river/datasets/synth/friedman.py
smastelini: Well, apart from clearly stating that `FriedmanDrift` is an extension of `Friedman`, there are no immediate gains. However, I think this is enough reason by itself.

Changed in commit commit [afa307d](https://github.com/online-ml/river/pull/387/commits/afa307da973ab1519b2d06ae3852ba566f5ee147).

river/datasets/synth/planes_2d.py
smastelini: Reworded in commit [afa307d](https://github.com/online-ml/river/pull/387/commits/afa307da973ab1519b2d06ae3852ba566f5ee147).

smastelini:
Thanks for the review, @jacobmontiel! I addressed your comments in the last commit.

> It might be worth exploring the possibility of expanding the ConceptDriftStream generator to account for the kinds of drift in the FriedmandDrift generator. This is outside of the scope of this PR but we could create a new issue for this purpose. Comments @smastelini?

I think they are interesting cases to explore in the future. I am just not sure how straightforward it would be to apply these cases for generic data generators. FriedmanDrift was specifically crafted for regression tasks and the Friedman synthetic case. For sure, we can create an issue to evaluate the applicability of these drift variants to other tasks.

Please, let me know if you need more changes :)

jacobmontiel:
Thanks for the latest changes @smastelini 

The implementation is correct.

PR #390, opened 2020-11-22
Changed iter_array returning y when y=None

mfsuve:
For streaming from data, when y is None it is unnecessarily returned. If the function only gets X it is expected to return only X, not y.

### DIFF
--- a/river/stream/iter_array.py
+++ b/river/stream/iter_array.py
@@ -71,5 +71,5 @@ def iter_array(X: np.ndarray, y: np.ndarray = None,
 
     else:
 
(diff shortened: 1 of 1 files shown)


MaxHalford:
Thanks for the PR! The idea is good but the logic is wrong. You need to add a check to see if `y` is `None` and then yield the `x`s. In other words there's 3 cases:

1. `y` is None
2. `y` is multi-output
3. `y` is single-output

 Do you see what I mean? Also you need to make unit tests pass. Tell me if something isn't clear.

Issue #270, opened 2020-03-22
Implement bandits


MaxHalford:
We now have `SuccessiveHalvingClassifier` and `SuccessiveHalvingRegressor` in the `model_selection` to module to perform, well, model selection. This allows doing hyperparameter-tuning by initializing a model with different parameter configuration and running them against each other. All in all it seems to be working pretty well and we should be getting some feedback on it soon. The implementations are handy because they implement the `fit_one`/`predict_one` interface and therefore make the whole process transparent to users. In other words you can use them as you would any other model. This design will go a long way and should make things simple in terms of deployment (I'm thinking of you [chantilly](https://github.com/creme-ml/chantilly)).

The next step would be to implement [multi-armed bandits](https://lilianweng.github.io/lil-log/2018/01/23/the-multi-armed-bandit-problem-and-its-solutions.html). In progressive validation, all the remaining models are updated. This is called a "full feedback" situation. Bandits, on the other hand, use partial feedback, because only one model is picked and trained at a time. This is more efficient because it results in less model evaluations, but might also converge more slowly. Most bandit algorithms assume that the performance of the models is constant through time (this is called "stationarity"). However, the performance of each model is bound to change through time because the act of picking modifies the model. Therefore ideally we need to looking into non-stationary bandit algorithms. [Here](https://smpybandits.github.io/NonStationaryBandits.html) are some more details and references.

Here are the algortithms I would like to see implemented:

- [Epsilon greedy](https://lilianweng.github.io/lil-log/2018/01/23/the-multi-armed-bandit-problem-and-its-solutions.html#%CE%B5-greedy-algorithm)
- [UCB](https://banditalgs.com/2016/09/18/the-upper-confidence-bound-algorithm/)
- [Discounted UCB](https://www.wikiwand.com/en/Multi-armed_bandit#/Non-stationary_bandit)
- [Sliding-Window UCB](https://www.wikiwand.com/en/Multi-armed_bandit#/Non-stationary_bandit)

Also see [this paper](https://arxiv.org/pdf/2003.10113.pdf). We can probably open separate issues for each implementation. I think that the current best way of proceding is to provide one implementation for regression and one for classification in each case, much like what is done for successive halving. There's also [this paper](https://arxiv.org/pdf/1306.0686.pdf) that discusses delayed feedback and how it affects bandits.

etiennekintzler:
[...]

PR #391, opened 2020-11-22
[WIP] Bandits regressors for model selection

etiennekintzler:
## Description

The PR  introduce bandits (epsilon-greedy and UCB) for model selection (see issue #270 ). The PR concerns only regressors, but I can add the classifiers in a subsequent PR.

The use of the classes are straightforward : 

```python
bandit = UCBRegressor(models=models, metric=metrics.MSE()),

for (x, y) in data.take(N):    
        y_pred = bandit.predict_one(x=x)
        bandit.learn_one(x=x, y=y)

best_model = bandit.best_model
```

There are convenience methods such as : 
- `percentage_pulled` : to get the percentage each arm was pulled
- `best_model` : return the model with the highest average reward

Also I added a method `add_models` where the user can add models on the fly.

I am also working on a [notebook](https://github.com/etiennekintzler/workbench/blob/bandits/Clean_Bandits.ipynb) that studies the behavior of the bandits for model selection. The notebook also include Exp3, which seems promising but has numerical stability issue and yields counter-intuitive results (see section 3 of the NB). That's why I kept it out of this PR. More generally, the performances of UCB and epsilon-greedy are rather good but there seems to be some variance in the performance.

## Improvements
It's still WIP on the following points : 
- docstring, mainly add examples + cleaning.
- typing
- some comments might be removed.
- certain functionalities I am not sure to include or not:
    - Save the history of metric value (cf `save_metrics` parameter) in `Bandits`
- the name of the classes and the methods are open for changes

I would appreciate inputs regarding the functionalities and the methods/classes naming.


river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+            str(self),
+            str(self.metric),
+            "Best model id: " + str(self._best_model_idx),
+            sep="\n\t"
+        )
+
+
+class EpsilonGreedyBandit(Bandit):
+    def __init__(self, epsilon=0.1, epsilon_decay=None, **kwargs):
+        super().__init__(**kwargs)
```
etiennekintzler, line 147: This part seems to make the tests fail (same part for UCBBandit). Should I enumerate all the parameters (that are in `Bandit.__init__`) instead of using `**kwargs` ?

river/expert/__init__.py
```diff
@@ -16,15 +16,22 @@
 
 """
 
+from .bandit import (
+     EpsilonGreedyRegressor,
+     UCBRegressor,
+)
+
```
MaxHalford, line 23: The current convention is:

```py
from .bandit import EpsilonGreedyRegressor
from .bandit import UCBRegressor
```

river/expert/__init__.py
```diff
@@ -16,15 +16,22 @@
...
 from .ewa import EWARegressor
 from .sh import SuccessiveHalvingClassifier
 from .sh import SuccessiveHalvingRegressor
 from .stacking import StackingClassifier
 
 
 __all__ = [
+    "EpsilonGreedyRegressor",
```
MaxHalford, line 31: Single quotes

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+
+import numpy as np
+
+from river import base
+from river import metrics
+from river import preprocessing
+
+
+__all__ = [
+    "EpsilonGreedyRegressor",
```
MaxHalford, line 13: Single quotes

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
+import abc
+import copy
+import typing
+
+import numpy as np
```
MaxHalford, line 5: In general, we prefer to use data types that are native to Python. For instance, we've found that the `random` module is faster than `np.random` for small sample sizes. Also numpy arrays are slower than Python lists for most small tasks. I don't mind numpy per se, but I would only use it for linear algebra, not for storing variables. The philosophy of River is to stay as close as possible to the Python standard library :)

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+        reward = self._compute_scaled_reward(y_pred=y_pred, y_true=y)
+        self._n_iter += 1
+        self._N[chosen_arm] += 1
+        self._average_reward[chosen_arm] += (1.0 / self._N[chosen_arm]) * \
+                                            (reward - self._average_reward[chosen_arm])
+
+        # Specific update of the arm for certain bandit class
+        self._update_arm(chosen_arm, reward)
+
+        if self.print_every:
```
MaxHalford, line 101: We don't to be printing stuff in the code. If you really want to add the possibility to log progress, I would advise using the `logging` module.

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+            if (self._n_iter % self.print_every) == 0:
+                self._print_info()
+
+        if self.save_percentage_pulled:
+            self.store_percentage_pulled += [self.percentage_pulled]
+
+        if self.save_metric_values:
+            self.metric_values += [self.metric._eval(y_pred, y)]
+
+        return self
```
MaxHalford, line 111: I think that a good design would be to have a `_learn_one` method with returns `self.percentage_pulled` and `self.metric._eval(y_pred, y)`. That way, power users can use `_learn_one` to collect information and display/analyse it in some way. I don't think we should be storing a trace of the values.

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def add_models(self, new_models):
+        if not isinstance(new_models, list):
+            raise TypeError("Argument `new_models` must be of a list")
```
MaxHalford, line 115: It's not Pythonic to check like this, just add a type hint to `new_models` and expect the user to read the documentation :)

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+        metric_value = self.metric._eval(y_pred, y_true)
+        metric_to_reward_dict = {
+            "metric": metric_value if self.metric.bigger_is_better else (-1) * metric_value
+        }
+        if update_scaler:
+            self.reward_scaler.learn_one(metric_to_reward_dict)
+        reward = self.reward_scaler.transform_one(metric_to_reward_dict)["metric"]
+        return reward
+
+    def _print_info(self):
```
MaxHalford, line 136: Maybe this could be in `__repr__`?

river/expert/bandit.py
MaxHalford: Yes, you should :)

river/expert/bandit.py
```diff
@@ -0,0 +1,269 @@
...
+        if not self.reward_scaler:
+            self.reward_scaler = preprocessing.StandardScaler()
+
+    def _pull_arm(self):
+        if np.random.rand() > self.epsilon:
+            chosen_arm = np.argmax(self._average_reward)
+        else:
+            chosen_arm = np.random.choice(self._n_arms)
+
+        return chosen_arm
```
MaxHalford, line 161: Again, a lot of this could be implemented in Python, which I suspect would be faster.

MaxHalford (commented):
Thanks for the PR! It's a good start, and I agree that starting with regression only is a good idea :). I've made a few comments to get the ball rolling. High five!

river/expert/bandit.py
etiennekintzler: Actually I try with it but the tests do fail again. I see also that expert.SuccessiveHalving* are purposely exclude from the test (in `get_all_estimators`) while `expert.EWARegressor` is not. Is the kind of regressor I implemented supposed to pass this test at all ?

MaxHalford: You're right, it's not supposed to pass the tests by default. You need implement a `_default_params` method, like what is done [here](https://github.com/online-ml/river/blob/d3e56400a0ece7113f454d26d46fe18139a19a21/river/multiclass/ovo.py#L73). This method will get called [here](https://github.com/online-ml/river/blob/master/river/test_.py#L96) to initialize the state of your model.

river/expert/bandit.py
etiennekintzler: Ok ! 

If necessary I can get rid of the numpy dependency completely but I'll have to implement argmax (which is not part of the Python stdlib).

river/expert/bandit.py
etiennekintzler: IMO the most pythonic way is to use try-catch cf EAFP (easier to ask for forgiveness than permission, see https://docs.python.org/3.6/glossary.html).

But if you prefer I can just add the type hint !

river/expert/bandit.py
etiennekintzler: Yes I agree, it's more clean to use `logging` than to clog the class with this functionality.

river/expert/bandit.py
MaxHalford: If you're just doing this in one place you can do:

```py
def argmax(l):
    return max(range(len(l)), key=l.__getitem__)
```

Either make this a function or put inline and add a comment :)

river/expert/bandit.py
etiennekintzler: yes, sure !

river/expert/bandit.py
etiennekintzler: Ok thank you !

Also `math.log`, `math.sqrt` etc are not vectorized by default. Should I use listcomp instead ?

river/expert/bandit.py
etiennekintzler: Ok, tell me if the way I've done it suits you

river/expert/bandit.py
etiennekintzler: I end up adding type hint

river/expert/bandit.py
etiennekintzler: Actually I think `argmax` should be put in `river.util.math` ; what do you think ?

etiennekintzler:
> Thanks for the PR! It's a good start, and I agree that starting with regression only is a good idea :). I've made a few comments to get the ball rolling. High five!

Thanks, also thank you for this review ! 

I incorporate all your points and left some _conversation_ unresolved (when your feedback is needed).

I will quickly test this new class since I introduce noticeable change by removing the numpy dependency. Also I need to finish writing docstring (Parameters and Example) then you can merge if it looks good to you. If you see other desirable change tell me too.

river/expert/bandit.py
etiennekintzler: Also, I just add default LinearRegression in a Pipeline and tests pass. It somehow surprise me since I didn't gave default arg for `metric` and `reward_scaler`.

**Not related to this PR** but one test failed : 

```python
======================================================================== FAILURES ========================================================================
___________________________________________________ [doctest] river.neighbors.sam_knn.SAMKNNClassifier ___________________________________________________
052     >>> from river import metrics
053     >>> from river import neighbors
054 
055     >>> dataset = synth.ConceptDriftStream(position=500, width=20, seed=1).take(1000)
056 
057     >>> model = neighbors.SAMKNNClassifier(window_size=100)
058 
059     >>> metric = metrics.Accuracy()
060 
061     >>> evaluate.progressive_val_score(dataset, model, metric)
Expected:
    Accuracy: 59.90%
Got:
    Accuracy: 58.20%

/path/to/package/river/neighbors/sam_knn.py:61: DocTestFailure
```

MaxHalford: @etiennekintzler any reason why you put the linear regression in a pipeline? Yes it's a bit weird that the tests are passing if those are required arguments :/. Could be worth investigating.

@smastelini any idea why the above test is failing?

river/expert/bandit.py
MaxHalford: > Also `math.log`, `math.sqrt` etc are not vectorized by default. Should I use listcomp instead?

Yes a list comprehension you should do fine. You can quickly test by yourself the performance versus numpy.

> Actually I think argmax should be put in river.util.math; what do you think?

Agreed!

river/expert/bandit.py
MaxHalford: Yes it looks good. But now that I think about it, you don't need to return `percentage_pulled` as it's already a property. Maybe you can return the pulled model? I guess the `_learn_one` method will change in the future depending on downstream usage. Also can just remove the `self` in the return.

river/expert/bandit.py
etiennekintzler: @MaxHalford Because most of the time I've used `models` that were `Pipeline` (that includes scaler and regressor), but actually it should work the same if I just put regressor.

etiennekintzler: I do confirm that it works without putting it in Pipeline (as it should since I just use the `predict_one` and `learn_one` which are defined for both Pipeline and Regressors).

smastelini: > @smastelini any idea why the above test is failing?

I could not reproduce that failure in my local setup (in the `master` branch). Pinging @jacobmontiel, since he has more experience with this specific k-NN variant than I do.

etiennekintzler: @MaxHalford @smastelini 

> I just add default LinearRegression in a Pipeline and tests pass. It somehow surprise me since I didn't gave default arg for metric and reward_scaler.

It passes on my local setup using `pytest` but it did fail on the CI pipeline (https://travis-ci.org/github/online-ml/river/jobs/745694837). It fails because I gave the wrong name for the argument name. If it had gone further it would have failed as well since I didn't provide default argument for parameters `metric` and `reward_scaler`

etiennekintzler: > > @smastelini any idea why the above test is failing?
> 
> I could not reproduce that failure in my local setup (in the `master` branch). Pinging @jacobmontiel, since he has more experience with this specific k-NN variant than I do.

FYU the `master` branch in my fork is 10 commits behind `master` in `online-ml/river` however these 10 commits are mostly doc so..

MaxHalford: > It passes on my local setup using pytest but it did fail on the CI pipeline (https://travis-ci.org/github/online-ml/river/jobs/745694837).

Very weird! Not sure what to say. Are you sure your development setup is correct?

etiennekintzler: @MaxHalford 

My bad :/ I forget that I add the bandits classes in the `ignored` tuple in `get_all_estimators`.

I can now confirm that it fails as in the CI when I do **not** provide the arguments.

river/expert/bandit.py
etiennekintzler: done

etiennekintzler:
The test `check_shuffle_features_no_impact` from module `estimator_checks` fails. The reason is not because of the shuffling of the features _per se_ but because **the bandit learning process is not deterministic**. 

To put it differently, if there is two arms (A1, A2) and two bandits (B1, B2), and if the bandit B1 pull arm A1 and bandit B2 pull arm A2 at round 0, they won't output the same prediction at round 1 because their internals (in particular `_average_reward`) will differ.

river/expert/bandit.py
jacobmontiel: > > @smastelini any idea why the above test is failing?

> I could not reproduce that failure in my local setup (in the master branch). Pinging @jacobmontiel, since he has more experience with this specific k-NN variant than I do.

I am confused, is this failing?

MaxHalford:
@etiennekintzler, can you close and reopen the pull request? We've switched to GitHub actions for CI, which should be much faster and more enjoyable than Travis :)

river/expert/bandit.py
etiennekintzler: I clone and rebuild the whole package (following https://github.com/online-ml/river/blob/master/CONTRIBUTING.md#installation) and it fails at the same place:

```python
======================================================================== FAILURES ========================================================================
___________________________________________________ [doctest] river.neighbors.sam_knn.SAMKNNClassifier ___________________________________________________
052     >>> from river import metrics
053     >>> from river import neighbors
054 
055     >>> dataset = synth.ConceptDriftStream(position=500, width=20, seed=1).take(1000)
056 
057     >>> model = neighbors.SAMKNNClassifier(window_size=100)
058 
059     >>> metric = metrics.Accuracy()
060 
061     >>> evaluate.progressive_val_score(dataset, model, metric)
Expected:
    Accuracy: 59.90%
Got:
    Accuracy: 58.20%

/home/etienne/PROJETS/river_test/river/river/neighbors/sam_knn.py:61: DocTestFailure

```

 I am using Python 3.7.7 and Ubuntu 20.04.1 LTS. 

I can open an issue to keep it out of the PR

MaxHalford:
@etiennekintzler: you need to a seed parameter and seed the random generation process to make it reproducible.

PR #397, opened 2020-11-26
Bandits regressors for model selection (new PR to use Github CI/CD)

etiennekintzler:
## Description

The PR  introduce bandits (epsilon-greedy and UCB) for model selection (see issue #270 ). The PR concerns only regressors, but I can add the classifiers in a subsequent PR.

The use of the classes are straightforward : 

```python
bandit = UCBRegressor(models=models, metric=metrics.MSE()),

for (x, y) in data.take(N):    
        y_pred = bandit.predict_one(x=x)
        bandit.learn_one(x=x, y=y)

best_model = bandit.best_model
```

There are convenience methods such as : 
- `percentage_pulled` : to get the percentage each arm was pulled
- `best_model` : return the model with the highest average reward

Also I added a method `add_models` where the user can add models on the fly.

I am also working on a [notebook](https://github.com/etiennekintzler/workbench/blob/bandits/Clean_Bandits.ipynb) that studies the behavior of the bandits for model selection. The notebook also include Exp3, which seems promising but has numerical stability issue and yields counter-intuitive results (see section 3 of the NB). That's why I kept it out of this PR. More generally, the performances of UCB and epsilon-greedy are rather good but there seems to be some variance in the performance.

## Improvements
It's still WIP on the following points : 
- docstring, mainly add examples + cleaning.
- some comments might be removed.
- the name of the classes and the methods are open for changes


etiennekintzler:
As in #391 :

The test `check_shuffle_features_no_impact` from module `estimator_checks` fails. The reason is not because of the shuffling of the features _per se_ but because **the bandit learning process is not deterministic**. 

To put it differently, if there is two arms (A1, A2) and two bandits (B1, B2), and if the bandit B1 pull arm A1 and bandit B2 pull arm A2 at round 0, they won't output the same prediction at round 1 because their internals (in particular `_average_reward`) will differ.

smastelini:
> The test check_shuffle_features_no_impact from module estimator_checks fails. The reason is not because of the shuffling of the features per se but because the bandit learning process is not deterministic.

Hi @etiennekintzler. Some while ago I had to set the `seed` parameter of some methods in order to make them pass the tests. The default seed parameter must be added in the `_default_params` method. Please, take a look at [that](https://github.com/online-ml/river/blob/4466595fcae1d6b6843dc874637f89354bacc43b/river/tree/hoeffding_adaptive_tree_regressor.py#L264-L266) as an example.

MaxHalford:
Indeed @etiennekintzler, as @smastelini is saying, you need to seed the randomized parts of your code. For instance, instead of calling `random.random()`, first initialize a random number generator in your `__init__`:

```py
self._rng = random.Random(seed)
```

where `seed` is a parameter provided by the user that defaults to `None`.

You can then do `self._rng.random()`.

etiennekintzler:
> Indeed @etiennekintzler, as @smastelini is saying, you need to seed the randomized parts of your code. For instance, instead of calling `random.random()`, first initialize a random number generator in your `__init__`:
> 
> ```python
> self._rng = random.Random(seed)
> ```
> 
> where `seed` is a parameter provided by the user that defaults to `None`.
> 
> You can then do `self._rng.random()`

@MaxHalford I misunderstood when you first talk about this in the 1st PR. I thought you meant fix a different seed for each bandit (for `model` and `shuffled` as named in the test) which would not solve the problem. You are talking about randomizing in the `_defaut_params` so the 2 bandits will pull the same arm ?

> > The test check_shuffle_features_no_impact from module estimator_checks fails. The reason is not because of the shuffling of the features per se but because the bandit learning process is not deterministic.
> 
> Hi @etiennekintzler. Some while ago I had to set the `seed` parameter of some methods in order to make them pass the tests. The default seed parameter must be added in the `_default_params` method. Please, take a look at [that](https://github.com/online-ml/river/blob/4466595fcae1d6b6843dc874637f89354bacc43b/river/tree/hoeffding_adaptive_tree_regressor.py#L264-L266) as an example.

Hello @smastelini, I check your code, I think I got it now, thanks you :+1:

river/expert/bandit.py
```diff
@@ -23,7 +24,7 @@
 
 class Bandit(base.EnsembleMixin):
 
-    def __init__(self, models: typing.List[base.Estimator], metric: metrics.Metric, reward_scaler: base.Transformer):
+    def __init__(self, models: typing.List[base.Estimator], metric: metrics.Metric, reward_scaler: base.Transformer, seed=None):
```
MaxHalford, line 27: Can you put `seed: int = None` everywhere `seed` is declared as a parameter? :)

river/expert/bandit.py
```diff
@@ -193,7 +198,8 @@ def _default_params(cls):
                 linear_model.LinearRegression(intercept_lr=.01)
             ],
             'metric': metrics.MSE(),
-            'reward_scaler': preprocessing.StandardScaler()
+            'reward_scaler': preprocessing.StandardScaler(),
+            'seed': 1
```
MaxHalford, line 202: No need to set the seed in the `default_params`. At least that's the current convention.

MaxHalford (commented):
Looking good!

river/expert/bandit.py
etiennekintzler: Ok ! I thought it was necessary for the tests

etiennekintzler:
> Looking good!

Thanks :)

Regarding the example in the docstring what dataset in `river` would you recommend for model selection ?

MaxHalford:
Not sure, but something like logistic regression / Hoeffding tree / GausianNB sounds good

etiennekintzler:
You are talking about methods nope ? (I was talking about dataset) 

Regarding methods I would have like to use online PCA (to do selection on the number of components) but doesn't seem to exist yet in river (just saw it on #3 )

MaxHalford:
Lol my bad: use Phishing for binary classification, ImageSegments for multi-class, and TrumpApproval for regression :). Indeed, we haven't implemented online PCA yet :)

MaxHalford:
Hey @etiennekintzler :). One thing you have to look into is a default values. For instance, `metric` and `reward_scaler` should have a default value IMO. You can put the default value to `None` and instantiate default values in the `__init__` value of each bandit. Check out how the `optimizer` is set in `LogisticRegression` to see how this is done :)

etiennekintzler:
Hello @MaxHalford 

Yes I've been thinking of that too. Especially regarding the reward, the user might be clueless regarding what scaler to set.

However regarding the reward system, I am currently facing issues. One issue is inherent to what we're trying to do (bandits applied to model selection), that is reward distribution shift upward with increase learning. The other one is due to the scaling of said reward, which is done online. To sum up, in some applications the model selection seems quite dependent on initial conditions such as model ordering, which is really bad.

As a quick example, let's suppose : 
- the model 1 is pulled the first 3 times with mse 9, 8 and 7 (the mse decrease with training)
- _then_ the model 2 is pulled with mse 9. 

```python
from river.stats import Mean

# Scaler for reward
sc = preprocessing.StandardScaler()

# Model 1 (pulled the first 3 times)
m1 = Mean()
for mse_m1 in [9, 8, 7]:
    r1 = dict(r=-mse_m1)
    tr1 = sc.learn_one(r1).transform_one(r1)
    m1.update(tr1["r"])
    print("MSE for M1:", mse_m1, "\tCurrent reward for M1:", round(tr1["r"], 2), "\tAverage reward for M1:", round(m1.get(), 2))

# Model 2 (pulled the 4th time)
m2 = Mean()
mse_m2 = 9
r2 = dict(r=-mse_m2)
tr2 = sc.learn_one(r2).transform_one(r2)
m2.update(tr2["r"])
print("MSE for M2:", mse_m2, "\tCurrent reward for M2:", round(tr2["r"], 2), "\tAverage reward for M2:", round(m2.get(), 2))
```
which yields:
```python
MSE for M1: 9 	Current reward for M1: 0.0 	Average reward for M1: 0.0
MSE for M1: 8 	Current reward for M1: 1.0 	Average reward for M1: 0.5
MSE for M1: 7 	Current reward for M1: 1.22 	Average reward for M1: 0.74

MSE for M2: 9 	Current reward for M2: -0.9 	Average reward for M2: -0.9
```

So the **model 2 has the same MSE (at first pull) than model 1 but has much worst reward and thus average reward**.  

I have try things like updating the reward only after a given number of iterations (so the scaler has time to adapt), or explore each arm first but without great success.

MaxHalford:
Would it help to have a "warm-up" period where each model is updated regardless of its cumulative reward?

etiennekintzler:
Hello @MaxHalford !

> Would it help to have a "warm-up" period where each model is updated regardless of its cumulative reward?

Yes I did, this what I meant when i was talking about "explore each arm first" strategy in my previous message.

After thinking and tinkering with the rewarding system, I found an alternative that seems to work well which is **to get rid of the online scaling for the reward and to use strong discounting for the first _n_ rewards**.

More specifically:
1. _fixed_ scaling system like sigmoid function for the reward instead of online scaler.
2. online scaler to standardize y. It allows sigmoid to be used for problems which have different scale for y (and thus  MSE/MAE).
3. discounting factor for the first rewards, like exponential decay or just setting reward to 0 for the first _n_ iterations. The goal is to have learning for the y scaler and for the models without skewing the average reward for the future. Sounds a bit like warm-up/explore first strategy but without messing with the average reward.

The benefits of abandoning the online scaler for the reward are also :
- simplify the choice the user has to make 
- avoid the risk that a model which has really bad reward skew the whole scaling.

However the main drawback is that some bandit model (like UCB) make hypothesis about the distribution of the reward (e.g subgaussian), which might be a bit different with what we obtained using sigmoid.

river/expert/bandit.py
```diff
@@ -0,0 +1,358 @@
...
+    'UCBRegressor',
+]
+
+
+class Bandit(base.EnsembleMixin):
+
+    def __init__(self, models: typing.List[base.Estimator], metric: metrics.Metric, explore_each_arm: int, start_after: int, seed: int = None):
+
+        if len(models) <= 1:
+            raise ValueError(f"You supply {len(models)} models. At least 2 models should be supplied.")
```
MaxHalford, line 27: You supplied

river/expert/bandit.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+    def _compute_scaled_reward(self, y_pred, y_true, c=1, scale_y=True):
+        if scale_y:
+            y_true = self._y_scaler.learn_one(dict(y=y_true)).transform_one(dict(y=y_true))["y"]
+            y_pred = self._y_scaler.transform_one(dict(y=y_pred))["y"]
+
+        metric_value = self.metric._eval(y_pred, y_true)
+        metric_value = metric_value if self.metric.bigger_is_better else (-1) * metric_value
+        reward = 1 / (1 + math.exp(- c * metric_value)) if c * metric_value > -30 else 0 # to avoid overflow
```
MaxHalford, line 150: I think there's a `sigmoid` function in `utils` that you can use.

river/expert/bandit.py
```diff
@@ -0,0 +1,358 @@
...
+                                            (reward - self._average_reward[chosen_arm])
+
+        # Specific update of the arm for certain bandit model
+        self._update_arm(chosen_arm, reward)
+
+
+    def _compute_scaled_reward(self, y_pred, y_true, c=1, scale_y=True):
+        if scale_y:
+            y_true = self._y_scaler.learn_one(dict(y=y_true)).transform_one(dict(y=y_true))["y"]
+            y_pred = self._y_scaler.transform_one(dict(y=y_pred))["y"]
```
MaxHalford, line 146: Does scaling `y` help significantly? I'm wondering if this isn't something we want to put in its own class.

MaxHalford (commented):
Looking good!

One important thing: since you've started this PR we've added more thorough code style rules. Essentially you'll want to run `pre-commit install --hook-type pre-push` on your work station. This will run `black` + `flake8` before you git push. You can fix the code style for `black` by running `black river --config .black`.

etiennekintzler:
> One important thing: since you've started this PR we've added more thorough code style rules. Essentially you'll want to run `pre-commit install --hook-type pre-push` on your work station. This will run `black` + `flake8` before you git push. You can fix the code style for `black` by running `black river --config .black`.

I had done it (as suggested in CONTRIBUTING.md) but had error when pushing from local. 

Thinking the issue was on my side, I pushed directly from Github (as 7bea5bb was just name change) and had the same error as the one I had in local, that is  : 

```bash
would reformat /home/runner/work/river/river/river/expert/bandit.py
Oh no! 💥 💔 💥
1 file would be reformatted, 311 files would be left unchanged.
would reformat /home/runner/work/river/river/river/expert/bandit.py
Oh no! 💥 💔 💥
1 file would be reformatted, 311 files would be left unchanged.

flake8...................................................................Failed
- hook id: flake8
- exit code: 1

river/utils/math.py:335:12: E741 ambiguous variable name 'l'
river/expert/bandit.py:143:5: E303 too many blank lines (2)

Error: Process completed with exit code 1.
```
does it mean that E741 and E303 are blocking and that I have to be resolve it by myself ?

Also, I don't know what to do with the `DocTestFailure` that happens because the docstring's Example should have the output of the command as well (when it's not stored in a variable). For example I can't use : 
```python
>>> for x, y in dataset:
...    bandit.learn_one(x=x, y=y)
```
since `learn_one` return `self`. Of course the result could be stored in a variable (I guess that's the reason why it's written like this in the README.md) but it looks a bit awkward to me.

river/expert/bandit.py
etiennekintzler: Yes this is what I found, especially since now there is no scaling of the reward.

The idea is to have a distribution of the reward that is invariant to the scale of y.  If you don't do this, the bandit mechanism depends on the scale of y, basically, divide y by 100 and you get only exploration, multiply by 100 and you have only exploitation.

Below is a graph of the distribution (ECDF) of the reward for a MSE in Unif(0, h) for h = [0.01, 0.1, 1, 10, 100].

image

MaxHalford:
> does it mean that E741 and E303 are blocking and that I have to be resolve it by myself ?

Yep those are `flake8` errors that you have to correct.

It seems that you also have some `black` errors, which you have correct by running `black river --config .black`. I advise installing the dev version of `black` before doing this: `pip install git+git://github.com/psf/black --upgrade`.

etiennekintzler:
> > does it mean that E741 and E303 are blocking and that I have to be resolve it by myself ?
> 
> Yep those are `flake8` errors that you have to correct.

Ok !

> It seems that you also have some `black` errors, which you have correct by running `black river --config .black`. I advise installing the dev version of `black` before doing this: `pip install git+git://github.com/psf/black --upgrade`.

Thank you :) work well after update

> > Also, I don't know what to do with the DocTestFailure that happens because the docstring's Example should have the output of the command as well (when it's not stored in a variable)
> 
> Alas you have to assign the output to a variable. Typically I would write `bandit = bandit.learn_one(x, y)`. Now that I think about it, we could consider removing the `return self` from each `fit_one` method altogether as I don't see it bringing any value.

Yep or you could just remove the related test (that will run for every `base.Regressor` I guess).

MaxHalford:
It's a bit more complicated than that because we need to update every example.

etiennekintzler:
Oh ok I see,

Should I set it to 

```python
>>> for x, y in dataset:
...    bandit = bandit.learn_one(x=x, y=y)
```
and keep `return self` in `learn_one` ... or wait for the change ?

MaxHalford:
I wouldn't wait for the changes, it'll take some time :)

etiennekintzler:
Hello @MaxHalford !

There are tests not related to this PR that are failing. 

They are located in `river/compat/test_sklearn.py`.

Also there is another test that results in internal error (in local as in the CI/CD pipeline). It happens right after `river/metrics/test_.py::test_pickling[_RollingRegressionReport]` (see in CI/CD logs), error is:

```python
warnings._OptionError: invalid module name: 'sklearn.metrics.classification'
```

MaxHalford:
Hey mate!

I fixed the tests :). It was all down to the new release of scikit-learn (0.24).

etiennekintzler:
Great ! Can you push it ? :D

etiennekintzler:
Cool, all tests passed now !

I've nothing else to add for now, so you can review and merge if it looks good to you :)

The main change since your last review is the `argmax` function defined directly in `river/bandits.py`. _Standard_ argmax will tend to favour pulling the first model since `argmax([0, 0, 0, 0])` will always return 0. _This_ argmax will pull a random arm in this very case. When there is one and only one maximum, both functions are of course the same.

Also, I don't know how you want to frame it but I think it could be good to mark this bandit class as **experimental** since this implementation doesn't really stem from theory (hence the importance of having researchers input on this issue).

river/expert/bandit.py
```diff
@@ -0,0 +1,504 @@
...
+    def best_model(self):
+        return self[self._best_model_idx]
+
+    @property
+    def percentage_pulled(self):
+        if not self.warm_up:
+            percentages = [n / sum(self._N) for n in self._N]
+        else:
+            percentages = [0] * self._n_arms
+        return percentages
```
MaxHalford, line 116: I would prefer the shorter:

```py
if not self.warm_up:
    return [n / sum(self._N) for n in self._N]
return percentages = [0] * self._n_arms
```

river/expert/bandit.py
```diff
@@ -0,0 +1,504 @@
...
+        if epsilon_decay:
+            self._starting_epsilon = epsilon
+
+    def _pull_arm(self):
+        if self._rng.random() > self.epsilon:
+            chosen_arm = argmax(self.average_reward, self._rng)
+        else:
+            chosen_arm = self._rng.choice(range(self._n_arms))
+
+        return chosen_arm
```
MaxHalford, line 213: Same here, could be shorter.

MaxHalford (commented):
Looks really good IMO! Ideally, I would like it if you could add some more comments to the internal functions. Maybe that adding a docstring to the `Bandit` would help. You could quickly describe the purpose of each function and how they work together. Not sure if I'm clear :). I just want newcomers to be able to grok how we're framing this. Then again, the code is really clear.

river/expert/bandit.py
```diff
@@ -0,0 +1,504 @@
...
+        # Explore all arms pulled less than `explore_each_arm` times
+        never_pulled_arm = [i for (i, n) in enumerate(self._N) if n < self.explore_each_arm]
+        if never_pulled_arm:
+            chosen_arm = self._rng.choice(never_pulled_arm)
+        else:
+            chosen_arm = self._pull_arm()
+
+        # Predict and learn with the chosen model
+        chosen_model = self[chosen_arm]
+        y_pred = chosen_model.predict_one(x)
```
MaxHalford, line 145: This might be `predict_proba` for a classifier right?

etiennekintzler: Yes, you're right, didn't anticipate that. Also, for classifier the whole scaling thing is less of a problem (since the target is {0, 1}).

MaxHalford:
I guess we can merge now, and take care of the classification aspect in another PR?

etiennekintzler:
> Looks really good IMO! Ideally, I would like it if you could add some more comments to the internal functions. Maybe that adding a docstring to the Bandit would help. You could quickly describe the purpose of each function and how they work together. Not sure if I'm clear :). I just want newcomers to be able to grok how we're framing this. Then again, the code is really clear.

Thank you really much ! I added some docstring

> I guess we can merge now, and take care of the classification aspect in another PR?

Yes, I think you can merge :ok_hand:   ! 

The classification aspect in `_compute_scaled_reward` can be treated when introducing classifiers. Before classifiers, I'll do a PR with Exp3 regressor which is straightforward to implement and offer good results ([see this WIP notebook](https://github.com/etiennekintzler/visualize_bandit_algorithms/blob/master/bandit_model_selection.ipynb)).

PR #404, opened 2020-11-28
Tests to check for emerging and absent input features

### LINKED ISSUE #403: Create tests to check for emerging and absent input features
Dictionaries are an efficient and flexible input representation to deal with online machine learning. They are also sparse by nature.
All of these characteristics are powerful allies of online machine learning researchers and practitioners.

However, we need to make sure that the current algorithms and upcoming ones do not break if new features emerge or are extinct (temporally or not). More than that, we might need to devise new strategies to leverage the maximum of the available information in such situations.

The first step is to create automatic tests checking whether or not the estimators, transformers, etc., are resilient to such situations.

MaxHalford:
Context is in #403 

I've added a couple of unit tests that apply to each model. This catches a lot of bugs, like `KeyError`s. I think that the best way to proceed is make individual pull requests to fix each model. We can identidy which models to fix by looking at the outputs from GitHub Actions and/or checking out this branch and running `pytest`. Then we can merge this PR. Sounds good @smastelini?

This closes #403 on merge.


smastelini:
I almost couldn't find PR #404. Daddy jokes aside, here go my comments.

> I think that the best way to proceed is make individual pull requests to fix each model.

I agree. Then we could keep pulling from master to make sure everything is passing as intended.

> We can identidy which models to fix by looking at the outputs from GitHub Actions and/or checking out this branch and running pytest. 

The outputs of Github actions work fine, at least to me. I expected to see some of the methods there failing, but others were a surprise.

river/utils/estimator_checks.py
```diff
@@ -130,6 +136,30 @@ def check_shuffle_features_no_impact(model, dataset):
...
+def check_disappearing_features(model, dataset):
+    """The model should work fine when features disappear."""
+
+    x, _ = next(iter(dataset))
+    features = list(x.keys())
+
+    for x, y in dataset:
+        random.shuffle(features)
+        model.learn_one(x, y)
+        model.predict_one({i: x[i] for i in features[:-3]})  # drop 3 features at random
```
smastelini, line 160: Wondeful!

MaxHalford: I think so too! We should get in the habit of writing these kinds of tests, it will make our code much more fool proof.

smastelini: Totally agreed!

smastelini:
#394 should fix some of the failing tests here.

smastelini:
@MaxHalford, could you please pull the latest changes brought by #394 to the trees so that I can check whether or not something is missing?

PR #405, opened 2020-11-28
Add methods to clone and reset estimators

jacobmontiel:
When dealing with estimators there are two cases to consider:

- **Reset** Sometimes it is necessary to get to the initial state of a model, in other words, the model is again empty. This is useful for example when dealing with concept drift. In some cases we want the model to "forget" the previous concept. This is the purpose of the `reset` method.
- **Clone** The `clone` method takes as input an estimator (possibly trained) and yields a new empty estimator that is empty. This is useful for example in ensembles.

### DIFF
--- a/river/base/__init__.py
+++ b/river/base/__init__.py
@@ -20,6 +20,7 @@ from .clusterer import Clusterer
 from .drift_detector import DriftDetector
 from .ensemble import EnsembleMixin
 from .estimator import Estimator
+from .estimator import clone
 from .multi_output import MultiOutputMixin
 from .regressor import Regressor
 from .regressor import MiniBatchRegressor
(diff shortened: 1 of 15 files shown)


jacobmontiel:
@smastelini @MaxHalford This is a WIP. At the moment I have achieved to make Adaptive Random Forest and Streaming Random Patches only relying on `clone` instead of `deepcopy`. The downside of this is to use a utility method (`clone`) instead of a class method as it would be with `reset`.

```python
estimator_1 = base.clone(estimator_1)
```

vs

```python
estimator_1.reset()
```

IMO the reset option is more elegant.

jacobmontiel:
Dropped in favor of the implementation in #406.

PR #406, opened 2020-11-28
Add clone method

MaxHalford:


river/compose/select.py
```diff
@@ -60,7 +60,7 @@ def __repr__(self):
     def _set_params(self, blacklist=None):
         if not blacklist:
             blacklist = self.blacklist
-        return Select(*blacklist)
+        return Discard(*blacklist)
```
MaxHalford, line 63: Caught a nice bug with the `check_clone` test :)

jacobmontiel:
This implementation of `clone` is better than the one in #405. It has been agreed that it will be an `Estimator` class method.

MaxHalford:
Just got a few kinks to figure out, it'll be merged tomorrow :)

PR #411, opened 2020-12-01
Propagate new Base class and the clone method

jacobmontiel:
This PR covers:

- Replacing `deepcopy()` with `clone()` in Streaming Random Patches and Adaptive Random Forest
- Making the `DriftDetector` class an extension of the `Base` class

Using `clone` in SRP and ARF simplifies the code. Thanks to the updates on the `_get_params` and `_set_params` in the `Base` class we can now use `**kwargs` in the ARF, decluttering the signature for this class. Now the ARF is consistent with the rest of the tree-based methods.

### DIFF
--- a/river/base/drift_detector.py
+++ b/river/base/drift_detector.py
@@ -2,15 +2,11 @@ import abc
 import numbers
 import typing
 
-from . import estimator
+from . import base
 
 
-class DriftDetector(estimator.Estimator):
+class DriftDetector(base.Base):
     """A drift detector."""
-
-    @property
-    def _supervised(self):
-        return False
 
     def __init__(self):
         super().__init__()
--- a/river/ensemble/adaptive_random_forest.py
+++ b/river/ensemble/adaptive_random_forest.py
@@ -203,10 +203,10 @@ class BaseTreeClassifier(HoeffdingTreeClassifier):
             )
 
     def new_instance(self):
-        params = self._get_params()
+        new_instance = self.clone()
         # Use existing rng to enforce a different model
-        params.update(seed=self._rng)
-        return self.__class__(**params)
+        new_instance._rng = self._rng
+        return new_instance
 
 
 class BaseTreeRegressor(HoeffdingTreeRegressor):
@@ -290,10 +290,10 @@ class BaseTreeRegressor(HoeffdingTreeRegressor):
             return new_adaptive
 
     def new_instance(self):
-        params = self._get_params()
+        new_instance = self.clone()
         # Use existing rng to enforce a different model
-        params.update(seed=self._rng)
-        return self.__class__(**params)
+        new_instance._rng = self._rng
+        return new_instance
 
 
 class AdaptiveRandomForestClassifier(BaseForest, base.Classifier):
@@ -467,7 +467,7 @@ class AdaptiveRandomForestClassifier(BaseForest, base.Classifier):
                  attr_obs_params: dict = None,
                  max_size: int = 32,
                  memory_estimate_period: int = 2000000,
-                 seed=None,
+                 seed: int = None,
                  **kwargs,
                  ):
         super().__init__(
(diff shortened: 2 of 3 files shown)


river/base/drift_detector.py
```diff
@@ -2,16 +2,12 @@
...
-class DriftDetector(estimator.Estimator):
+class DriftDetector(base.Base):
     """A drift detector."""
 
-    @property
-    def _supervised(self):
-        return False
-
     def __init__(self):
         super().__init__()
```
MaxHalford, line 12: No need for this I think.

river/ensemble/adaptive_random_forest.py
```diff
@@ -303,25 +290,10 @@ def _new_learning_node(self, initial_stats=None, parent=None):   # noqa
...
-                              max_features=self.max_features,
-                              nominal_attributes=self.nominal_attributes,
-                              attr_obs=self.attr_obs,
-                              attr_obs_params=self.attr_obs_params,
-                              min_samples_split=self.min_samples_split,
-                              max_depth=self.max_depth,    # noqa
-                              seed=self._rng)
+        params = self._get_params()
+        # Use existing rng to enforce a different model
+        params.update(seed=self._rng)
```
MaxHalford, line 295: Is this the reason why you document that `seed` can be a `np.random.RandomState`? Another possibility could simply be:

```py
new = self.clone()
new._rng = self._rng
```

river/ensemble/adaptive_random_forest.py
```diff
@@ -489,23 +455,21 @@ def __init__(self,
...
                  leaf_prediction: str = 'nba',
                  nb_threshold: int = 0,
                  nominal_attributes: list = None,
                  attr_obs: str = 'gaussian',
                  attr_obs_params: dict = None,
-                 max_depth: int = None,
-                 seed=None):
+                 max_size: int = 32,
+                 memory_estimate_period: int = 2000000,
+                 seed=None,
```
MaxHalford, line 470: Should be `seed: int = None`. It's not the end of the world not to specify it but it will help when we enforce `mypy` :)

MaxHalford (commented):
Nice PR!

river/base/drift_detector.py
jacobmontiel: I removed it as the property is defined in the `Estimator` class. I am not completely sure that it applies to drift detectors.

river/ensemble/adaptive_random_forest.py
jacobmontiel: It is important to make sure that models are different inside the ensemble, we had some issues in the past where we were (wrongly) using the same empty models which could lead to under-performance. Using `RandomState` makes things easier.

I changed the code as suggested in cb3e7f2835520439edcd5d1b3f3be9bee18cf278

river/ensemble/adaptive_random_forest.py
jacobmontiel: Done cb3e7f2835520439edcd5d1b3f3be9bee18cf278 :)

MaxHalford (approved):

smastelini:
Hi @jacobmontiel, this PR LGTM. My only concern is that some memory-based parameters appear in both the `__init__` and the documentation of the methods. They belong to `BaseHoeffdingTree`. Are they intended to appear there?

Moreover, should we tackle the failing tests of #404 in this PR?

jacobmontiel:
> My only concern is that some memory-based parameters appear in both the `__init__` and the documentation of the methods. They belong to `BaseHoeffdingTree`. Are they intended to appear there?

`max_size` and `memory_estimate_period` are documetned since they use different values to the default in `BaseHoeffdingTree`.

> Moreover, should we tackle the failing tests of #404 in this PR?

I am not sure to follow the connection with this PR.

smastelini:
> max_size and memory_estimate_period are documetned since they use different values to the default in BaseHoeffdingTree.

Got it! You are right.

> I am not sure to follow the connection with this PR.

Both ARF and SRP fail when new features appear or some of them might be missing.

smastelini (approved):

PR #417, opened 2020-12-05
Enable parallel updates in Covariance estimates

smastelini:
- Uses existing [formulas](https://doi.org/10.1145/3221269.3223036) to enable parallel updates of covariance estimates
- Add support to `frequency weights` in `stats.Cov` updates
- Create tests for the new functionality


smastelini:
Although I used `git mv` to rename the test file, it seems Github thinks I deleted the old file and created a new one.

river/stats/cov.py
```diff
@@ -39,17 +47,130 @@ def __init__(self, ddof=1):
...
+
+        # Scale the covariances
+        result.cov = scale_a * self.cov + scale_b * other.cov
+        # Apply correction factor
+        result.cov += (
+            (self.mean_x.get() - other.mean_x.get()) * (self.mean_y.get() - other.mean_y.get())
+            * ((self.mean_x.n * other.mean_x.n) / result.mean_x.n)
+        )
+        # Reapply scale
+        result.cov /= (result.mean_x.n - result.ddof)
```
MaxHalford, line 83: Can't you do something like the following?

```py
result = copy.deepcopy(self)
result += other
```

I didn't think about it much but it might save you some lines of code.

river/stats/cov.py
```diff
@@ -39,17 +47,130 @@ def __init__(self, ddof=1):
...
+        self.cov += (
+            (old_mean_x - other.mean_x.get()) * (old_mean_y - other.mean_y.get())
+            * ((old_n * other.mean_x.n) / self.mean_x.n)
+        )
+        # Reapply scale
+        self.cov /= (self.mean_x.n - self.ddof)
+
+        return self
+
+    def __sub__(self, other):
```
MaxHalford, line 115: Same comment :)

MaxHalford (commented):
LGTM! Just made a comment about code reuse.

river/stats/cov.py
smastelini: Yeap! Nice catch. It simplifies the code and reduces the amount of repeated logic :D

PR #421, opened 2020-12-15
Minor fixes to ARFReg

smastelini:
- Make the weighted vote be applied only when aggregation is `'mean'`
- Fix typo
- Update test

### DIFF
--- a/.github/ISSUE_TEMPLATE/discussion.md
+++ b/.github/ISSUE_TEMPLATE/discussion.md
@@ -1,0 +1,14 @@
+---
+name: Discussion
+about: Ask a question or start a discussion
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!--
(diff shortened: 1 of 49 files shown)


smastelini:
Hey @jacobmontiel, just one question: shouldn't the drift detectors be updated before the base models are updated with new instances?

While reading the code, this got my attention.

jacobmontiel:
> shouldn't the drift detectors be updated before the base models are updated with new instances?

Pinging @hmgomes

This also applies to the classifier version.

smastelini:
@jacobmontiel, could we close this one and address the more theoretical bit in another PR, if necessary?

jacobmontiel (approved):
Changes are correct. Thanks @smastelini

jacobmontiel:
Tests fail due to upstream issues. Closing the PR.

Issue #398, opened 2020-11-26
Support mini-batches in the naive_bayes module


MaxHalford:
I believe that it shouldn't be too difficult to upgrade the models from the `naive_bayes` by allowing them to process mini-batches (i.e. pandas DataFrames), just like what we do in the `linear_model` module and in `preprocessing.StandardScaler`. Indeed, Naive Bayes models essentially boil down to "counting". The one model that might need to be refactored a bit more is the `GaussianNB`.

Of course, scikit-learn's `naive_bayes` module already supports mini-batches. However, we can bring some value to the table, just like we did for the models in `linear_model`:

- We support pandas DataFrames rather than numpy arrays.
- We can add/drop features in between batches (scikit-learn requires the batches to have the same number of columns).
- Because we look at column names rather than positions, our models will still work if mini-batches have shuffled features.

Ideally, I would like to **not** be the person to implement this, to reduce the [bus factor](https://www.wikiwand.com/en/Bus_factor). I think it would be great if someone else could have a go at processing mini-batches.

I don't have the time to detail things right now, but I would be glad to answer questions and review pull requests.

raphaelsty:
I'll take care of it.

I'm wondering about the data entry format for Naïve Bayes in mini-batch mode. I share here my intuition while waiting for confirmation.

I think that a *Term Frequency* dataframe is the input adapted to the Naïve Bayes model in mini-batch mode. It seems to me that it is the most coherent data structure compared to the type of input we chose for the incremental version: `collections.counter`. 
```python
>>> docs = [
...     ('Chinese Beijing Chinese', 'yes'),
...     ('Chinese Chinese Shanghai', 'yes'),
...     ('Chinese Macao', 'yes'),
...     ('Tokyo Japan Chinese', 'no')
... ]
```

<img width="354" alt="image" src="https://user-images.githubusercontent.com/24591024/102290394-3fc17800-3f41-11eb-953f-a12981497f9e.png">

If we select the *Term Frequency* DataFrame then it is the role of the `feature_extraction.BagOfWords` module to convert a pandas.Series of text into a *Term Frequency* DataFrame. 🙂

```python
class BagOfWords:

    transform_many(self, X: pd.Series):
        pass
```

Finally, if we want to vectorize NaiveBayes and take full advantage of the mini-matches, we should also vectorize `feature_extraction.VectorizerMixin` which is responsible for pre-processing the text.

```python
class VectorizerMixin:

    def process_text_many(self, X: pd.Series):
          pass
[...]

PR #424, opened 2020-12-18
Support mini-batches in the naive_bayes module #398

raphaelsty:
Hello 🙂

This draft pull request is related to the issue #398.

This draft pull-request is intended to provide a first version of the update of the `NaiveBayes` module to support mini-batches. The incremental models of the `NaiveBayes` module all have a **very competitive** `learn_one` method in terms of execution time. The `predict_one` method is a less optimized method for **larger** mini-batches.

I generated benchmarks to better measure the effect of the pull-request. I start by creating a dataset (10000 rows) with 300 tokens and 100 different classes.

```python
import random

import pandas as pd 

from river import naive_bayes
from river import feature_extraction

X = pd.Series([(i + random.randint(1, 1000)) % 100 for i in range(1, 10000)])
Y = X % 100
X = X.astype(str) + 'B ' + X.astype(str) + 'C ' + X.astype(str) + 'A'
```

#### Runtime in ms:

<table><thead><tr><th></th><th colspan="3">Learn</th><th colspan="3">Predict</th></tr></thead><tbody><tr><td>Model</td><td>incremental</td><td>mini-batch</td><td>sklearn</td><td>incremental</td><td>mini-batch</td><td>sklearn</td></tr><tr><td>BernoulliNB</td><td>149</td><td>108</td><td>91.1</td><td>8600</td><td>154</td><td>89</td></tr><tr><td>ComplementNB</td><td>226</td><td>129</td><td>91.4</td><td>6410</td><td>147</td><td>85.8</td></tr><tr><td>MultinomialNB</td><td>208</td><td>108</td><td>91.2</td><td>7560</td><td>157</td><td>89.2</td></tr></tbody></table>

My first benchmark between river and sklearn wasn't right. The performance of the sklearn implementation relies on sparse matrices. Benchmarking without sparse matrices means that sklearn models are very slow. With a sparse matrix their models are very fast, excuse me sklearn. 😅

As @MaxHalford proposed it and as sklearn does, I used sparse matrices to speed up calculations and limit the memory usage of the models.

Our implementation is very similar to sklearn's with some differences that justify the differences in execution time in batch mode:
- The Naive Bayes models of river do not know the vocabulary in advance and can adapt.
- The internal data structure of the River Naive Bayes models are based on dictionaries to enjoy the best of both worlds (incremental & batch learning). 

To provide a mini-match version of `NaiveBayes` I added a `transform_many` method to the `feature_extraction.BagOfWords` module. It produce a sparse pandas dataframe.

```python
bag = feature_extraction.BagOfWords()

def transform_one(bag, X):
    for x in X:
        bag.transform_one(x)
```

```python
# incremental
%timeit transform_one(bag, X)
38.7 ms ± 730 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

```python
# mini-batches 
%timeit bag.transform_many(X)
61.2 ms ± 537 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Here is the output of `bag.transform_many(X)`:

image

List of units tests provided:

- River Naive Bayes models provide same results as sklearn models. 
- The index is kept.
- Batch learning provide same results as incremental learning.
- Batch predicting provide same results as incremental predicting

Test coverage of `naive_bayes` module: 100%

I will update the `naive_bayes.GaussianNB` model in a future PR because it requires a bit more refactoring.

Raphaël


river/base/classifier.py
```diff
@@ -71,6 +71,36 @@ def predict_one(self, x: dict) -> base.typing.ClfTarget:
...
+
+        """
+        raise NotImplementedError
+
+    def predict_many(self, X: pd.DataFrame) -> pd.Series:
+        """Predict the labels of a DataFrame `X`.
+
+        Parameters
+        ----------
+        x
```
MaxHalford, line 94: `X`

river/feature_extraction/vectorize.py
```diff
@@ -59,7 +62,9 @@ def find_ngrams(tokens: typing.List[str], n: int) -> typing.Iterator[N_GRAM]:
     return zip(*[tokens[i:] for i in range(n)])
 
 
-def find_all_ngrams(tokens: typing.List[str], ngram_range: range) -> typing.Iterator[N_GRAM]:
+def find_all_ngrams(
+    tokens: typing.List[str], ngram_range: range
+) -> typing.Iterator[N_GRAM]:
```
MaxHalford, line 67: I think you have some code-style issues. We're now using `pre-commit` to enforce consistency. See the [contribution guide](https://github.com/online-ml/river/blob/master/CONTRIBUTING.md) to find out how to set this up locally :)

river/feature_extraction/vectorize.py
```diff
@@ -156,6 +163,11 @@ def process_text(self, x):
             x = step(x)
         return x
 
+    def process_text_many(self, X: pd.Series):
+        for step in self.processing_steps:
+            X = X.apply(step)
+        return X
+
```
MaxHalford, line 170: Is this used? I see you're using `process_text` in a list comprehension underneath, which I think is better. I would only add a `process_text_many` if it brought performance benefits. The list comprehension makes it clear that this is "just" a for loop.

river/feature_extraction/vectorize.py
```diff
@@ -259,6 +271,28 @@ class BagOfWords(base.Transformer, VectorizerMixin):
...
+                    idx[x] = i
+                    i += 1
+
+        array = np.zeros((len(X_p), len(idx)), dtype=int)
+
+        for row, x in enumerate(X_p):
+            for col in x:
+                array[row][idx[col]] += 1
+
+        return pd.DataFrame(array, columns=idx.keys(), index=X.index)
```
MaxHalford, line 291: Interesting! Could you try out using [scipy's `lil_matrix`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.lil_matrix.html)? You might be able to kill two birds with one stone:

1. Possibly.
2. You don't need to do numpy -> pandas full -> pandas sparse (which could save time and memory). The problem with doing `bag.transform_many(X).astype('Sparse[int]')` is not just the time overhead: it incurs a memory spike that we would like to avoid.

river/feature_extraction/vectorize.py
```diff
@@ -259,6 +271,28 @@ class BagOfWords(base.Transformer, VectorizerMixin):
...
+
+    def learn_many(self, X):
+        pass
```
MaxHalford, line 294: Should be `return self`. In fact there should be a unit test that checks that `learn_one` and `learn_many` always return `self` (I think I did it for `learn_one`)

river/naive_bayes/bernoulli.py
```diff
@@ -92,6 +99,80 @@ class BernoulliNB(base.BaseNB):
...
+
+    >>> cp('Beijing', 'no') == (0 + 1) / (1 + 2)
+    True
+    >>> cp('Macao', 'no') == (0 + 1) / (1 + 2)
+    True
+    >>> cp('Shanghai', 'no') == (0 + 1) / (1 + 2)
+    True
+
+    >>> unseen_data = pd.Series(
+    ...    ['Taiwanese Taipei', 'Chinese Shanghai'], name = 'docs', index=['river', 'rocks'])
```
MaxHalford, line 163: I would prefer:

```py
>>> unseen_data = pd.Series(
...    ['Taiwanese Taipei', 'Chinese Shanghai'],
...    name='docs',
...    index=['river', 'rocks']
... )
```

By the way nice index ;)

river/naive_bayes/bernoulli.py
```diff
@@ -92,6 +99,80 @@ class BernoulliNB(base.BaseNB):
...
+
+    >>> model.predict_proba_many(unseen_data)
+            yes        no
+    river  0.883154  0.116846
+    rocks  0.952731  0.047269
+
+    >>> model.predict_many(unseen_data)
+    river    yes
+    rocks    yes
```
MaxHalford, line 172: <3

river/naive_bayes/bernoulli.py
```diff
@@ -136,3 +217,66 @@ def joint_log_likelihood(self, x):
             )
             for c in self.class_counts
         }
+
+    def learn_many(self, X: pd.DataFrame, y: pd.Series):
+        X = (X > self.true_threshold) * 1
+
+        agg, index = base.Groupby(keys=y).apply(np.sum, X.values)
```
MaxHalford, line 224: Could you try comparing with a for loop over the values of `y`? So something like:

```py
for y_val in y.unique():
    counts = X[y == y_val].sum()
    # do your update
``` 

I don't know, maybe your approach is better. I'm just wondering if there isn't a much faster way to do this that we aren't seeing :)

river/naive_bayes/complement.py
```diff
@@ -53,6 +57,45 @@ class ComplementNB(base.BaseNB):
...
+    >>> model.predict_proba_one('food job meat')
+    {'health': 0.779191, 'butcher': 0.220808}
+
+    >>> unseen_data = pd.Series(
+    ...    ['food job meat', 'Taiwanese Taipei'], name = 'docs', index = ['river', 'rocks'])
+
+    >>> model.predict_proba_many(unseen_data)
+        butcher    health
+    river  0.220809  0.779191
+    rocks  0.623077  0.376923
```
MaxHalford, line 92: I think there's a formatting issue.

river/naive_bayes/complement.py
```diff
@@ -53,6 +57,45 @@ class ComplementNB(base.BaseNB):
...
+    >>> model['nb'].p_class('health') == 2 / 3
+    True
+    >>> model['nb'].p_class('butcher') == 1 / 3
+    True
+
+    >>> model.predict_proba_one('food job meat')
+    {'health': 0.779191, 'butcher': 0.220808}
+
+    >>> unseen_data = pd.Series(
+    ...    ['food job meat', 'Taiwanese Taipei'], name = 'docs', index = ['river', 'rocks'])
```
MaxHalford, line 87: Same thing here about style, it's your call :)

river/naive_bayes/complement.py
```diff
@@ -12,7 +15,8 @@
 class ComplementNB(base.BaseNB):
     """Naive Bayes classifier for multinomial models.
 
-    The input vector has to contain positive values, such as counts or TF-IDF values.
+    This estimator supports learning with mini-batches. The input vector has to contain positive
+    values, such as counts or TF-IDF values.
```
MaxHalford, line 19: Maybe you want to mention that said counts and TF-IDF values can be obtained via `BagOfWords` and `TFIDF`.

MaxHalford (changes requested):
Really good PR! I didn't put in the effort to deep-dive into the details but it looks good overall. The comparison with scikit-learn is really good. Although I'm sure that we should be able to outperform them on `predict_many` as well :).

Regarding the sparse dataframe, let's re-assess once you've tried `lil_matrix`. If it's not better then we can give up on making `BagOfWords` return a sparse dataframe. However we should make sure that our models accept sparse dataframes. A unit test that uses scikit-learn's `CountVectorizer`would be ideal if so.

Again, good job!

river/naive_bayes/bernoulli.py
raphaelsty: It seems that the GroupBy class is more efficient on my benchmark, up to 3 times faster than a for loop. Agree with you, we need to benchmark things. 🙂

```python
def learn_many(self, X: pd.DataFrame, y: pd.Series):
    for y_val in y.unique():
        agg = X[y == y_val].sum()
        self.class_counts.update({y_val: y[y == y_val].sum()})
        self.feature_counts.update((agg.T).to_dict())
        self.class_totals.update({y_val: agg.sum()})

    return self

X_t = feature_extraction.BagOfWords().transform_many(X)

model = naive_bayes.MultinomialNB()

%timeit learn_many(self=model, X=X_t, y=Y)
269 ms ± 46 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
 model = naive_bayes.MultinomialNB()

%timeit model.learn_many(X_t, Y)
73.8 ms ± 15.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

river/naive_bayes/base.py
```diff
@@ -25,6 +35,34 @@ def predict_proba_one(self, x):
...
+
     @property
     def _multiclass(self):
         return True
+
+
+def from_dict(data: dict) -> pd.DataFrame:
+    """Convert a dict into a pandas dataframe."""
+    dict_data, index = list(data.values()), list(data.keys())
+    return pd.DataFrame(data=dict_data, index=index, dtype="float32")
```
MaxHalford, line 52: Can you do this with `pd.DataFrame.from_dict`?

river/naive_bayes/bernoulli.py
```diff
@@ -9,19 +13,23 @@
 class BernoulliNB(base.BaseNB):
     """Bernoulli Naive Bayes.
 
+    Bernouilli Naive Bayes model learns from occurrences between features such as word counting
```
MaxHalford, line 16:
```suggestion
    Bernoulli Naive Bayes model learns from occurrences between features such as word counting
```

river/naive_bayes/bernoulli.py
```diff
@@ -9,19 +13,23 @@
...
+    and discrete classes. The input vector must contain positive values, such as
+    counts or TF-IDF values.
+
     Parameters
     ----------
     alpha
-        Additive (Laplace/Lidstone) smoothing parameter (use 0 for no smoothing).
+    Additive (Laplace/Lidstone) smoothing parameter (use 0 for no smoothing).
```
MaxHalford, line 23:
```suggestion
        Additive (Laplace/Lidstone) smoothing parameter (use 0 for no smoothing).
```

river/naive_bayes/bernoulli.py
```diff
@@ -9,19 +13,23 @@
...
     true_threshold
-        Threshold for binarizing (mapping to booleans) features.
+    Threshold for binarizing (mapping to booleans) features.
```
MaxHalford, line 25:
```suggestion
        Threshold for binarizing (mapping to booleans) features.
```

river/naive_bayes/bernoulli.py
```diff
@@ -9,19 +13,23 @@
...
 
     Attributes
     ----------
     class_counts : collections.Counter
-        Number of times each class has been seen.
+    Number of times each class has been seen.
     feature_counts : collections.defaultdict
-        Total frequencies per feature and class.
+    Total frequencies per feature and class.
```
MaxHalford, line 32: Add spaces to the descriptions :)

river/naive_bayes/bernoulli.py
```diff
@@ -104,6 +226,16 @@ def __init__(self, alpha=1.0, true_threshold=0.0):
...
 
     def learn_one(self, x, y):
+        """Updates the model with a single observation.
+
+        Args:
+            x: Dictionary of term frequencies.
+            y: Target class.
+
+        Returns:
+            self
```
MaxHalford, line 236:
```suggestion
        Parameters
        -----------
        x
            Dictionary of term frequencies.
        y
            Target class.

        Returns
        --------
        self
```

river/naive_bayes/bernoulli.py
```diff
@@ -119,7 +251,21 @@ def p_feature_given_class(self, f: str, c: str) -> float:
...
+        )
+
     def joint_log_likelihood(self, x):
+        """Computes the joint log likelihood of input features.
+
+        Args:
+            x: Dictionary of term frequencies.
+
+        Returns:
+            Mapping between classes and joint log likelihood.
```
MaxHalford, line 266: Same here, you're not using the correct format.

river/naive_bayes/multinomial.py
```diff
@@ -11,7 +15,10 @@
 class MultinomialNB(base.BaseNB):
     """Naive Bayes classifier for multinomial models.
 
-    The input vector has to contain positive values, such as counts or TF-IDF values.
+    Multinomial Naive Bayes model learns from occurrences between features such as word counting
```
MaxHalford, line 18:
```suggestion
    Multinomial Naive Bayes model learns from occurrences between features such as word counts
```

river/naive_bayes/bernoulli.py
```diff
@@ -285,9 +292,16 @@ def joint_log_likelihood(self, x):
...
 
-        Args:
-            X: Term-frequency or TF-IDF pandas dataframe.
-            y: Target classes.
+        Parameters
+        ----------
+            X
+                Term-frequency or TF-IDF pandas dataframe.
+            y
+                Target classes.
```
MaxHalford, line 300:
```suggestion
        X
            Term-frequency or TF-IDF pandas dataframe.
        y
            Target classes.
```

PR #426, opened 2020-12-23
Add AdaptiveStandardScaler

MaxHalford:
This uses an exponentially weighted moving average and variance to scale the data. Initially I was going to add an `alpha` parameter to `StandardScaler` and use some `if`s  but it looked really ugly, so I created a separate class, which I think is fine.


smastelini:
LGTM! The failing tests are related to the new release of `sklearn`. The fix is already in the `master`.

Nice work, @MaxHalford! I am excited to see this new transformer in action.

smastelini (approved):

PR #447, opened 2021-01-20
Add Clustream method while adding sample_weight for learn_one

hoanganhngo610:


MaxHalford:
Hey there, just a quick challenge: is CluStream really an online method? The prediction phase feels really heavy with the KMeans and all. I think it's great to have a reference implementation, and the code quality is high, but maybe this should be part of the river-extras package. Happy to be challenged on this :)

hoanganhngo610:
@MaxHalford Actually Clustream is one of the first online methods that dated back to 2003. The prediction phase (offline phase as they mentioned) is heavy as you will have to apply KMeans for every iteration. However, as it is the foundation of online clustering, I think it would be nice to have it as a benchmark for future implementations of more advanced/modern ones.

MaxHalford:
Ok no problem. I would recommend putting a flag/disclaimer in the docstring to highlight that it's a costly method :)

smastelini:
Another aspect worth noting: since k-Means is used, `sklearn` becomes a dependency.

`sklearn` is not required to install `river`. I would rather keep things this way. Could we do something like what is done in the `compat` module and check if `sklearn` is installed on demand? Thus, we do not force users to install `sklearn` if they are not going to use functionalities relying on batch learning algorithms. What do you think, @MaxHalford?

MaxHalford:
Yes I completely agree, we don't want to introduce scikit-learn as a dependency.

jacobmontiel:
Disclaimer: This is a WIP based on a prototype.

One of the goals is to try to leverage the existing `kmeans` implementation in `river` to avoid dependencies.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+        assert closest_a != closest_b
+        self.kernels[closest_a].add(self.kernels[closest_b])
+        self.kernels[closest_b] = ClustreamKernel(x=x, sample_weight=sample_weight,
+                                                  timestamp=self.time_stamp, T=self.T, M=self.M)
+
+    def get_micro_clustering_result(self):
+
+        if not self.initialized:
+            return {}
+        res = {i: None for i in range(len(self.kernels))}
```
jacobmontiel, line 222: Use `dict` comprehension: https://www.python.org/dev/peps/pep-0274/#id7


```res = {i: ClustreamKernel(cluster=self.kernels[i], T=self.T, M=self.M) for i in range(len(self.kernels))} ```

File: river/cluster/clustream.py [hunk 2]
```diff
@@ -0,0 +1,393 @@
...
+
+    def get_clustering_result(self):
+
+        if not self.initialized:
+            return {}
+        micro_cluster_centers = {}
+        for i in len(self.get_micro_clustering_result()):
+            micro_cluster_centers[i] = self.get_micro_clustering_result()[i].center
+
+        # turn micro_cluster_centers to numpy array to use kmeans
```
jacobmontiel, line 235: Keep for the moment to evaluate if the prototype works.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+    def learn_predict_one(self, x, sample_weight=None):
+        """
+        Compute cluster centers and predict cluster index for each sample.
+
+        Convenience method; equivalent to calling learn_one(x) followed by predict_one(x).
+
+        Parameters
+        ----------
+        x
+            A dictionary with n items
```
jacobmontiel, line 255: Delete line, not applicable.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+            Instance attributes
+
+        sample_weight
+            Float or dictionary-like
```
jacobmontiel, line 259: Remove line.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+            Instance weights. If not provided, uniform weights are assumed
+
+        Returns
+        -------
+        y
+            Integer
+            Cluster label
+        """
+        if sample_weight is None:
```
jacobmontiel, line 268: Replace with `self.learn_one(x, sample_weight)`

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+            sample_weight = 1.0
+        if sample_weight != 0.0:
+            self._train_weight_seen_by_model += sample_weight
+            self._learn_one(x, sample_weight)
+
+        micro_cluster_centers = {}
```
jacobmontiel, line 274: Might be possible to replace with `self.predict_one()`

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+            distance between closest kernel to the given instance
+        """
+
+        min_distance = sys.float_info.max
+        closest_kernel = None
+        closest_kernel_index = -1
+        for i, micro_cluster in micro_clusters.items():
+            distance = 0
+            for j in range(len(x)):
+                distance += (micro_cluster.center[j] - x[j])**2
```
jacobmontiel, line 327: `a * a` is faster than `a**2`

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+            distance = math.sqrt(distance)
+            if distance < min_distance:
+                min_distance = distance
+                closest_kernel = micro_cluster
+                closest_kernel_index = i
+        return closest_kernel_index, min_distance
+
+    @staticmethod
+    def implements_micro_clustering():
```
jacobmontiel, line 336: Not sure if this is necessary.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+        return True
+
+    @property
+    def name(self):
```
jacobmontiel, line 340: Not applicable. Remove.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+    def _distance(point_a, point_b):
+        distance = 0.0
+        for i in range(len(point_a)):
+            d = point_a[i] - point_b[i]
+            distance += d * d
+        return math.sqrt(distance)
+
+    def predict_one(self, x):
+        """
+        predict cluster index for each sample.
```
jacobmontiel, line 353: The description must be in the first line after """

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+
+        Returns
+        -------
+        label
+            Integer
+            Cluster label
+
+        """
+
+        micro_cluster_centers = {}
```
jacobmontiel, line 371: Replace with dict comprehension.

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+from abc import ABCMeta, abstractmethod
+import math
+
+EPSILON = 0.00005
+MIN_VARIANCE = 1e-50
+
+# from file clustream_kernel.py in scikit-multiflow/clustering
+
+
+class ClustreamKernel(base.Clusterer, metaclass=ABCMeta):
```
jacobmontiel, line 11: This auxiliary class is not an extension of `base.Clusterer`

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+            super().__init__(cluster=cluster)
+            self.N = cluster.N
+            self.LS = cluster.LS.copy()
+            self.SS = cluster.SS.copy()
+            self.LST = cluster.LST
+            self.SST = cluster.SST
+
+    @property
+    def center(self):
+        assert not self.is_empty()
```
jacobmontiel, line 41: Avoid using `assert`

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+        res = {n: 0.0 for n in range(len(self.LS))}
```
jacobmontiel, line 42: Use dict comprehension

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+        variance = self._variance_vector
+        sum_of_deviation = 0
+        for i in range(len(variance)):
+            d = math.sqrt(variance[i])
+            sum_of_deviation += d
+        return sum_of_deviation / len(variance)
+
+    @property
+    def _variance_vector(self):
+        res = {n: 0.0 for n in range(len(self.LS))}
```
jacobmontiel, line 67: Change to `res = {}`

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+            ls_div_n_squared = ls_div_n * ls_div_n
+            ss_div_n = ss / self.weight
+            res[i] = ss_div_n - ls_div_n_squared
+
+            if res[i] <= 0.0:
+                if res[i] > - EPSILON:
+                    res[i] = MIN_VARIANCE
+        return res
+
+    # implemented from cluster_feature.py
```
jacobmontiel, line 81: Comment is not relevant.

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def add(self, cluster):
+        assert len(cluster.LS) == len(self.LS)
+        self.N += cluster.N
+        self.LST += cluster.LST
+        self.SST += cluster.SST
+        self.add_vectors(self.LS, cluster.LS)
+        self.add_vectors(self.SS, cluster.SS)
+
+    # implemented from cluster_feature.py
```
jacobmontiel, line 146: Delete comment

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+    @staticmethod
+    def add_vectors(v1, v2):
+        assert v1 is not None
+        assert v2 is not None
+        assert len(v1) == len(v2)
+        for i in range(len(v1)):
+            v1[i] += v2[i]
+
+    @abstractmethod
+    def get_CF(self):
```
jacobmontiel, line 156: Not necessary, remove.

jacobmontiel (changes requested):
Thanks for the first draft @hoanganhngo610 

Please see my comments and address the changes.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+      When deciding to add a new data point to a micro-cluster, the maximum boundary
+      is defined as a factor of the kernel_radius_factor of the RMS deviation of the
+      data points in the micro-cluster from the centroid
+
+    number_of_clusters: int (Default : 5)
+        the clusters returned by the Kmeans algorithm using the summaries statistics
+
+    Attributes
+    ----------
+    centers : dict
```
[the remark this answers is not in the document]
jacobmontiel: Confirmed.

river/cluster/clustream.py
```diff
@@ -0,0 +1,393 @@
...
+    1
+
+    >>> clustream_test.predict_one({0: 4, 1: 4})
+    0
+
+    >>> clustream_test.centers #make sure that clustream can return centers under the correct format
+
+    References
+    ----------
+    .. [1] A. Kumar , A. Singh, and R. Singh. An efficient hybrid-clustream algorithm
```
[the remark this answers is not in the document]
MaxHalford: The convention I've been using is to use the "Harvard" format. I have a Google Scholar plugin in my browser that gives me the citation when I type in a paper.

river/cluster/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Unsupervised clustering."""
 from .k_means import KMeans
+from .clustream import Clustream
```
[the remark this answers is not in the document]
MaxHalford: This can be automated this via `isort`, I'll put it in place so that the git push hooks do it :)

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def _mu_time(self):
+        return self.LST / self.N
+
+    @property
+    def _sigma_time(self):
+        return math.sqrt(self.SST/self.N - (self.LST/self.N) * (self.LST/self.N))
+
+    def _quantile(self, z):
+        assert (z >= 0 and z <= 1)
```
hoanganhngo610: Replaced with `assert 0 <= z <= 1`.

hoanganhngo610 (commented):
Dear Jacob,

Thank you so much for your comment! I have pushed the new version fixing all the comments that we mentioned.

hoanganhngo610:
@MaxHalford @smastelini In the latest version, I have used the `KMeans` that are implemented in-house in `River` to replace for the original `KMeans` in `sklearn`. This allows us not to introduce `sklearn` as a dependency, while making the code much cleaner by not having to transform `dict` to `np.array` and vice versa!

MaxHalford:
You're the man! I'll take a look soonish :)

hoanganhngo610:
@MaxHalford Thank you so much! I really appreciate it! :D 
@jacobmontiel I have successfully passed all the tests and it is now ready for review! :)

river/base/clusterer.py
```diff
@@ -11,14 +11,18 @@ def _supervised(self):
...
+    def learn_one(self, x: dict, sample_weight: int = None) -> "Clusterer":
         """Update the model with a set of features `x`.
 
         Parameters
         ----------
         x
             A dictionary of features.
 
+        sample_weight
+            Integer, optional (default=None)
```
MaxHalford, line 23: Don't specify the type and the default value. This should all be done in the signature.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+    ----------
+    seed
+       If int, seed is the seed used by the random number generator;
+       If RandomState instance, seed is the random number generator;
+       If None, the random number generator is the RandomState instance used
+       by `np.random`.
+       It is used in the `KMeans` algorithm for the offline clustering, which is the native `KMneans` implemented in `River`.
+
+    time_window
+      The rang of the window
```
MaxHalford, line 27: You need to use 4 spaces to indent each parameter description, as you did for the `seed` parameter.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+    ...     [1, 4],
+    ...     [1, 0],
+    ...     [4, 2],
+    ...     [4, 4],
+    ...     [4, 0]
+    ... ]
+
+    >>> clustream_test = cluster.Clustream(time_window = 1, max_kernels = 3, number_of_clusters = 2, seed = 0, halflife = 0.4)
+
+    >>> for i, (x, _) in enumerate(stream.iter_array(X)):
```
MaxHalford, line 72: No point in using `.` right?

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+    >>> X = [
+    ...     [1, 2],
+    ...     [1, 4],
+    ...     [1, 0],
+    ...     [4, 2],
+    ...     [4, 4],
+    ...     [4, 0]
+    ... ]
+
+    >>> clustream_test = cluster.Clustream(time_window = 1, max_kernels = 3, number_of_clusters = 2, seed = 0, halflife = 0.4)
```
MaxHalford, line 70: Don't use spaces around `=`. You can put the declaration on multiple lines to improve readability.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+
+        Tasks performed before training:
+        * Verify instance weight. if not provided, uniform weights (1.0) are assumed.
+        * Update weight seen by model.
+
+        Training tasks:
+        * determinate closest kernel
+        * Check whether instance fits into closest Kernel:
+            1- if data fits, put into kernel
+            2- if data does not fit, we need to free some space to insert a new kernel
```
MaxHalford, line 126: I'm not sure this formatting renders well in the docs. You can run `make livedoc` locally to see what this would look like.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+        )  # add this to retrieve centers later for the evaluation of models through
+        self.kernel_radius_factor = kernel_radius_factor
+        self.max_kernels = max_kernels
+        self.number_of_clusters = number_of_clusters
+        self._train_weight_seen_by_model = 0.0
+        self.seed = seed
+        self.kwargs = kwargs
+
+    def learn_one(self, x, sample_weight=None):
+        """Incrementally trains the model.
```
MaxHalford, line 116: If I were you I would put all this description in the class docstring and leave this docstring empty. I think it's better to describe the model in one place. If you don't specify the `learn_one` docstring, then the docstring from the `Clusterer` class will be inherited, which should be good enough.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+            if len(self.buffer) < self.buffer_size:
+                self.buffer[len(self.buffer)] = ClustreamKernel(
+                    x=x,
+                    sample_weight=sample_weight,
+                    timestamp=self.time_stamp,
+                    T=self.kernel_radius_factor,
+                    M=self.max_kernels,
+                )
+                return self
+            else:
```
MaxHalford, line 163: No need for the `else` here because there's a `return` in the `if`. In general I think it's a good idea to try to un-nest logic as much as possible.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+                        timestamp=self.time_stamp,
+                        T=self.kernel_radius_factor,
+                        M=self.max_kernels,
+                    )
+            self.buffer.clear()
+            self.initialized = True
+
+            return self
+
+        """determine closest kernel"""
```
MaxHalford, line 177: Prefer `# determine closest kernel` to `"""determine closest kernel"""`

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+        closest_kernel = None
+        min_distance = sys.float_info.max
```
MaxHalford, line 179: I would prefer `math.inf`

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+            return {}
+        res = {
+            i: ClustreamKernel(
+                cluster=self.kernels[i], T=self.kernel_radius_factor, M=self.max_kernels
+            )
+            for i in range(len(self.kernels))
+        }
+        return res
+
+    def learn_predict_one(self, x, sample_weight=None):
```
MaxHalford, line 255: I wouldn't add this function. Fundamentally it doesn't do anything more than calling `learn_one` followed by `predict_one`.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+
+        min_distance = sys.float_info.max
+        closest_kernel_index = -1
+        for i, micro_cluster in micro_clusters.items():
+            distance = 0
+            for j in range(len(x)):
+                distance += (micro_cluster.center[j] - x[j]) * (
+                    micro_cluster.center[j] - x[j]
+                )
+            distance = math.sqrt(distance)
```
MaxHalford, line 313: Is there any reason why this is not using the `_distance` static method?

MaxHalford (changes requested):
Here's a first pass from me. If possible I would put a lot of the logic in separate functions that are easier to unit test instead of having a big `learn_one` method. Also it seems to me that a lot of the distance and nearest neighbor search is redundant with what there is in `KMeans`. It would be preferable to have some utility methods to work with dicts to evaluate distances and search for nearest neighbors instead of duplicating logic. We should some love into the `neighbors` module. I'm not knocking what you did, it's really cool and I'm sure a lot of people will find it useful.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
+import math
+import sys
+
+from river import base, cluster
+from river.cluster.clustream_kernel import ClustreamKernel
+
+
+class Clustream(base.Clusterer):
```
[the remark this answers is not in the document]
jacobmontiel: +1

river/cluster/clustream.py
jacobmontiel: Since `learn_one` returns `self`, the user can do `clusterer.learn_one(x).predict_one(x)`. I guess it is kind of elegant to have a single call, but have no strong opinion. 

However, I see two issues:
- Should the order be inverted? `predict_learn(x)`
- If we keep this method, we should make it a method of the `Clusterer` class.

Comments @MaxHalford @smastelini ?

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def predict_one(self, x):
+        """Predict cluster index for each sample.
+
+        Convenience method; equivalent to calling partial_fit(x) followed by predict(x).
+
+        Parameters
+        ----------
+        x
+            A dictionary with n items
```
jacobmontiel, line 335: This line is not necessary.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+            Instance attributes
+
+        Returns
+        -------
+        label
+            Integer
```
jacobmontiel, line 341: Remove types as they are colleced from the signature.

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,151 @@
...
+EPSILON = 0.00005
+MIN_VARIANCE = 1e-50
+
+
+class ClustreamKernel(metaclass=ABCMeta):
+    def __init__(
+        self, x=None, sample_weight=None, cluster=None, timestamp=None, T=None, M=None
+    ):
+
+        self.T = T
```
jacobmontiel, line 13: Please change to an explicit variable name to ease the maintenance of the code.

river/cluster/clustream_kernel.py
```diff
@@ -0,0 +1,151 @@
...
+        zprod *= z2  # z ^ 11
+        res += (34807 * zprod) / 364953600
+
+        zprod *= z2  # z ^ 13
+        res += (20036983 * zprod) / 797058662400
+
+        return res
+
+    def add(self, cluster):
+        assert len(cluster.linear_sum) == len(self.linear_sum)
```
jacobmontiel, line 138: `assert` must be used with caution. If really needed then as an error message so the user knows what caused the error. We must assume that not all users are experts.

jacobmontiel (changes requested):
This is looking good @hoanganhngo610 !

Thanks for the latest updates. Please look at Max's and my comments/requests.

river/cluster/clustream.py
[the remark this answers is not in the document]
smastelini: +1

river/cluster/clustream_kernel.py
MaxHalford: Yep I missed but we definitely don't want asserts in the code. Raise an exception instead.

river/cluster/clustream.py
smastelini: I would keep something of this nature only if the speedups were clear. As @MaxHalford said, the function chains the call to two other already available functions.

Regarding the ordering, I am not sure hahaha
In general, clustering is a whole different monster when compared to supervised learning. The task is descriptive rather than predictive. So, I am really not sure we need to "predict" before learning.

river/base/clusterer.py
hoanganhngo610: Resolved.

river/cluster/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Unsupervised clustering."""
+from .clustream import Clustream
 from .k_means import KMeans
 
-__all__ = ["KMeans"]
+__all__ = ["Clustream", "KMeans"]
```
hoanganhngo610, line 5: modified `Clustream into `CluStream`.

river/cluster/clustream.py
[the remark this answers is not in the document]
hoanganhngo610: Resolved.

river/cluster/clustream.py
hoanganhngo610: Resolved.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+       by `np.random`.
+       It is used in the `KMeans` algorithm for the offline clustering, which is the native `KMneans` implemented in `River`.
+
+    time_window
+      The rang of the window
+      if the current time is T and the time window is h, we should only consider
+      about the data that arrived within the period (T-h,T)
+
+    max_kernels
+      The Maximum number of micro kernels to use
```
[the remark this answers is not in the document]
hoanganhngo610: Resolved :D

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+
+    kernel_radius_factor
+      Multiplier for the kernel radius
```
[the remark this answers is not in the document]
hoanganhngo610: Resolved.

river/cluster/clustream.py
```diff
@@ -0,0 +1,363 @@
...
+      When deciding to add a new data point to a micro-cluster, the maximum boundary
+      is defined as a factor of the kernel_radius_factor of the RMS deviation of the
+      data points in the micro-cluster from the centroid
+
+    number_of_clusters
+        the clusters returned by the Kmeans algorithm using the summaries statistics
+
+    kwargs
+        Other parameters passed to the incremental kmeans at `river.cluster.KMeans`
```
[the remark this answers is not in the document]
hoanganhngo610: Resolved.

river/cluster/clustream.py
hoanganhngo610: Resolved.

river/cluster/clustream.py
hoanganhngo610: I do not really understand what you mean by using `.`...

river/cluster/clustream.py
hoanganhngo610: As I believe, each `learn_one` function of each algorithm is fairly different from each other, so it's kind of necessary to specify what this function does for each algorithm (although the purpose is the same, which is "incrementally trains the model").

river/cluster/clustream.py
jacobmontiel: Thanks for the comments to both. We agree that the functionality can be achieved by calling `learn_one` and `predict_one`. On the other hand, the order if we were to keep this function is not really clear. So let's remove it for the moment and only if a use-case is identified we would re-evaluate its implementation.

river/cluster/clustream.py
hoanganhngo610: Resolved.

river/cluster/clustream.py
hoanganhngo610: Thanks all for all of your comments! After the discussion with Jacob yesterday, we both agreed that we would remove it for the moment and the user can simply do `clusterer.learn_one().predict_one()` as an equivalence to this function.

river/cluster/clustream.py
hoanganhngo610: As I have defined it as a static method, so if we want to use the `_distance` function, we have to pass `self` into the function, which makes the `_get_closest_kernel` not a static method anymore.

river/cluster/clustream.py
hoanganhngo610: Removed.

river/cluster/clustream.py
hoanganhngo610: Removed.

river/cluster/clustream_kernel.py
hoanganhngo610: I have changed `assert` into an `if ` condition so as if a new cluster passes by while its dimension does not match with the dimension of `self`, it will do nothing instead of raising an error.

river/cluster/clustream_kernel.py
hoanganhngo610: I have changed `T` to `kernel_radius_factor` and `M` to `max_kernels`.

river/cluster/clustream.py
hoanganhngo610: Resolved. Also separated it to a new function called `initialise_learn_one`.

river/cluster/clustream.py
hoanganhngo610: I have changed `1-`, `2-` into`*` and this works well locally in Pycharm!

hoanganhngo610:
@MaxHalford @smastelini @jacobmontiel Thank you so much for all of your comments! I have fixed my code based on your recommendations!

jacobmontiel (approved):
Thanks for the last changes @hoanganhngo610 

The implementation is ready. There were some minor code-standards details that I changed. No functional changes.

river/cluster/clustream.py
MaxHalford: Haha I meant the `i` variable

PR #448, opened 2021-01-20
Add ML-KNN

kavorite:
hi all

I was working on a multilabel classification project, and I thought to myself, it'd be very nice if `online-ml/river` had an incremental version of `skmultilearn.adapt.MLkNN`. I missed it very much in `river` because I wanted a version that would learn incrementally, and I wanted to know how it worked anyway, so I made my own implementation which is more familiar to `river`'s interface. While I realize that `river.neighbors` already included a multi-output classification algorithm that operates on this premise, it didn't perform as well as the `skmultilearn` implementation for recalling multiple labels for my application; this one gets comparable results on the yeast dataset to the classifier chain already included in `river.multioutput`

I hope this is to your satisfaction, I wasn't really sure how to integrate things with the test suite so I included an evaluation script in my implementation of the module

I want to get back to work on my project, though, so I'll wait for feedback on how to proceed with a more idiomatic set of unit tests that play a bit nicer with what's being used upstream

### DIFF
--- a/river/ensemble/adaptive_random_forest.py
+++ b/river/ensemble/adaptive_random_forest.py
@@ -264,7 +264,7 @@ class BaseTreeRegressor(HoeffdingTreeRegressor):
         leaf_model: base.Regressor = None,
         model_selector_decay: float = 0.95,
         nominal_attributes: list = None,
-        attr_obs: str = "gaussian",
+        attr_obs: str = "e-bst",
         attr_obs_params: dict = None,
         min_samples_split: int = 5,
         seed=None,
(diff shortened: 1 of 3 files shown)


smastelini:
Hi @kavorite, thanks for contributing. We do not have many multi-label classification and multi-target regression algorithms yet. So, a multi-label classification k-NN implementation ought to be a nice addition. I am not familiar with this specific implementation, so I'll learn a bit about it.

In the meanwhile, could you please fix the formatting issues that are causing the CI tests to fail? Hence, we can start the reviewing process.

PR #457, opened 2021-01-28
Use the log-sum-exp trick in the NB leaf models

jacobmontiel:
The log-sum-trick protects the Naive Bayes calculation against underflow.

For more information: https://stats.stackexchange.com/a/253319


smastelini (approved):
Thanks for the PR, @jacobmontiel! It will make river's tree models robust against underflow issues while calculating NB predictions.

PR #459, opened 2021-01-30
Implementation of the discriminative drift detector

ogozuacik:
Hello everyone,
I recently graduated from my masters study, and this is the implementation of the method which my thesis is about. (Unsupervised concept drift detection with a discriminative classifier, CIKM 2019)

I'm open to suggestions, and will update the PR according to your comments.


MaxHalford:
Hey there, thanks for the PR. I'm not an expert on drift but one thing apparent to me is that you're using lists instead dicts as features. We really don't want to encourage and prefer for implementation to use dicts natively. For instance, your method breaks if the features order changes between updates.

ogozuacik:
OK, I'll update the PR. (use dict instead)

MaxHalford:
Also, I forgot to tell you, but we don't want to add scikit-learn as a dependency... The whole point of River is to move away from scikit-learn and batch libraries.

Another that might be interesting to do/think about. We basically have a bunch of unit tests that run for every classifier and regressor. They're in `utils/check_estimator.py`. It would be great to have some tests for drift detectors as well. For instance, a good test would be to check that the drift detectors produce the same results if the features are shuffled. Tell me if this isn't clear :)

ogozuacik:
Thanks you for your comments. I'll make the changes. I'll look into the `check_estimators.py`. Something like check_drift_detectors can be nice. I'll ping you if I have issues related in that part.

ogozuacik:
Hi again. I updated the code, and now it uses dicts as an input instead of lists. I removed the dependency for AUC calculation and StratifiedKFold, but I still use LogisticRegression from sklearn. I'm trying to use the built-in LogisticRegression in river, but it does not output the results that I desire (that are somewhat similar to the paper). That's why, I still use that from sklearn.

D3 is an unsupervised drift drift detection methodology. It takes an instance from a stream as an input ( dimensions=(N,1)). In simplest words, it stores latest samples from the data stream. Lets say parameters are 100, 0.3, 0.7. It stores the latest 130 samples. Labels old data (first 100 samples in the window) "1", and new data "0". It splits the data into training and test in stratified form. Then trains a batch classifier (a simple one) to classify whether old and new data can be classified. If old and new are separable, the classification will be successful and yield high AUC. In that case a drift is signaled. My point is, in some level I need to use a batch classifier. 

Regarding tests for drift detectors. Other drift detectors in river are supervised methods and only take a single value as an input. (Like 1 or 0). It is the classification result of the stream classifier for most drift detectors (True or False). Only unsupervised approaches use whole data stream instance. I guess they will not be affected by the features getting shuffled (as they mostly accept a single value)

I made D3 to be resistant to the shuffled features. I'm sorting the keys.

Btw, I guess there is an issue with the KSWIN test. It seems to be ambivalent. In some of the runs, it signals a drift in indices that are not specified in the expected indices.

river/drift/d3.py
```diff
@@ -0,0 +1,189 @@
...
+        np.random.seed(seed)
+        super().reset()
+
+    def _is_sliding_window_full(self):
+        """Checks if the sliding window is full."""
+        if self.sliding_window_index < self.full_window_size:
+            return False
+        return True
+
+    def _discriminative_drift_detection(self, old_data, new_data):
```
jacobmontiel, line 96: As mentioned, we are trying to minimize dependencies. 

the method uses a batch model in a stream setting. This is not wrong. However, the implementation has some limitations:

- The implementation is restricted to `sklearn` models. The `fit` and `predict_proba` are not standard in `river`, so trying to use a `river` model is not possible.
- There is overhead from converting the data to lists/numpy-arrays.

From the documentation:
> It is advised to use a simple model as the goal of this classifier
> is to determine if the old data and the new data are separable,
> not to classify them.

Then there is no reason to not use the incremental implementation of [`LogisticRegression`](https://riverml.xyz/latest/api/linear-model/LogisticRegression)  in `river`

This is useful because:
- The implementation is compatible with the rest of the codebase.
- No need to convert the data.
- `np.concatenate` is slow. A faster alternative is a fixed length [`deque`](https://docs.python.org/3/library/collections.html#collections.deque).
- The testing (`predict_one`) and metric `update` can be performed inside the same loop.
- Numpy methods can be replaced by standard methods.

river/drift/d3.py
```diff
@@ -0,0 +1,189 @@
...
+class D3(DriftDetector):
+    r"""Drift Detection Method
+
+    D3 (Discriminative Drift Detector) is an unsupervised drift detection
+    method which uses a discriminative classifier that can be used with any
+    online algorithm without a built-in drift detector. It holds a fixed
+    size sliding window of the latest data having two sets: the old and the
+    new. A simple classifier is trained to distinguish these sets. It
+    detects a drift with respect to classifier performance (AUC).
+
```
jacobmontiel, line 17: Consider mentioning that this method looks for drift in multi-dimensional data. To differentiate it from the rest of the methods that work on uni-dimensional data.

jacobmontiel (changes requested):
Thanks for the last changes @ogozuacik 

Please look at my comments and let me know if you need more information.

river/drift/d3.py
ogozuacik: Thanks for your comments and suggestions. I'll update the PR till the end of this month. I'm a bit busy right now due to work.

jacobmontiel: That is alright. Thanks for looking into this.

ogozuacik:
I've made bunch of changes according to your comments. In my 2cb1787, the tests were all passing, but after I changed some stuff in the d3 doc, the tests failed on 9e99486. Any ideas on solving possible circular import?

GCBallesteros:
Hi @ogozuacik ,

I've been playing around with the new algorithm for a work project and I'm getting good results. I got the best results when going into the code and tweaking `sub_window_size`. May I suggest that `sub_window_size` is exposed as a hyperparameter? Either as a fraction of the window size, i.e. 
```python
if window_size_divisor <= 1:
    raise SomeException

# some other code...

self.sub_window_size = int(self.window_size / window_size_divisor) # window_size_divisos is the new hyperparam default is 2
```
or just the absolute value which should be checked to be smaller than self.window_size
```python
if sub_window_size < window_size:
    raise SomeException
if sub_window_size is None:
    sub_window_size = int(window_size / 2)

# some other code...
self.sub_window_size = sub_window_size
```

MaxHalford:
Closing this because of lack activity. I'm also not very satisfied with the fact that an intermediary batch model is necessary. It isn't really adher to our philosophy sadly :/. Feel free to reopen if necessary!

ogozuacik:
OK Max. Yes, it is a bit different than the logic you have in river. I'll keep the forked version for anyone interested. Thanks for your help and comments.

@GCBallesteros I'm happy to hear that the results are good. In the original implementation in the paper, sub window size was a parameter as well. I tried to simplify it while implementing in here. Feel free to change the code.

PR #468, opened 2021-02-04
Update the implementation of DDM and Page-HinckleyDM

michaelchiucw:
Update the implementation of DDM and Page-HinckleyDM to follow the description in their paper respectively.

### DIFF
--- a/river/drift/ddm.py
+++ b/river/drift/ddm.py
@@ -1,4 +1,5 @@
-import numpy as np
+import numbers
+import math
 
 from river.base import DriftDetector
 
@@ -30,7 +31,7 @@ class DDM(DriftDetector):
     * $s_i$: The standard deviation at instant $i$.
 
     The conditions for entering the warning zone and detecting change are
-    as follows:
+    as follows [see implementation note bellow]:
 
     * if $p_i + s_i \geq p_{min} + 2 * s_{min}$ -> Warning zone
 
@@ -67,14 +68,22 @@ class DDM(DriftDetector):
     >>> data_stream = np.random.randint(2, size=2000)
     >>> # Change the data distribution from index 999 to 1500, simulating an
     >>> # increase in error rate (1 indicates error)
-    >>> data_stream[999:1500] = 1
+    >>> data_stream[1000:1200] = [np.random.binomial(1, .8) for _ in range(200)]
 
     >>> # Update drift detector and verify if change is detected
     >>> for i, val in enumerate(data_stream):
     ...     in_drift, in_warning = ddm.update(val)
     ...     if in_drift:
     ...         print(f"Change detected at index {i}, input value: {val}")
-    Change detected at index 1077, input value: 1
+    Change detected at index 1157, input value: 1
+
+    Notes
+    -----
+    In this implementation, the conditions to signal drift and warning are
+    $p_i + s_i > thershold$ instead of $p_i + s_i \geq thershold$. This is to
+    avoid a corner case when a classifier is consistently wrong (`value=1`)
+    that results in DDM indicating a drift every `min_num_instances`. This
+    modification is consistent with the implementation in MOA.
 
     References
     ----------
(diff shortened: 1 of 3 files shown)


river/drift/ddm.py
```diff
@@ -133,7 +133,7 @@ def update(self, value):
         self._in_warning_zone = False
         self.delay = 0
 
-        if self.sample_count < self.min_num_instances:
+        if self.sample_count <= self.min_num_instances:
```
jacobmontiel, line 136: The current implementation is correct. DDM skips the drift detection until it has "seen" at least `min_num_instances`. This change is not necessary.

river/drift/page_hinkley.py
```diff
@@ -86,12 +88,18 @@ def update(self, value) -> tuple:
 
         self.sample_count += 1
 
+        if self.sum <= self.Mint:
```
jacobmontiel, line 91: This is not necessary. We are already accounting for this in line 87. Based on the original paper:

image

jacobmontiel (changes requested):
Thanks for the PR @michaelchiucw 

Nice catch on the test condition in DDM.

On the other hand, other changes are not necessary. Please check my comments.

river/drift/page_hinkley.py
michaelchiucw: oh okay!! Thanks.

I seem to understand why they're equivalent now.
In eq.6, it defines `S'_0 = 0` and enforces the calculation on the positive side. So, the min(S_i) will always be 0 and we don't need to find the `self.Mint` anymore. Because min(S_i) is 0, so eq.5 (s_n - min(S_i) >= h) is equivalent to eq.7.

Do I missing or misunderstanding anything here? Many thanks!

river/drift/page_hinkley.py
```diff
@@ -86,12 +88,18 @@ def update(self, value) -> tuple:
...
+
+        pht = self.sum - self.Mint
+
         self._in_concept_change = False
+        self._in_warning_zone = False
 
         if self.sample_count < self.min_instances:
             return False, False
 
-        if self.sum > self.threshold:
```
michaelchiucw, line 94: Also, here should be `if self.sum >= self.threshold:` according to the original paper. Am I right?

river/drift/ddm.py
jacobmontiel: Sorry for the delay. You are right. Although I expect the impact on the performance to be minimal, it is indeed different to the paper. Thanks for pointing this out.

river/drift/page_hinkley.py
jacobmontiel: That is correct :-)

river/drift/page_hinkley.py
jacobmontiel: Correct.

jacobmontiel (changes requested):
Thanks for the explanation. Once the discussed changes are in place we can close the PR.

river/drift/page_hinkley.py
michaelchiucw: I seem not able to the change the code from here. Should I close this pull request and make another one? thanks.

jacobmontiel: That is strange. Are you getting an error? Here is some useful information: 

https://github.com/online-ml/river/blob/master/CONTRIBUTING.md

You shouldn't need to create a new PR.

michaelchiucw (commented):
I have just updated the Page-HinckleyDM.java according to the previous discussion.

river/drift/page_hinkley.py
michaelchiucw: I have managed to find a way to edit it. Thanks!

jacobmontiel:
I finally had time to look into why tests were failing. The test was triggering a corner case where a classifier is consistently wrong (`value=1`). Using the condition from the paper (`>=`) results in false drift signals every `num_min_instances`. I rolled back the condition and added a note for DDM regarding this change and the corner case. I also took the chance to update the test cases in the docstrings and the internal ones to generate more realistic data rather than the corner case mentioned.

I am closing the PR, sorry for the delay.

jacobmontiel (approved):

PR #471, opened 2021-02-13
Benchmarks refactoring

MaxHalford:
I added a `Track` class to the `evaluate` module. I revamped the `benchmarks` directory and added a `logistic_regression` directory to show how this works. I used `rich` to display some nice progress bars. I also added some logic to write down the benchmark results to the README of the `logistic_regression` directory.

Screenshot 2021-02-13 at 21 56 24

All in all I think that this setup provides a high-level yet flexible way to benchmark a model. Having tracks means that we can define a lot of tedious stuff that developers/researchers won't have to think about. Basically they just need to call `track.run(model)` and they will receive a generator of results that they can monitor. 

Benchmarking multiple models boils down to the following function:

```py
def run():

    results = []
    tracks = load_binary_clf_tracks()
    n_checkpoints = 10

    with Progress() as progress:
        bar = progress.add_task("Models", total=len(MODELS))

        for model_name, model in MODELS.items():
            model_bar = progress.add_task(f"[green]{model_name}", total=len(tracks))

            for track in tracks:
                track_bar = progress.add_task(f"[cyan]{track.name}", total=n_checkpoints)

                for step in track.run(model, n_checkpoints=n_checkpoints):
                    step['Model'] = model_name
                    step['Track'] = track.name
                    results.append(step)
                    progress.advance(track_bar)
                progress.advance(model_bar)
            progress.advance(bar)

    return pd.DataFrame(results).set_index(['Model', 'Track', 'Step']).reset_index()
```

This is completely customisable of course. I don't think that this logic should be part of the `Track` class because it wouldn't allow as much freedom. 

Let me know what you think! If you agree on the design then I/we can add more tracks and more benchmarks.


benchmarks/logistic_regression/README.md
```diff
@@ -0,0 +1,99 @@
...
+## Final results
+
+| Model         | Track    |   Accuracy |       F1 | Time                   | Memory   |
+|:--------------|:---------|-----------:|---------:|:-----------------------|:---------|
+| PyTorch       | Bananas  |   0.54566  | 0.202297 | 0 days 00:00:02.354718 | 16.59 KB |
+| PyTorch       | Phishing |   0.814667 | 0.762068 | 0 days 00:00:00.587769 | 16.59 KB |
+| River         | Bananas  |   0.543019 | 0.195349 | 0 days 00:00:00.508606 | 4.12 KB  |
+| River         | Phishing |   0.888    | 0.872263 | 0 days 00:00:00.155899 | 5.48 KB  |
+| Vowpal Wabbit | Bananas  |   0.547075 | 0.160224 | 0 days 00:00:00.556017 | 388 B    |
+| Vowpal Wabbit | Phishing |   0.8044   | 0.741133 | 0 days 00:00:00.160629 | 388 B    |
```
smastelini, line 12: Just a random thought: does the memory measurements here are correct? I think this might be related to Cython. What do you think? As more and more portions of our codebase are implemented in Cython, we might think of a generic and effective way to measure such structures' memory usage.

benchmarks/logistic_regression/run.py
```diff
@@ -0,0 +1,179 @@
...
+        # We only know how many features a dataset contains at runtime
+        if self.network is None:
+            self.network = self.network_func(n_features=len(x))
+            self.optimizer = self.optimizer_func(self.network.parameters())
+
+        x = torch.FloatTensor(list(x.values()))
+        p = self.network(x).item()
+        return {True: p, False: 1.0 - p}
+
+class VW2CremeBase:
```
smastelini, line 69: River :)

river/base/base.py
```diff
@@ -222,7 +222,7 @@ def find(params):
         return find(self._get_params())
 
     @property
-    def _memory_usage_raw(self) -> int:
+    def _raw_memory_usage(self) -> int:
```
smastelini, line 225: Is our only chance relying on `__sizeof__` when using Cython?

river/evaluate/tracks.py
```diff
@@ -0,0 +1,67 @@
...
+
+__all__ = ["load_binary_clf_tracks"]
+
+
+class Track:
+    """A track evaluate a model's performance.
+
+    The following metrics are recorded:
+
+    - FLOPS: floating point operations per second.
```
smastelini, line 14: That's going to be a gem! I can't wait to incorporate that in my PhD's experiments :D

smastelini (commented):
Great work, @MaxHalford. I really like the simplicity of benchmarking different models using tracks. 

Besides that, although jupyter notebooks are very useful for prototyping, I'm an old school guy, so I really enjoy having something more structured XD.

Is the idea to add more tracks to expose at that README file? For instance, multiclass learners and regression models?

river/evaluate/tracks.py
MaxHalford: Yes I agree! There's [this](https://github.com/firedrakeproject/PyPAPI) Python package but I haven't managed to make it work on MacOS.

MaxHalford:
> Is the idea to add more tracks to expose at that README file? For instance, multiclass learners and regression models?

The idea is to have:

- tracks, which are datasets + metrics + evaluation protocal
- benchmarks, which are thematic. For instance, there is one logistic regression where different logistic regression implementation are compared on several binary classification tracks.

benchmarks/logistic_regression/README.md
MaxHalford: The measurements are correct for River. They're incorrect for Vowpal Wabbit but that's down to the way their Python package works.

river/base/base.py
MaxHalford: Yes I think so.

smastelini:
What do you say @MaxHalford? Should we remove FLOPS from the documentation for the time being and merge the Tracks in their current state? Or wait till the maintainers of the package that do the measurements fix the problem you are facing with OSX?

MaxHalford:
Let's merge :)

PR #476, opened 2021-02-18
[FEATURE] Quantization Observer (Hoeffding Tree Regressors)

smastelini:
This PR introduces the Quantization Observer (QO) for regression tasks, presented [in](https://www.sciencedirect.com/science/article/abs/pii/S0167865521000520):

```
Mastelini, S.M. and de Leon Ferreira, A.C.P., 2021. Using dynamical quantization to
perform split attempts in online tree regressors. Pattern Recognition Letters.
```
This algorithm is an alternative to the current binary search tree solution (E-BST). QO's main focus is to reduce the memory footprint and running time of incremental regression trees, whereas minimally impacting the predictive performance.

In addition, I also add a truncated version of E-BST (TE-BST) that first reduces the numeric precision of the incoming data before adding points to the binary search tree. By doing so, fewer points ought to be stored, which may result in some nice speedups.

An in-depth discussion and benchmark of QO are available in the manuscript (which is also available at [ArXiv](https://arxiv.org/abs/2012.00083)).

---

I ran some basic benchmarks to evaluate the impact of QO in the trees and ARFReg. In the single-target case, I generated 100K instances of the Fried generator and kept all hyperparameters at their default values for trees and attribute observers alike. In the multi-target case, a multivariate version of the Fried dataset was used (the same one employed in many incremental multi-target regression papers). The tests were performed in a machine running a linux-based OS, with an 8th generation i7-8550U processor, and 16GB of RAM. Here it goes!

- **HTR**

ao_benchmarks_htr

- **HATR**

ao_benchmarks_hatr

- **ARFReg**

Firstly, using the default values (the memory usage of QO is higher than the other attribute observers because both E-BST and TE-BST have a mechanism to prune themselves, in case split attempts fail):

ao_benchmarks_arfreg

No good :/ But one can either increase the grace period

ao_benchmarks_arfreg_gp200

ao_benchmarks_arfreg_gp500

or tune a little bit the radius parameter of QO (as a `StandardScaler` is applied in the pipeline, the features' std is around `1`, so I can use a fixed quantization radius rather than a dynamical one :D)

ao_benchmarks_arfreg_qo_0 75

- **iSOUPTree**

ao_benchmarks_isoup_tree


river/ensemble/adaptive_random_forest.py
```diff
@@ -328,7 +333,7 @@ def _new_learning_node(self, initial_stats=None, parent=None):  # noqa
                 initial_stats,
                 depth,
                 self.attr_obs,
-                self.attr_obs_params,
+                attr_obs_params,
```
MaxHalford, line 336: Just an opinion: I prefer to directly pass the attribute observer as a parameter. I don't really like passing a `str` that selects the attribute observer and then passing parameters for said observer on the side. If you look at `LogisticRegression`, the `optimizer` parameter is a `optim.Optimizer`, it's not a `str`. I think it's important to have this discussion, even if we disagree. It's not the end of the world if different parts of the code base take different approaches, but we should at least discuss it.

A nice benefit of directly passing a class instead of a string is that you don't have to document the parameters of the attribute observer in each tree. This is particularly relevant in a "big" code base like ours. The user just knows that they should look at the attribute observer module and take a look at the parameters that are there.

I'd be really interested to know what you think about this. Please tell me if I'm not clear! :)

river/tree/hoeffding_tree_regressor.py
```diff
@@ -406,6 +449,24 @@ def _attempt_to_split(self, node: LearningNode, parent: SplitNode, parent_idx: i
...
+                        # Update the var estimates using the estimators carried by the QO instances
+                        if feat_id in self._feat_var:
+                            self._feat_var[feat_id] += ao.x_var
+                        else:
+                            self._feat_var[feat_id] = ao.x_var
+
+                        self._qo_radii[feat_id] = {
+                            # Update radii estimates
+                            "radius": math.sqrt(self._feat_var[feat_id].get())
+                            / self._qo_std_div
```
MaxHalford, line 467: Is it worth catching a `ZeroDivisionError`?

MaxHalford (commented):
Not much to say on my side: your PR is really good! I'm impressed by the figures you attached. It's super cool that you're implementing your research in River.

I haven't made the effort to deep dive into the logic because, well, I don't have time, plus you're the expert :)

I've added a meta-comment about the way in which we pass structured parameters to our classes. Would love to discuss it on Discord (or live!).

river/ensemble/adaptive_random_forest.py
smastelini: I wish I had thought about that when we first made the attribute observer selection dynamic. It totally makes sense and facilitates the usage of the trees. By using strings and dictionaries we add a lot of clutter to the documentation, and even so, what each attribute observer exactly does is not clear. I would happily change that. What do you think, @jacobmontiel?

But here is where I need some help to plan:

- At this moment, the inner structures of the trees are "private": `_submodules`. How to proceed? Should we make everything public, like splits, nodes, and so on? Only the attribute observers? I am worried about suddenly adding too much information to the user process (documentation-wise).
- Secondly, and maybe more critical, E-BST and TE-BST would easily work under this new approach. It would be just a matter of using a clone of sorts or deep copies to populate the tree leaves. QO, on the other hand, has different parameters passed in runtime. I am not sure how the tree could modify the radius passed to QO in an online fashion. Here' some ideas:
    * Use `set_params`?
    * *(I don't like this one very much)* Make the radius static: we could require the user to apply a normalization step (StandardScaler or MinMaxScaler) when using QO. I know, I know, the user already does that because model tree leaves work better when all features are scaled. But, if the mean predictor is used instead, it would be nice to have the freedom to avoid an unnecessary preprocessing step and infer the adequate quantization radius on the go.

What do you think guys? @online-ml/devs, any feedback is welcomed!

MaxHalford: > At this moment, the inner structures of the trees are "private": _submodules. How to proceed? Should we make everything public, like splits, nodes, and so on? Only the attribute observers? I am worried about suddenly adding too much information to the user process (documentation-wise).

I understand. I would only make public the stuff that the user will specify. I'm not 100% aware of all the parameters you're exposing, but having a `split` module where you put all the attribute observers (I call them splitters) seems like a good idea.

> Secondly, and maybe more critical, E-BST and TE-BST would easily work under this new approach. It would be just a matter of using a clone of sorts or deep copies to populate the tree leaves. QO, on the other hand, has different parameters passed in runtime. I am not sure how the tree could modify the radius passed to QO in an online fashion. Here' some ideas: 

I guess your ideas make sense. You could inherit from `base.Base` and that will simplify your life.

smastelini: Inheriting from `base.Base` might do the trick. Should I address such a change in our codebase in this PR? I could also proceed with these changes in another PR to limit the scope. There we can discuss naming convention and documentation. 

What do you think @MaxHalford and @jacobmontiel?

MaxHalford: I would say do what feels more comfortable for you :)

smastelini: Cool! Let's wait for Jacob's feedback :D

jacobmontiel: Sorry for the delay. I also prefer passing an object rather than a string. In the end, this is an advanced option and makes sense to give more freedom/responsibility to the user.

We are also using this approach is in the `ARF` and `SRP` where the user can pass a drift detector ;-)

> You could inherit from `base.Base` and that will simplify your life.

+1

> I would only make public the stuff that the user will specify.

+1 We can always expose other things down the road as long as it is necessary.

> Should I address such a change in our codebase in this PR? I could also proceed with these changes in another PR to limit the scope.

I recommend using a different PR because

1. The change may impact other methods outside of the current scope of this PR. Historically, we have seen this happening on similar refactorings.
2. There are non-trivial issues to address like the passing of parameters at runtime for the QO.

Let me know if this is useful.

river/tree/_attribute_observer/__init__.py
```diff
@@ -20,4 +26,6 @@
     "NumericAttributeClassObserverGaussian",
     "NumericAttributeClassObserverHistogram",
     "NumericAttributeRegressionObserver",
+    "NumericAttributeRegressionQuantizerObserver",
```
jacobmontiel, line 29: Our nemesis, long class names.

What about `NumericAttributeRegQuantizerObserver`. 

However, we might need something even shorter since we will expose the obersvers to the user @_@

river/tree/_attribute_observer/nominal_attribute_class_observer.py
```diff
@@ -19,6 +19,10 @@ def __init__(self):
         self._missing_weight_observed = 0.0
         self._att_val_dist_per_class = {}
 
+    @property
+    def is_numeric(self):
```
jacobmontiel, line 23: Nice :-)

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def probability_of_attribute_value_given_class(self, x, y):
+        raise NotImplementedError
+
+    def best_evaluated_split_suggestion(
+        self, criterion, pre_split_dist, att_idx, binary_only=True
+    ):
+        candidate = AttributeSplitSuggestion(None, [{}], -math.inf)
+
+        p_x = None
```
jacobmontiel, line 54: The meaning of `p_x` is not clear.

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
```diff
@@ -0,0 +1,163 @@
...
+        num_att_binary_test = NumericAttributeBinaryTest(att_idx, split_point, True)
+        candidate = AttributeSplitSuggestion(
+            num_att_binary_test, post_split_dists, merit
+        )
+
+        return candidate
+
+
+class Slot:
+    def __init__(
```
jacobmontiel, line 92: Docstring missing. Since it is an internal class just a general description should be enough.

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
```diff
@@ -0,0 +1,163 @@
...
+            self._update_estimator = self._update_estimator_univariate
+
+    def _update_estimator_univariate(self, target, sample_weight):
+        self.y_stats.update(target, sample_weight)
+
+    def _update_estimator_multivariate(self, target, sample_weight):
+        for t in target:
+            self.y_stats[t].update(target[t], sample_weight)
+
+    def __iadd__(self, o):
```
jacobmontiel, line 124: What does `o` mean? If observer then what about `obs` or `observer`?

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
```diff
@@ -0,0 +1,163 @@
...
+        self.y_stats += o.y_stats
+
+        return self
+
+    def update(self, x, y, sample_weight):
+        self.x_stats.update(x, sample_weight)
+        self._update_estimator(y, sample_weight)
+
+
+class FeatureQuantizer:
```
jacobmontiel, line 135: Docstring missing. Since it is an internal class just a general description should be enough.

river/tree/hoeffding_tree_regressor.py
```diff
@@ -72,13 +82,25 @@ class HoeffdingTreeRegressor(BaseHoeffdingTree, base.Regressor):
     Hoeffding trees rely on Attribute Observer (AO) algorithms to monitor input features
     and perform splits. Nominal features can be easily dealt with, since the partitions
     are well-defined. Numerical features, however, require more sophisticated solutions.
-    Currently, only one AO is supported in `river` for regression trees:
+    Currently, three AO algorithms are supported in `river` for regression trees:
```
jacobmontiel, line 85: The `notes` section seems now a good candidate for a page in the documentation. It is a nice chance to see how to extend documentation and direct the user from the implementation to this extra information. This can be done in the new PR for AOs.

jacobmontiel (commented):
Awesome contribution @smastelini 

I left a couple of minor comments/questions.

river/tree/_attribute_observer/__init__.py
smastelini: I like the terminology "attribute observer" because, well, it sounds pretty cool :) 

So, I prefer it to splitter. But I don't have strong feelings. We might go with either:

- splitter 
- AO (for Attribute Observer, as I did in my paper)

We also need to define/revise our naming pattern in this upcoming PR. Maybe something like:

- QOSplitter
- EBSTSplitter

or

- AORegQO
- AORegEBST 

Any thoughts, @jacobmontiel and @MaxHalford ?

river/tree/hoeffding_tree_regressor.py
smastelini: I agree. Thus, each AO/Splitter would have its own documentation and we would also provide a nice and high level comparison of them. 

I know it has been a long time now, but I still want to do a walkthrough in the tree module. Soon, we should have Stochastic Gradient Trees too, so our comparisons will be broader.

river/tree/_attribute_observer/__init__.py
MaxHalford: I don't like AO because it's not an obvious acronym. I like splitter :)

smastelini: I see your point. I'm slowly starting to like such a naming scheme:

> - QOSplitter
> - EBSTSplitter

Time heals everything, hahaha

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
smastelini: Nothing that fancy :D

I think pycharm did the trick. As far as I am concerned, `o` stands for "other". Should I change it?

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
smastelini: I added a simple description in commit [fec06b8](https://github.com/online-ml/river/pull/476/commits/fec06b85a8b38f880bc9286c2ef570edc549717f).

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
smastelini: Description added in commit [fec06b8](https://github.com/online-ml/river/pull/476/commits/fec06b85a8b38f880bc9286c2ef570edc549717f).

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
smastelini: I renamed it to `prev_x` and added a comment stating that this variable carries the previous value of `x`. Both changes were implemented in commit [fec06b8](https://github.com/online-ml/river/pull/476/commits/fec06b85a8b38f880bc9286c2ef570edc549717f).

river/tree/_attribute_observer/nominal_attribute_class_observer.py
smastelini: We might consider dropping `Class` and `Regression/Reg` from the AO/Splitter names and using a property instead.

river/tree/hoeffding_tree_regressor.py
smastelini: Good catch! I ensured that the denominator value is always greater than zero in commit [6fc5dc5](https://github.com/online-ml/river/pull/476/commits/6fc5dc54f897fe383e62ca78a5cdf33f153891d9).

smastelini: Anyway, this solution is temporary. In the next PR, by passing the splitters directly (yes, I am getting used to this name :-D), we can simplify this portion.

river/tree/_attribute_observer/numeric_attribute_regression_quantizer_observer.py
jacobmontiel: My personal preference is to avoid variable names with a single character as it can become somewhat obscure. If there is a meaningful word why not use it. Being said that, I am ok with using `o` in this case.

river/tree/_attribute_observer/__init__.py
jacobmontiel: My two cents. Both terms are valid.

- **Attribute Observers** is the name traditionally used in stream learning literature. 
- **Splitters** is the counterpart used in batch learning (found in `scikit-learn` trees).

What about a compromise:

- We can use "Attribute Observers" in the documentation: "Attribute Observers (splitters) are..."
- We use "splitters" in the implementation as they are shorter, e.g. `QOSplitter`.
- We can have a module `attribute_observers` that contains `XXXXSplitter` classes.
- This way we enforce the fact that they represent the same thing.

jacobmontiel (approved):
Thanks for the changes @smastelini 

The PR is ready to merge.

river/tree/_attribute_observer/__init__.py
smastelini: I am ok with only keeping Attribute Observer in the documentation, to make them familiar to the stream gardeners, like me. Maybe, calling the module `splitter` will make it familiar to those used to `scikit-learn` and also shorten the name.

PR #483, opened 2021-02-23
Add "External metrics for clustering" jupyter notebook

hoanganhngo610:
This notebook is an investigation on the external metrics of `sklearn` and `river` for clustering problems. Besides, the notebook also includes a small comparision of the efficiency between different algorithms of the two packages: `Kmeans` from `sklearn`, incremental `KMeans` and `CluStream` from `river`. 

A summary of the results can be drawn as follows:
- The external metrics can not handle the permutation of cluster labels, thus a custom script is written to obtain such permutations.
- All algorithms have the same accuracy of 100% on the toy dataset, and the two `KMeans` of two packages have comparable elapsed time. However, `CluStream`, as discussed, is very expensive since it runs KMeans every time a new instance is passed.


MaxHalford:
Closing this for tidyness, feel free to reopen if work continues!

PR #484, opened 2021-02-24
Implementation of internal metrics for clustering under the submodule `metrics.cluster`

hoanganhngo610:
To create internal metrics for incremental clustering algorithms, a new submodule is created inside the `metrics` module, i.e `metrics.cluster`. Util present, there have been **21** metrics implemented for this submodule. 

First, the 4 most basic/widely used metrics that is mentioned in [Chapter 9, MOA Book](https://book.moa.cms.waikato.ac.nz/chapter_9.html/), as follows:
- **Cohesion** (average of distance to assigned centroids) 
- **SSQ** (The sum of squared distances from data points to their assigned centroids. Closely related to cohesion. The smaller the better.) The SSQ will be split into **SSW** (Sum-of-Squares Within Clusters) and **SSB** (Sum-of-Squares Between Clusters)
- **Separation** (Average distance from a point to the points assigned to other clusters. The larger the better.)
- **Silhouette coefficient** (Roughly speaking, the ratio between cohesion and the average distances from the points to their second-closest centroid. It rewards clusterings where points are very close to their assigned centroids and far from any other centroids, that is clusterings with good cohesion and good separation.)

The next two basic metrics that are also of importance is **MSSTD** and **RMSSTD**, i.e (Root) Mean Squared Standard Deviation. These metrics consider the variance of all attributes, which measure only the compactness of found clusters.

Next, one metrics that is also considered in regression problems, which measures only the separation between clusters is the **R-Squared**. To calculate this, the center of all points must be calculated, which takes advantage from the `stats.Mean` function implemented in River previously.

Finally, there are more complicated metrics that considers both the compactness and separation, while also taking into account various other parameters (distances between clusters, variance within clusters, variance of the whole dataset, etc.). These include:
- **Davies-Bouldin index (DB)**
- **I index (I)**
- **Xie-Beni index (XB)**
- **SD validity index** (which lies among the recent ones)
- **Generalized Dunn's index** (the distance that we are focusing on is Euclidean distance, so the Dunn's indices that we are implementing is **GD43** and **GD53**
- **I Index**
- **Partition Separation (PS)**
- **R-Squared for clustering (R2)**
- **Calinski-Harabasz (CH)** index, **Hartigan** index and **WB** index (both of these are based on either SSW or SSB, in general)

The literature also considers another metric, which is **S_Dbw**, however, since these two metrics require us to keep information of all points, I chose not to implement these.

Last but not least, we have the **CMM**, which is the Cluster Mapping Measure. However, the implementation of this index is an order of complexity higher than those that has been implemented here. As a result, it would be implemented in a separate PR.

To facilitate this, the `inspect` file of `utils` module is also updated to check whether a model is a Clusterer.


MaxHalford:
Hey, thanks for the work.

To be honest I'm not very comfortable with this pull request:

- The fact that you have to pass the clusterer to the `update` method is really clunky.
- Each time you pass the clusterer, the centers are different. Is this really a good idea? Isn't it just fine to evaluate a clusterer's performance with [scikit-learn's silhouette score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.silhouette_score.html) et al.? In other words I'm not sure it makes sense to calculate distances to centers in an online manner because the centers are moving, which biases the performance metric.
- You have a lot of commits in this pull request that have nothing to do with clustering metrics, is this intended?
- The docstrings do not conform to the convention's we've been using elsewhere in the code. I would suggest building the docs with `make livedoc` and seeing what your docstrings produce.
- `InternalMetric` is a really bad name, especially when it's exposed to users in the public API.
- If we include these metrics, we really want to unit test them and compare them to scikit-learn's implementations. I'm not very comfortable integrating metrics that offer different results than scikit-learn. All other metrics in River are benchmarked in such a way.

You've done a lot of work. In the future I would suggest first opening a discussion on GitHub or Discord so that we can scope our work correctly. In this case I don't feel comfortable with having this in River. But don't take it badly: I can see you're putting in a lot of effort.

jacobmontiel:
Just for clarification, this is a draft PR. All the comments/questions are valid and are on the table to discuss.

hoanganhngo610:
@MaxHalford First of all, thank you so much for your comments! Some of the points that you are raising is also of my concern, so I really hope that my points below can partially answer some of them.
This draft PR is the result of the discussion between me and Jacob last week. We want to have a prototype that can validate whether the metrics are able to be implemented, and which problems would it raise. 
About the concerns that you have raised:
- I agree that the fact that I pass the whole clusterer to the `update` method is a really clunky. We can consider replacing it by passing only the `centers`, but I want to keep the whole clusterer in case we want any informantion for more complicated metrics.
- "Each time you pass the clusterer, the centers are different." This is one of the problems that can not be solved when we incrementally calculate the metrics. If we want to calculate the metrics perfectly based on the current information that the clusterer holds, we have to save all the information about the points that passed, their labels and the clusters' centers. When the number of points reach 100k or higher, the problem will then become very expensive. This is the reason why we prefer not to use the sklearn's Silhouette coefficient, since it requires us to pass all the labels and their centers, which we do not aim to. Instead, the user can do that on their own, if they prefer to.
- Moreover, the definition between the traditional Silhouette score for offline clustering is somewhat different from the definition of Silhouette for online clustering in [Chapter 9](https://book.moa.cms.waikato.ac.nz/chapter_9.html/) of the MOA book. While the traditional Silhouette score calculates the pairwise distances between the points, the Silhouette score for online clustering calculates distance between the point to nearest and second-nearest centroid.
- The only file that I made changes, apart from the file that contains the metrics and their scripts, is the `utils/inspect.py` file, where I add the `isclusterer` function to use in `works_with`. The previous commits appears because I cloned this branch from my original `master` branch, so it contains all the commits from the beginning.
- When I run `make livedoc` locally on my machine, the docs work just fine. All of the definitions of the metrics are copied directly from the MOA book, so I think they would appear just fine.
- I also feel `InternalMetric` is a bad name. I hesitated to change it to`IntrinsicMetric`, however, since it is a little bit hard to understand, I decided to keep it the simplest way possible in my first implementation.

I feel like only describing these by words are hard to understand, so I decided to first create a draft so that we can easily come to the decision on how to build a framework for these. I totally agree with you that we have to unit test it and compare with what is currently available at scikit-learn, so once we agree on how to implement it, I will try to benchmark it immediately.
I hope that my answer wouldn't disappoint you so much! Once again, thank you so much for putting a lot of attention and for your faithful comments. I really appreciate it!

smastelini:
Just thinking out loud here. 

If the cost to calculate the metrics grows linearly with the number of observations, wouldn't be the case to use a rolling window and consider only the last examples?

river/metrics/__init__.py
```diff
@@ -26,6 +26,7 @@
     RegressionMetric,
     WrapperMetric,
 )
+from .base_internal_clustering import InternalClusteringMetrics
```
smastelini, line 29: We might come up with a shortened name for that. Besides, the name should be in singular form. Some options discussed in the meeting @hoanganhngo610, @jacobmontiel, and I had:

- IntClustMetric
- InternalClusMetric
- IClusMetric
- InterClusMetric (this one might be misleading)
- ... other ideas?

What do you think, @MaxHalford?

river/metrics/base_internal_clustering.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Mother class of all internal clustering metrics.
+    """
+
+    # Define the format specification used for string representation.
+    _fmt = ",.6f"  # Use commas to separate big numbers and show 6 decimals
+
+    @abc.abstractmethod
+    def update(
+        self, centers, point, y_pred, sample_weight
```
smastelini, line 20: As I mentioned in the meeting yesterday, I do prefer using `.update(self, x, labels, sample_weight, centers)` instead. I believe that by using `labels` rather than `y_pred` we explicitly show that clustering is a different monster other than classification/regression. Here we are not dealing with predictive tasks, but descriptive ones.

smastelini: I also prefer `x` instead of `point` for consistency.

smastelini (changes requested):
@hoanganhngo610, @jacobmontiel, and I had a nice talk about the future of clustering evaluation in `river`. Some highlights:

- External metrics (the ones that rely on an "expected" value, such as F1, rand score, etc.) are a challenge in online scenarios. Label permutations will totally impact the obtained result, as demonstrated in #483. Fixing that is not trivial and might be way too costly to do in an online fashion with the tools/information we have at the moment. We need to study how label permutations are tackled in batch-learning solutions and try to come up with a nice solution to our case.
- Regarding internal validation metrics (the most realistic ones, that do not rely on supervision), we concluded that the bare minimum set of information to calculate the most popular online clustering metrics to date is:
    - `x`: the input point per se
    - `label`: the cluster id assigned to `x` by the clustering model
    - `sample_weight` (Do all clusterers work with that?)
    - `centers`: the current centers of the clusters

This minimal set of parameters applies even to density-based clustering, at least if we consider the most popular solutions so far. Hence, that would be API access to the internal clustering validation metrics in `river`. I also mentioned the possibility of using properties in the cluster models instead, to provide such validation metrics. But that would result in a lot of redundancy and impeding the metrics of being applied standalone.

If in the future someone comes up with a different model that does not rely on the concept of prototypes, i.e., centroids, medoids, or anything similar, they would need to manually calculate a central tendency metric of the found clusters in order to use our metrics.

Any thoughts, @MaxHalford?

Lastly, regarding the code so far, and that's just my personal opinion/coding style, I would opt for shorter variable names as much as possible. I know that in some cases the longer names help the reader to understand the code. But even so, there are some cases where short identifiers also do the job.

smastelini:
@hoanganhngo610 also agreed in writing a notebook highlighting the differences between online clustering metrics and the batch ones (and also benchmarking them). Although many metrics bear the same names, their definition is somewhat different in incremental learning. We can brainstorm on the contents, organization, and where to place this notebook. Does that sound reasonable?

MaxHalford:
Thanks for the notes, it helps.

My opinion would be to put these clustering metrics in the `cluster` module.

smastelini:
> Lastly, regarding the code so far, and that's just my personal opinion/coding style, I would opt for shorter variable names as much as possible. I know that in some cases the longer names help the reader to understand the code. But even so, there are some cases where short identifiers also do the job.

For instance, I am still in favor of adopting `w` instead of `sample_weight` across our codebase. I believe the combination `x`, `y`, `w` is a clean and easy-to-remember pattern (and almost self-explainable). We would document that, for sure.

jacobmontiel:
> My opinion would be to put these clustering metrics in the `cluster` module.

This could be confusing to the user and would break the existing structure. I vote to keep them inside metrics. Also, some metrics (external) are used for both classification and clustering.

We can improve the documentation of this module. For example, the `scikit-learn` documentation has a table:

image

MaxHalford:
Yes you're probably right @jacobmontiel. I agree that a table would be very helpful :)

river/metrics/internal_clustering_metrics.py
```diff
@@ -110,68 +202,34 @@ class MSSTD(base_internal_clustering.InternalClusteringMetrics):
...
-class RMSSTD(MSSTD):
-    """Root mean squared standard deviation.
 
-    This is the square root of the pooled sample variance of all the attributes, which
-    measures only the compactness of found clusters.
+class DaviesBouldin(base_internal_clustering.InternalClusteringMetrics):
+    """Davies-Bouldin index (DB).
+
+    The Davies-Bouldin index (DB) is an old but still widely used inernal validaion measure.
+    DB uses intra-cluster variance and inter-cluster cener disance to find the worst partner
```
MaxHalford, line 230: center

hoanganhngo610: Fixed!

river/metrics/__init__.py
jacobmontiel: I vote for `InternalClusMetric `

river/metrics/base_internal_clustering.py
jacobmontiel: Given that we are using `x` and `y` in external metrics, I suggest we keep that convention:

```
update(self, x, y_pred, sample_weight, centers)
```

For example:
```
accuracy.update(y_true, y_pred, sample_weight)
ssq.update(x, y_pred, sample_weight, centers)
```

river/metrics/__init__.py
hoanganhngo610: I would also vote `InternalClusMetric`. In this case, I would opt to use this name first before we conclude on any further changes!

river/metrics/cluster/__init__.py
```diff
@@ -0,0 +1,33 @@
+"""Internal clustering metrics
+
+This submodule includes all internal clustering metrics that are updated with one sample,
+its label and the current cluster centers at a time. Using this, we can track the performance
+of the clustering algorithm without having to store information of all previously passed points.
+
+"""
+
+from .base import InternalClusMetric
```
MaxHalford, line 9: I would simply call this `Metric` now that's in the `cluster` submodule :)

hoanganhngo610: I'm just afraid that calling it `Metric` will cause a bit of confusion, since in clustering we still have internal metrics and external metrics. The external metrics are exactly the ones that River has had before, e.g. F1, Cohen, Kappa, Jaccard, etc.

river/metrics/cluster/calinskiharabasz.py
```diff
@@ -0,0 +1,125 @@
+import math
+
+from river import stats, utils
+
+from . import base
+
+__all__ = ["CalinskiHarabasz"]
```
smastelini, line 7: I am not sure `__all__` is needed here. This is the only class in this file, right? I think the import in `__init__.py` already does the job. The same would apply to other similar cases

river/metrics/cluster/calinskiharabasz.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class CalinskiHarabasz(base.InternalClusMetric):
+    """Calinski-Harabasz index (CH).
+
+    The Calinski-Harabasz index (CH) index measures the two criteria simultaneously
+    with the help of average between and within cluster sum of squares. The numerator
```
smastelini, line 14: I am not sure I follow:

"(...) measures the two criteria" -> "(...) measures two criteria"?

Are these criteria the numerator and denominator described after this sentence?

river/metrics/cluster/daviesbouldin.py
```diff
@@ -0,0 +1,109 @@
...
+from . import base
+
+__all__ = ["DaviesBouldin"]
+
+
+class DaviesBouldin(base.InternalClusMetric):
+    """Davies-Bouldin index (DB).
+
+    The Davies-Bouldin index (DB) [^1] is an old but still widely used inernal validaion measure.
+    DB uses intra-cluster variance and inter-cluster center disance to find the worst partner
```
smastelini, line 14: disance -> distance

river/metrics/cluster/rmsstd.py
```diff
@@ -0,0 +1,137 @@
...
+
+from river import utils
+
+from . import base
+
+__all__ = ["MSSTD", "RMSSTD"]
+
+
+class MSSTD(base.InternalClusMetric):
+    """Mean squared standard deviation.
```
smastelini, line 11: Please provide a brief description of the metric.

Besides, I would capitalize the letters to better match the acronym.

river/metrics/cluster/rmsstd.py
```diff
@@ -0,0 +1,137 @@
...
+        except ZeroDivisionError:
+            return math.inf
+
+    @property
+    def bigger_is_better(self):
+        return False
+
+
+class RMSSTD(MSSTD):
+    """Root mean squared standard deviation.
```
smastelini, line 93: Use capital letters :D

river/metrics/cluster/sd_validation.py
```diff
@@ -0,0 +1,175 @@
...
+
+    The SD validity index (SD) [^1] is a more recent clustering validation measure. It is composed of
+    two terms:
+
+    * Scat(NC) stands for the scattering within clusters,
+
+    * Dis(NC) stands for the dispersion between clusters.
+
+    Like DB and SB, SD measures the compactness with variance of clustered objects and separation
+    with distnace between cluster centers, but uses them in a different way. The smaller the value
```
smastelini, line 21: distnace -> distance

river/metrics/cluster/sd_validation.py
```diff
@@ -0,0 +1,175 @@
...
+    of SD, the better.
+
+    In the original formula for SD validation index, the ratio between the maximum and the actual
+    number of clusters is taken into account. However, due to the fact that metrics are update live
+    along with the data stream, this ratio will be automatically set to default as 1.
```
smastelini, line 26: Suggestion:

"(...) due to the fact that metrics are updated in an incremental fashion, this ratio (...)"

river/metrics/cluster/sd_validation.py
```diff
@@ -0,0 +1,175 @@
...
+                distance_ij = math.sqrt(
+                    utils.math.minkowski_distance(centers[i], centers[j], 2)
+                )
+                if distance_ij > max_distance_clusters:
+                    max_distance_clusters = distance_ij
+                if distance_ij < min_distance_clusters:
+                    min_distance_clusters = distance_ij
+                sum_inverse_distances += 1 / distance_ij
+
+        return max_distance_clusters / min_distance_clusters * sum_inverse_distances
```
smastelini, line 93: Are divisions by zero a concern here? (and possibly in other similar situations)

river/metrics/cluster/sd_validation.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @staticmethod
+    def _norm(x):
+        origin = {i: 0 for i in x}
+        return math.sqrt(utils.math.minkowski_distance(x, origin, 2))
+
+    def update(self, x, y_pred, centers, sample_weight=1.0):
+
+        if not self._initialized:
+            self._center_all_points = self._overall_variance = {
```
smastelini, line 103: Should both the variables point to the same region in memory?

If not, the initialization should be separate. Maybe other cases similar to this one escaped my attention.

Besides, is it possible to use `stats.Var` to keep the running variance? From my personal experience (and plenty of text-book references :P), it's always better to use the Welford algorithm whenever it's possible.

river/metrics/cluster/silhouette.py
```diff
@@ -0,0 +1,113 @@
...
+
+__all__ = ["Silhouette"]
+
+
+class Silhouette(base.InternalClusMetric):
+    """
+    Silhouette coefficient [^1], roughly speaking, is the ratio between cohesion and the average distances
+    from the points to their second-closest centroid. It rewards the clustering algorithm where
+    points are very close to their assigned centroids and far from any other centroids,
+    that is clustering results with good cohesion and good separation.
```
smastelini, line 15:
```suggestion
    that is, clustering results with good cohesion and good separation.
```

river/metrics/cluster/silhouette.py
```diff
@@ -0,0 +1,113 @@
+import math
+
+from river import utils
+
+from . import base
+
+__all__ = ["Silhouette"]
+
+
+class Silhouette(base.InternalClusMetric):
```
smastelini, line 10: Is this metric the one whose definition differs from the traditional batch version? If so, I advise adding a brief note warning the user.

smastelini (changes requested):
Hi @hoanganhngo610! Thanks for all the work! You did a lot!

I finished the first pass throughout the last changes and left some comments.

This PR is in good shape.

river/metrics/cluster/base.py
```diff
@@ -0,0 +1,78 @@
+import abc
+import numbers
+import typing
+
+from river import stats
+
+__all__ = ["InternalClusMetric"]
+
+
+class InternalClusMetric(abc.ABC):
```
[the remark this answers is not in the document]
hoanganhngo610: I inherited this from the `metrics/base.py` file, since the `Metric` in this file also use `abc.ABC`.

river/metrics/cluster/base.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self):
+        self._mean = stats.Mean()
+
+    @abc.abstractmethod
+    def _eval(
+        self,
+        x: typing.Dict[typing.Union[str, int], typing.Union[float, int]],
+        y_pred: numbers.Number,
+        centers,
+    ):
```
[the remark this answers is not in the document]
hoanganhngo610: I totally forgot about that! Would add it immediately!

river/metrics/cluster/calinskiharabasz.py
hoanganhngo610: You are right. `CalinskiHarabasz` is the only class in this file. I would proceed to delete all of them.

river/metrics/cluster/calinskiharabasz.py
hoanganhngo610: Yes indeed, two criteria taken into account by the CH index are the numerator and denominator described in the following sentences. To make it clearer, I can change the wording as you suggested, while separating the two criteria into bullets.

river/metrics/cluster/daviesbouldin.py
hoanganhngo610: Resolved!

river/metrics/cluster/rmsstd.py
hoanganhngo610: Resolved!

river/metrics/cluster/rmsstd.py
hoanganhngo610: This is pretty similar to the definition of `MSSTD`, which only removes the `root` part! I have added a brief description on that!
Capitalizing the letters resolved!

river/metrics/cluster/sd_validation.py
hoanganhngo610: Resolved!

river/metrics/cluster/sd_validation.py
hoanganhngo610: Really nice suggestion! Thank you so much!!
Resolved.

river/metrics/cluster/sd_validation.py
hoanganhngo610: Usually, since the number of centers passed through the function have a length different from `0`, the result would also be different from `0` (as I believe the user would be responsible enough :D). However, since a division by zero is also feasible, I would implement a `try - except` here and everywhere concerned.

river/metrics/cluster/__init__.py
MaxHalford: Ok but it's the same case for `R2` no? As long as the clustering metrics and the base class are in the `cluster` submodule I think it's fine.

river/metrics/cluster/base.py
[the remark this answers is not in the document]
MaxHalford: Yes for the moment let's not make metrics inherit from `Base`. We might do later.

river/metrics/cluster/base.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @abc.abstractmethod
+    def revert(self, x, y_pred, centers, sample_weight=1.0) -> "InternalClusMetric":
+        """Revert the metric."""
+
+    @abc.abstractmethod
+    def get(self) -> float:
+        """Return the current value of the metric."""
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
MaxHalford: Definitely should be a property.

river/metrics/cluster/base.py
```diff
@@ -0,0 +1,78 @@
...
+
+    """
+
+    def __init__(self):
+        self._mean = stats.Mean()
+
+    @abc.abstractmethod
+    def _eval(
+        self,
+        x: typing.Dict[typing.Union[str, int], typing.Union[float, int]],
```
MaxHalford, line 51: It's better to do this:

```py
from river.base.typing import FeatureName
from typing import Number

typing.Dict[FeatureName, Number]
```

Same goes for all the other type hints :)

river/metrics/cluster/calinskiharabasz.py
MaxHalford: Yes I'm not 100% sure on this. It does seem superfluous to have `__all__` when there's only one class.

river/metrics/cluster/sd_validation.py
hoanganhngo610: You are correct! Indeed, I previously thought that variance of whole dictionary can not be calculated, but it seems that calculating variance of each dimension using `stats.Var` is still feasible!

river/metrics/cluster/base.py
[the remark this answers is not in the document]
smastelini: I agree that this class should be abstract. The thing is, `metrics.Metric` also inherit from `base.Base`:

https://github.com/online-ml/river/blob/b829a57fae98c52022ba2ee52416e7cdb4b45e62/river/metrics/base.py#L23

Is there a reason not to do that here too?

river/metrics/cluster/sd_validation.py
smastelini: This seems to be a good compromise!

river/metrics/cluster/base.py
smastelini: +1 to that!

river/metrics/cluster/__init__.py
hoanganhngo610: You're right. I'm having a short conversation with @jacobmontiel this Thursday, and I think we can finalize on this real quick.

river/metrics/cluster/base.py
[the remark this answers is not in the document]
hoanganhngo610: Resolved!

river/metrics/cluster/calinskiharabasz.py
hoanganhngo610: In this case I would proceed to delete all `__all__` where there is only one class!

river/metrics/cluster/silhouette.py
hoanganhngo610: Resolved!

river/metrics/cluster/silhouette.py
hoanganhngo610: It is indeed different. I have also added an explanation on why and how is it different!

river/metrics/cluster/base.py
hoanganhngo610: Resolved!

smastelini: That's the solution :D

river/metrics/cluster/__init__.py
jacobmontiel: This will be renamed as `InternalMetric`

river/metrics/cluster/wb.py
```diff
@@ -0,0 +1,144 @@
+import math
+
+from river import stats, utils
+
+from . import base
+
+
+class WB(base.InternalClusMetric):
+    """R-Squared
```
jacobmontiel, line 9: The docstring does not match the class.

hoanganhngo610: Fixed!

river/metrics/cluster/__init__.py
hoanganhngo610: Resolved and changed!

river/metrics/cluster/bic.py
```diff
@@ -0,0 +1,150 @@
...
+from . import base
+
+
+class BIC(base.InternalMetric):
+    r"""Bayesian Information Criterion (BIC).
+
+    In statistics, the Bayesian Information Criterion (BIC) [^1], or Schwarz Information
+    Criterion (SIC), is a criterion for model selection among a finite set of models;
+    the model with the lowest BIC is preferred. It is based, in part, on  the likelihood
+    function and is closely related to the Akaike Information Criterion (AIC).
```
smastelini, line 14: Does the clustering BIC differ from the traditional one? In other words, although the results are always negative, we are seemingly looking for the highest ones, right? Maybe the interpretation is what changed. Could you please elaborate on that, @hoanganhngo610?

river/metrics/cluster/bic.py
```diff
@@ -0,0 +1,150 @@
...
+
+        const_term = 0.5 * self._n_clusters * math.log(total_points) * (self._dim + 1)
+
+        for i in self._n_points_by_clusters:
+            BIC += (
+                self._n_points_by_clusters[i] * math.log(self._n_points_by_clusters[i])
+                - self._n_points_by_clusters[i] * math.log(total_points)
+                - (self._n_points_by_clusters[i] * self._dim)
+                / 2
+                * math.log(2 * math.pi * variance)
```
smastelini, line 140: `2 * math.pi * variance` can be zero if the variance is zero. Better safe than sorry: what about adding try/except statements?

river/metrics/cluster/sd_validation.py
```diff
@@ -0,0 +1,152 @@
...
+                    utils.math.minkowski_distance(centers[i], centers[j], 2)
+                )
+                if distance_ij > max_distance_clusters:
+                    max_distance_clusters = distance_ij
+                if distance_ij < min_distance_clusters:
+                    min_distance_clusters = distance_ij
+                sum_inverse_distances += 1 / distance_ij
+
+        try:
+            return max_distance_clusters / min_distance_clusters * sum_inverse_distances
```
smastelini, line 92: What is the precedence here?

(A / B) * C

or

A / (B * C) ?

river/metrics/cluster/ssq_based.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+class WB(base.InternalMetric):
+    """WB Index
+
+    WB Index is a simple sum-of-square method, calculated by dividing the within
+    cluster sum-of-squares by the between cluster sum-of-squares. Its effect is emphasized
+    by multiplying the number of clusters. The advantages of the proposed method
+    are that it determines the number of clusters by minimal value of it without
+    any knee point detection method, and it is easy to be implemented.
```
smastelini, line 193: Suggested rewording:

"The advantages of the proposed method are that one can determine the number of clusters by minimizing the WB value, without relying on any knee point detection, and this metric is straightforward to implement."

river/metrics/cluster/ssw.py
```diff
@@ -0,0 +1,285 @@
...
+    def bigger_is_better(self):
+        return False
+
+
+class Xu(base.InternalMetric):
+    """Xu Index
+
+    Xu Index is among the most complicated sum-of-squares based metrics [^1].
+    It is calculated based on the Sum-of-Squares Within Clusters (SSW), total
+    number of points, number of clusters and dimension of the cluserting problem.
```
smastelini, line 194:
```suggestion
    number of points, number of clusters, and the dimension of the clustering problem.
```

smastelini (changes requested):
Hi @hoanganhngo610 and @jacobmontiel. I finished another pass through the code and left some minor comments.

Given the sheer volume of additions, some details might have escaped my attention, unfortunately. A general question kept coming to me: what is the effect of the default values raised in case of errors in an evaluation pipeline? Let's say that someone is using `progressive_val_score` to evaluate some cluster model and the metrics start to output `{+|-}math.inf`. What now? The aggregated values would be a mess, no? :P

Suggestions?

river/metrics/cluster/wb.py
```diff
@@ -0,0 +1,144 @@
+import math
+
+from river import stats, utils
+
+from . import base
+
+
+class WB(base.InternalClusMetric):
```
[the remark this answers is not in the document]
hoanganhngo610: This is an outdated version of the commit, since the implementation of `WB` has been merged into a new file called `ssq_based.py` :D

river/metrics/base.py
```diff
@@ -84,7 +84,7 @@ def bigger_is_better(self):
         return True
 
     def works_with(self, model) -> bool:
-        return utils.inspect.isclassifier(model)
+        return utils.inspect.isclassifier(model) or utils.inspect.isclusterer(model)
```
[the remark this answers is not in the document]
hoanganhngo610: I have thought about it when I updated this line. However, this is also the reason why I put `or` instead of `and`. This will also facilitate us in the evaluation file later on. @jacobmontiel Do you have any other thoughts?

river/metrics/cluster/bic.py
hoanganhngo610: Fairly reasonable! I will add a `try/except` statement here, also to align with the raise of the statement everywhere else.

river/metrics/cluster/bic.py
hoanganhngo610: You are correct. The BIC here is not actually different from the traditional ones. I have also stated below that although the values are always negative, we are still looking for the highest ones.

river/metrics/cluster/sd_validation.py
hoanganhngo610: The correct expression is (A / B) * C. Necessary changes have been made!

river/metrics/cluster/ssq_based.py
hoanganhngo610: Resolved!

river/metrics/cluster/ssw.py
hoanganhngo610: Resolved!

hoanganhngo610 (commented):
Thank you so much for your comments @smastelini! Necessary changes have been made, while there are 1 or 2 topics that are left open for discussion!

river/metrics/cluster/wb.py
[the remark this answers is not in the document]
smastelini: Cool! Thanks for your answer :)

river/metrics/base.py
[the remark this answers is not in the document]
smastelini: Although this still needs to be discussed, functionality-wise I am pretty happy with the results.

Hence, I'll mark this PR as approved, from my side!

smastelini (approved):
Thanks once again, @hoanganhngo610, for the great work! We are waiting for @jacobmontiel's opinion on one topic, but I'm already happy with the results!

river/metrics/base.py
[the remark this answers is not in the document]
jacobmontiel: We can override the `works_with` method in `MultiClassMetric` so binary metrics will only work with classifiers and multi-class metrics will work with both classifiers and clusteres.

jacobmontiel:
The last change did the trick. I am happy as well with the PR. Thanks @hoanganhngo610 !!!!

BrandonKMLee:
Sorry for asking, but it seems that some of the indices here cannot be found in the main repo.

PR #488, opened 2021-02-25
Make tree splitters public

smastelini:
As discussed in #476, currently the attribute observers, a.k.a. splitters, are picked by passing strings to the trees. Their parameters are set with dictionaries. This choice is not optimal and makes it difficult to document and add new splitters in the long run.

This PR makes the splitter selection similar to what is done with linear models' optimizers.

In summary, this PR brings to the table:

- [x] Rename: Attribute observer (AO) -> splitter.
- [x] Make splitters public and standardize their naming scheme.
- [x] Add a property to differentiate classification and regression splitters (suggestions for a better property name are welcomed).
- [x] Splitters can be passed as a parameter to the trees. They are "deepcopied" to new leaves as they are created.
- [x] Remove redundant portions of the tree code.
- [x] Change indirectly impacted methods (ARF variants).
- [x] Apply misc/minor improvements to the `tree` module, e.g., fix `mypy` issues and standardize some variable naming patterns. 
- [x] Improve documentation in `tree.splitter` and possible mentions to splitters in the trees' docstrings.

What still needs to be done:
- [x] ~Create a page in the documentation explaining and comparing different splits in standalone trees and ensembles thereof. `progressive_val_score` is a good option, of course, but we could also use Trackers (#471).~
- [x] Decide whether or not other internal parts of the `tree` module should be refactored/streamlined.
- [x] Ensure the final documentation renders nicely. :D


smastelini:
Those changes bring joy to my heart :smile:

However, this PR is just a stub yet. Suggestions are welcomed.

@MaxHalford and @jacobmontiel, thanks for the suggestions in #476. Is that what you had in mind?

river/tree/_nodes/efdtc_nodes.py
```diff
@@ -85,7 +76,7 @@ def best_split_suggestions(self, criterion, tree):
         best_suggestions = []
         pre_split_dist = self.stats
 
-        for idx, obs in self.attribute_observers.items():
+        for idx, obs in self.splitters.items():
```
smastelini, line 79: Forgot to change to `splitter`. I will do that later.

river/tree/_nodes/efdtc_nodes.py
```diff
@@ -123,44 +114,31 @@ class EFDTSplitNode(SplitNode, BaseEFDTNode):
...
+    splitter
         The numeric attribute observer algorithm used to monitor target statistics
         and perform split attempts.
-    attr_obs_params
-        The parameters passed to the numeric attribute observer algorithm.
-    attribute_observers
+    existing_splitters
         Existing attribute observers from previous nodes passed to provide a warm-start.
+    **kwargs
+        Other parameters passed to the learning nodes.
```
smastelini, line 123: What is the correct choice for docstrings? `kwargs` or  `**kwargs`?

river/tree/_nodes/htr_nodes.py
```diff
@@ -65,8 +51,8 @@ def manage_memory(self, criterion, last_check_ratio, last_check_vr, last_check_e
         last_check_e
             Hoeffding bound value calculated in the last split attempt.
         """
-        for obs in self.attribute_observers.values():
-            if isinstance(obs, NumericAttributeRegressionObserver):
+        for obs in self.splitters.values():
+            if isinstance(obs, EBSTSplitter):
```
smastelini, line 55: Need to change this too.

river/tree/base_hoeffding_tree.py
```diff
@@ -91,8 +93,8 @@ def __init__(
         merit_preprune: bool = True,
     ):
         # Properties common to all the Hoeffding trees
-        self._split_criterion: str
-        self._leaf_prediction: str
+        self._split_criterion: str = ""
+        self._leaf_prediction: str = ""
```
smastelini, line 97: Did that to avoid `mypy` errors.

river/tree/splitter/__init__.py
```diff
@@ -0,0 +1,21 @@
+from .base_splitter import Splitter
+from .ebst_splitter import EBSTSplitter
+from .exhaustive_splitter import ExhaustiveSplitter
```
smastelini, line 3: This also could be named `BSTSplitter`. Some papers do that if I recall it correctly. The regression version "extends" this algorithm, so it is called "Extended Binary Search Tree {Splitter} Attribute Observer" -> `EBSTSplitter`.

Comments, @jacobmontiel?

river/tree/splitter/base_splitter.py
```diff
@@ -1,11 +1,17 @@
...
 
 
-class AttributeObserver(metaclass=ABCMeta):
-    """Abstract class for observing the class data distribution for an attribute.
-    This observer monitors the class distribution of a given attribute.
+class Splitter(base.Estimator, metaclass=ABCMeta):
+    """Base class for the tree splitters.
+
+    Attribute Observers (AO) or Splitters are responsible by monitoring an input feature
+    and evaluating possible split points to them. They can also perform other tasks related
```
smastelini, line 12: Review that

river/tree/splitter/qo_splitter.py
```diff
@@ -2,26 +2,41 @@
...
+    On the other hand, if the radius is too large, fewer slots will be created, less memory and
+    running time will be required, but at the cost of coarse split suggestions.
+
+    QO assumes that all features have the same range. It is always advised to scale the features
+    to apply this splitter. That can be done using the `preprocessing` module. A good "rule of
+    thumb" is to scale data using `preprocessing.StandardScaler` and define the radius as a
+    proportion of the features' standard deviation. For instance, the default radius value
+    would correspond to one quarter of the normalized features' standard deviation (since the
+    scaled data has zero mean and unit variance). If the features come from normal
+    distributions, by following the empirical rule, roughly `32` hash slots will be created.
```
smastelini, line 32: The plan is to have something like that for each splitter. Adding documentation to this one was easy because I am the author hahaha

For the other ones, any help is welcomed! @jacobmontiel, you probably know more than I do about many of the other splitters. If you want and have time for that, I would be happy to collaborate on this documentation workload.

MaxHalford:
Looking good!

smastelini (commented):
I left some comments regarding my own changes. Some of them are reminders to myself, others are questions for further discussion. As I see it, the splitters will need a boost in documentation to facilitate their usage by the users.

river/tree/_nodes/efdtc_nodes.py
MaxHalford: `kwargs`

river/tree/_nodes/efdtc_nodes.py
smastelini: Changed in commit [d6931d9](https://github.com/online-ml/river/pull/488/commits/d6931d9e1632978aeb7276b9a13cff1ac2a3ec09).

river/tree/_nodes/efdtc_nodes.py
smastelini: Thanks! I corrected the docstrings in commit [d6931d9](https://github.com/online-ml/river/pull/488/commits/d6931d9e1632978aeb7276b9a13cff1ac2a3ec09).

river/tree/_nodes/htr_nodes.py
smastelini: Changed in commit [d6931d9](https://github.com/online-ml/river/pull/488/commits/d6931d9e1632978aeb7276b9a13cff1ac2a3ec09).

river/tree/splitter/base_splitter.py
smastelini: Changed in commit [d6931d9](https://github.com/online-ml/river/pull/488/commits/d6931d9e1632978aeb7276b9a13cff1ac2a3ec09).

river/tree/splitter/qo_splitter.py
smastelini: I improved the documentation a bit in a recent commit.

smastelini:
Since #471 is not ready yet, I think we can merge this PR when we are happy with the changes and wait for the `Tracks` to update our documentation page.

Again, I intend to provide some benchmarks of the splitters using synthetic datasets.

What do you think, @jacobmontiel and @MaxHalford?

smastelini:
Hey @MaxHalford and @jacobmontiel, I added a page in the user guides that is a walkthrough in the tree module. I talked about the different tree models available, model inspection, memory management, and splitters. Hope that can help the users.

Your feedback is welcomed.

(Later today I'll add some basic coverage tests for the splitters)

smastelini:
The failing tests are related to classifier chains. I believe they were recently updated. Pinging @MaxHalford.

smastelini:
All set. If you require more changes before merging, please let me know, @jacobmontiel and @MaxHalford.

river/tree/__init__.py
```diff
@@ -7,11 +9,13 @@
 from .label_combination_hoeffding_tree import LabelCombinationHoeffdingTreeClassifier
 
 __all__ = [
+    "splitter",
+    "HoeffdingTree",
```
MaxHalford, line 13: Not sure we want to expose this base class, right?

river/tree/_nodes/base.py
```diff
@@ -1,13 +1,14 @@
...
+import typing
 from abc import ABCMeta, abstractmethod
-from typing import Dict, Iterator, List, Union
 
 from river import base
 from river.stats import Var
-from river.tree._attribute_test import AttributeSplitSuggestion  # noqa
-from river.tree._attribute_test import InstanceConditionalTest  # noqa
+
+from .._attribute_test import AttributeSplitSuggestion, InstanceConditionalTest
```
MaxHalford, line 11: I would call this simply `SplitSuggestion` instead of `AttributeSplitSuggestion`, but your call!

MaxHalford (commented):
Good job, I mean it! It looks like you put in a lot of work. The code looks much nicer IMO. And the fact that the parameters are not duplicated everywhere in the docs feels better too.

river/tree/__init__.py
smastelini: It's kind of tricky. It has some important documentation, that's the only reason it's public right now. I'm not sure this is the best approach, but it was a way to avoid repetition.

river/tree/_nodes/base.py
smastelini: I like that. I am so used to this code that I didn't realize it was a long name till you mentioned it. I'll rename it tomorrow :)

river/tree/__init__.py
MaxHalford: I'm not 100% but I think we can play with the `__doc__` attribute of classes. So you could have some string that contains the doc, and just add it the `__doc__` of the class you want to add the documentation to. I've never done it but I've seen it done in (serious) projects :)

river/tree/_nodes/base.py
smastelini: Renamed in [2e63165](https://github.com/online-ml/river/pull/488/commits/2e63165b10a2c12af7ada79bc606cfb5c44805ad).

river/tree/base_hoeffding_tree.py
```diff
@@ -23,11 +23,11 @@
     GRAPHVIZ_INSTALLED = False
 
 
-class BaseHoeffdingTree(ABC):
+class HoeffdingTree(ABC):
```
jacobmontiel, line 26: I would rather keep the name as `BaseHoeffdinTree` as we do in other places (such as for ensembles). As previously mentioned, this class should not be public.

river/tree/extremely_fast_decision_tree.py
```diff
@@ -46,21 +47,14 @@ class ExtremelyFastDecisionTreeClassifier(HoeffdingTreeClassifier):
...
-        evaluate (`n_splits` -- defaults to `32`) can be adjusted.</br>
-        See 'Notes' for more information about the AOs.
-    attr_obs_params
-        Parameters passed to the numeric AOs. See `attr_obs` for more information.
+    splitter
+        The Splitter or Attribute Observer (AO) used to monitor the class statistics of numeric
+        features and perform splits. Splitters are available in the `tree.splitter` module.
+        Different splitters are available for classification and regression tasks. Classification
+        and regression splitters can be distinguished by their property `is_target_class`.
+        By default, `tree.splitter.GaussianSplitter` is used if `splitter` is `None`.
```
jacobmontiel, line 55: I would add: "This is an advanced option. Special care must be taken when choosing a different splitter." The same applies to other descriptions similar to this one.

river/tree/splitter/ebst_splitter.py
```diff
@@ -280,3 +243,56 @@ def _remove_bad_split_nodes(self, current_node, parent=None, is_left_child=True)
                 return True
 
         return False
+
+
+class Node:
```
jacobmontiel, line 248: In excess of caution, should we name this as `EBSTNode`?

river/tree/splitter/base_splitter.py
```diff
@@ -14,30 +21,32 @@ def __init__(self):
...
+    def cond_proba(self, att_val, class_val: base.typing.ClfTarget) -> float:
         """Get the probability for an attribute value given a class.
 
         Parameters
         ----------
         att_val
             The value of the attribute.
-        target_val
-            The target value.
+        class_val
```
jacobmontiel, line 48: I suggest keeping it as `target_val`, for consistency with the `update` method.

river/tree/splitter/exhaustive_splitter.py
```diff
@@ -194,3 +170,30 @@ def _search_for_best_split_option(
         )
 
         return current_best_option
+
+
+class Node:
```
jacobmontiel, line 175: Same as before, what about `ExhaustiveNode`? 

This is optional.

river/tree/splitter/histogram_splitter.py
```diff
@@ -76,28 +76,29 @@ def best_evaluated_split_suggestion(
...
-                merit = criterion.merit_of_split(pre_split_dist, post_split_dist)
-
-                if best_suggestion is None or merit > best_suggestion.merit:
-                    num_att_binary_test = NumericAttributeBinaryTest(
-                        att_idx, at, equal_passes_test=False
-                    )
-                    best_suggestion = AttributeSplitSuggestion(
-                        num_att_binary_test, post_split_dist, merit
-                    )
+                if y in cdfs:
```
jacobmontiel, line 86: Could you elaborate on this? Code after `if` is no longer part of the loop?

river/tree/splitter/nominal_class_splitter.py
```diff
@@ -1,16 +1,16 @@
...
 )
-from .attribute_observer import AttributeObserver
+from .base_splitter import Splitter
 
 
-class NominalAttributeClassObserver(AttributeObserver):
-    """Class for observing the class data distribution for a nominal attribute.
-    This observer monitors the class distribution of a given attribute.
-    Used in naive Bayes and decision trees to monitor data statistics on leaves.
+class NominalClassSplitter(Splitter):
```
jacobmontiel, line 9: The new names are kind of confusing. As they got shorter what about: `NominalSplitterClassif` and `NominalSplitterReg` respectively?

jacobmontiel (changes requested):
Awesome work @smastelini 

I really like how the new naming and structure tidies-up the code. Nice catch on those unnecessary parts of the code. The QO splitter refactoring is also nice. Not sure if it is listed in the PR description.

I left a couple of comments and changes. In general, the PR looks good.

Let me know if you need further information.

river/tree/splitter/__init__.py
smastelini: Let's keep the name ExhaustiveSplitter since this is the common choice in the literature.

river/tree/__init__.py
smastelini: That's a good point. Keep in mind, however, that we do have some other abstract classes that are exposed in the documentation. For example, `optim.Optimizer`. I am not sure whether or not this was intended.

> On the other hand, it is unlikely that people will find or search for the documentation in a different class.

That is a valid point. Currently, we point to `tree.HoeffdingTree` from the other trees. Something like: "for more information check `tree.HoeffdingTree`".

Again, I would rather keep the plethora of somewhat obscure (yet so useful) parameters documented in the base class than exposing them everywhere and confusing the user with too many choices to pick from. A page in the documentation could do the work, but I find it strange to document parameters outside of a docstring. If we opt for the first option and replicate the parameter information everywhere, we can follow Max's suggestion.

[`matplotlib`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html) also relies a lot on `**kwargs`. Maybe we could provide some pieces of information about what the user can do and then point to the base class. Hence, we balance the amount of the "data overhead" for the user.

river/tree/_nodes/base.py
```diff
@@ -80,25 +82,6 @@ def path(self, x) -> Iterator["Node"]:
     def iter_edges(self):
         yield None, 0, None, self, None
 
-    @property
```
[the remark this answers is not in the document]
smastelini: I see your point. However, as we don't make copies of the internal structures when returning the properties, one could still do something like:

```python
n_stats = node.stats  # property access

n_stats['class_name'] = 100000

print(node.stats['class_name'])
>> 100000
```
and the node's statistics would still be changed, right? :P

But nodes are a private module that is only accessed by the trees internally. So, it is (probably) fine to keep the nodes' attributes public. What do you think?

river/tree/base_hoeffding_tree.py
smastelini: Regardless of making this class public or not, I still do prefer shorter names hahaha

I think we need to define which pattern we are going to use across the codebase. For instance, we have:

- `base.Classifier` (and its friends)
- `metrics.Metric`

but we also have:

- `ensemble.BaseForest`
- `base.WrapperMixin`

I don't think this is a pressing matter whatsoever, but it would be nice to use the same pattern everywhere. Personally, I prefer using shorter names as much as possible: Classifier, Metric, Forest, HoeffdingTree, etc.

What do you think, @jacobmontiel and @MaxHalford ?

river/tree/splitter/base_splitter.py
smastelini: I decided to use `class_val` to make it clear that conditional probabilities only apply to the classification case.
What do you think? I am totally cool with both of them.

river/tree/splitter/nominal_class_splitter.py
smastelini: Yes! Much better haha

What was I thinking? Nominal Reg/Class Splitter, LOL

Changed in commit [b71f853](https://github.com/online-ml/river/pull/488/commits/b71f853f5379e0e61b1b70902babc2ff3a4f3ce0).

river/tree/_attribute_test/__init__.py
```diff
@@ -1,13 +1,15 @@
-from .attribute_split_suggestion import AttributeSplitSuggestion
 from .instance_conditional_test import InstanceConditionalTest
 from .nominal_attribute_binary_test import NominalAttributeBinaryTest
```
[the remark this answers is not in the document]
smastelini: Wonderful! Renamed in commit [9aeda54](https://github.com/online-ml/river/pull/488/commits/9aeda5439cde9080d45254a45d4044aa885e36d8).

river/tree/splitter/ebst_splitter.py
smastelini: I agree with you, it is better to make things clear. Renamed in commit [3b4fed1](https://github.com/online-ml/river/pull/488/commits/3b4fed13fee51b7f77a3b5226ac24f27894f44ff).

river/tree/splitter/exhaustive_splitter.py
smastelini: Renamed in commit [3b4fed1](https://github.com/online-ml/river/pull/488/commits/3b4fed13fee51b7f77a3b5226ac24f27894f44ff).

river/tree/_nodes/arf_htc_nodes.py
```diff
@@ -112,22 +89,21 @@ class RandomLearningNodeNB(BaseRandomLearningNode, LearningNodeNB):
         Initial class observations.
     depth
         The depth of the node.
-    attr_obs
+    splitter
         The numeric attribute observer algorithm used to monitor target statistics
         and perform split attempts.
-    attr_obs_params
-        The parameters passed to the numeric attribute observer algorithm.
-    max_features
```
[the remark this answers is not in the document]
smastelini: My bad. I'll fix it!

river/tree/_nodes/arf_htc_nodes.py
```diff
@@ -84,23 +61,23 @@ class RandomLearningNodeMC(BaseRandomLearningNode, LearningNodeMC):
...
-        The parameters passed to the numeric attribute observer algorithm.
     max_features
         Number of attributes per subset for each node split.
     seed
         If int, seed is the seed used by the random number generator;
         If RandomState instance, seed is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    kwargs
+        Other parameters passed to the learning nodes the ARF implementations randomize.
```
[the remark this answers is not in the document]
smastelini: I don't even know anymore :D. It doesn't make much sense, so I removed this portion (commit [6430a0e](https://github.com/online-ml/river/pull/488/commits/6430a0e4df6ce0fc2f07fc228384a93716c2f1da)). Thanks for mentioning that.

river/tree/_nodes/hatc_nodes.py
```diff
@@ -191,7 +191,7 @@ def n_leaves(self):
         num_of_leaves = 0
         for child in self._children.values():
             if child is not None:
-                num_of_leaves += child.n_leaves
+                num_of_leaves += child.n_leaves  # noqa
```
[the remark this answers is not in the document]
smastelini: Nice! Added in [a0ff412](https://github.com/online-ml/river/pull/488/commits/a0ff4122f2263214bf3cff25b9412655b0caa3e7)

river/tree/splitter/histogram_splitter.py
smastelini: This was a bug :/

We need to iterate over all classes in `pre_split_dist` before evaluating the resulting post-split distribution. We didn't have tests for the different splitters until now so that one escaped my attention.

river/tree/_nodes/base.py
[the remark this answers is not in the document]
jacobmontiel: Agree. Nodes are internal classes and it is nice to avoid the overhead from the properties. On the other hand, we must be careful to protect the nodes when exposing results to the user at the estimator level.

river/tree/extremely_fast_decision_tree.py
smastelini: I added this comment in the last commit!

river/tree/splitter/base_splitter.py
smastelini: I updated all the occurrences to `target_val` for consistency.

river/tree/__init__.py
smastelini: @MaxHalford, would it be possible to add this portion of the documentation at the "index" page of the tree module? We do not have that now, but it could be a viable solution.

I am also convinced that `HoeffdingTree` should not be public.

MaxHalford: Yes we can and should documentation in the `__init__.py` file if that's what you mean.

river/tree/base_hoeffding_tree.py
MaxHalford: Actually my opinion is that we should not add the `Base` prefix. In any case these classes are not to be user-facing.

river/tree/__init__.py
smastelini: Sounds like a plan. Probably I will not use the same notation, (the `Parameters` section, for instance) but it will do the work!

smastelini:
I updated the documentation in the last split to make `tree.HoeffdingTree` hidden again. @jacobmontiel, please let me know if the documentation placement is satisfactory now :D

smastelini:
Disclaimer: the failing test is unrelated to this PR.

jacobmontiel (approved):
Thanks for the last changes @smastelini 
Issues found will be addressed in another PR.

smastelini:
Thanks for all your help and support, @jacobmontiel and @MaxHalford!

Issue #492, opened 2021-03-03
Unspecified C++ Build Tools dependency on Windows


ColdTeapot273K:
## Versions

**river version**: current master (03.03.2021)
**Python version**: 3.8.8
**Operating system**: Windows 10 Pro v2004

## Describe the bug
When installing `river`, build fails because it requires C++ build tools. The dependency is not specified in the docs. but think I installed `river` 2 weeks ago with `--no-deps` flag with no problems (cannot reproduce though)
```bash
Complete output (14 lines):
    running develop
    running egg_info
    writing river.egg-info\PKG-INFO
    writing dependency_links to river.egg-info\dependency_links.txt
    writing requirements to river.egg-info\requires.txt
    writing top-level names to river.egg-info\top_level.txt
    writing manifest file 'river.egg-info\SOURCES.txt'
    reading manifest file 'river.egg-info\SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    writing manifest file 'river.egg-info\SOURCES.txt'
    running build_ext
    building 'river.metrics.confusion' extension
    error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
```

## Steps/code to reproduce

```bash
pip install git+https://github.com/online-ml/river --upgrade
```
or pull git repo and:
```bash
pip install -e ".[dev]"
```

MaxHalford:
See [this thread](https://github.com/online-ml/river/issues/429).

MaxHalford:
Should be good now.

PR #493, opened 2021-03-03
Add instructions for Windows/Windows + Anaconda users

ColdTeapot273K:
Add instructions to address 
[this issue](https://github.com/online-ml/river/issues/492#issue-820893113)

and [this issue](https://stackoverflow.com/questions/50103486/installation-graphviz-no-module-named-graphviz/59365894#59365894)

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,12 @@ Or, through SSH:
 pip install git+ssh://git@github.com/online-ml/river.git --upgrade
 ```
 
+For **Windows** users:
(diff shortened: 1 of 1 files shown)


MaxHalford:
Thanks for this! However, let's wait for our package to be on River, in which case there will be wheels that will avoid these problems entirely. We're currently still waiting on the name to be released...

ColdTeapot273K:
> Thanks for this! However, let's wait for our package to be on River, in which case there will be wheels that will avoid these problems entirely. We're currently still waiting on the name to be released...

Yeah, seems to be taking awhile...

Although the second issue (about graphviz) won't be resolved this way. Alternative is to add python-graphviz as a conda-windows-specific dependency into setup.py/config

MaxHalford:
Closing this because we can now install wheels from PyPI :)

PR #501, opened 2021-03-05
Implementation of STREAMKMeans

hoanganhngo610:
This is the implementation of STREAMKMeans, an alternative algorithm of STREAMLSEARCH using KMeans proposed by [O'Callanghan et al.](https://ieeexplore.ieee.org/document/994785).

However, in this implementation, instead of using the traditional static KMeans, we will use the incremental KMeans in River. This reduces a significant amount of computation, since now we only need to update the KMeans instead of redoing it after every chunk of points are passed through.


river/cluster/streamkmeans.py
```diff
@@ -0,0 +1,128 @@
...
+        self.n_clusters = n_clusters
+        self.chunk_size = chunk_size
+        self.kwargs = kwargs
+
+        self._kmeans = cluster.KMeans(n_clusters=self.n_clusters, **self.kwargs)
+        self._temp_chunk = {}
+        self.centers = {}
+
+    @staticmethod
+    def _distance(point_a, point_b):
```
jacobmontiel, line 97: This is a good chance to evaluate the need of this method. Given that it wraps the `minkowski_distsnce` utility function. We could simplify the implementation by directly using the utility function. We would need to propagate this decision into the other methods where the same method is duplicated.

jacobmontiel: After looking again it seems that this method is not actually used. My previous comment is still valid for other methods though, but we could address that in a different PR.

jacobmontiel (changes requested):
Thanks @hoanganhngo610 

Just one comment from my review. Please take a look.

river/cluster/streamkmeans.py
hoanganhngo610: I totally agree. However, the only reason I wrap the `minkowski_distance` within `get_distance` function is to make the code cleaner and shorter. Based on the actual need of the algorithm, we can choose whether to implement this function or not. However, at this stage, I would suggest to call it in every new algorithm, whenever necessary. We can address this problem overall in one of the final commits.

Regarding this algorithm, I wil delete this `_distance` function since it is redundant.

hoanganhngo610:
@jacobmontiel Thank you so much for your comment! I believe that the PR is ready to merge now!

jacobmontiel (approved):

jacobmontiel:
Nice work @hoanganhngo610 

The PR is ready.

Issue #499, opened 2021-03-04
predict_one has side effects


Styren:
## Versions

**river version**: master branch
**Python version**: 3.9
**Operating system**: osx

## Describe the bug

`predict_one` causes the weights to change and yields a different result. When we tried to do time series forecasting and predicting a value in the future we noticed that the weights we got was different - and it failed to learn even when both features and labels consistently increased.
Am I misunderstanding something here?
## Steps/code to reproduce

```python
  from river import linear_model, metrics, preprocessing, compose

  dataset = [(dict(val=5*x), x) for x in range(10)]

  model = compose.Pipeline(
      ('scale', preprocessing.StandardScaler()),
      ('lin_reg', linear_model.LinearRegression())
  )
  metric = metrics.MSE()

  for x, y in dataset:
      y_pred = model.predict_one(x)
      model.learn_one(x, y)
      metric.update(y, y_pred)
      print(metric)

  model_pred_future = compose.Pipeline(
      ('scale', preprocessing.StandardScaler()),
      ('lin_reg', linear_model.LinearRegression())
  )
  metric_pred_future = metrics.MSE()


  for x, y in dataset:
      y_pred = model_pred_future.predict_one(x)
      model_pred_future.learn_one(x, y)
      metric_pred_future.update(y, y_pred)

      y_pred = model_pred_future.predict_one(dict(val=200000))
      print(metric_pred_future)
      print(f"y_pred_future: {y_pred}")
```
Output:
```
MSE: 0.
MSE: 0.5
MSE: 1.608
MSE: 3.23633
MSE: 5.297142
MSE: 7.709677
MSE: 10.402261
MSE: 13.312451
MSE: 16.386476
MSE: 19.578424
MSE: 0.
y_pred_future: 0.0
MSE: 0.5
y_pred_future: 0.005858394718508348
MSE: 1.624937
y_pred_future: 0.01308476518140858
MSE: 3.322002
y_pred_future: 0.02086116442166372
MSE: 5.539981
y_pred_future: 0.028884526973104035
MSE: 8.230667
y_pred_future: 0.037010755533317063
MSE: 11.349102
y_pred_future: 0.0451613383229445
MSE: 14.853522
y_pred_future: 0.053289764175274146
MSE: 18.705213
y_pred_future: 0.06136696916931006
MSE: 22.868334
y_pred_future: 0.06937418419188068
```

MaxHalford:
[...]

PR #503, opened 2021-03-06
Add no learn parameter to pipeline transformations

Styren:
As described in #499 it's not possible to predict without updating the scalers. This PR adds the `no_learn` parameter that avoids training the non-supervised parts.

I'm not sure this is the best solution so lets have a discussion on this.

Firstly without beeing a "clean code" zealot I still feel like adding binary arguments should generally be avoided in favour of separate methods with descriptive naming - is there no better name for the process of calculating a predictions while also updating weights than `predict_one`? Backwards compatibility notwithstanding. Or should we name the non-learning methods `predict_nolearn_one`, for example?

Secondly, if this is the proper solution - should it rather be named `learn` or `update_transformers` and default to true? As an added negation also adds some unnecessary complexity.

### DIFF
--- a/river/compose/pipeline.py
+++ b/river/compose/pipeline.py
@@ -324,7 +324,7 @@ class Pipeline(base.Estimator):
 
         return self
 
-    def _transform_one(self, x: dict, no_learn: bool=False):
+    def _transform_one(self, x: dict, no_learn: bool = False):
         """This methods takes care of applying the first n - 1 steps of the pipeline, which are
         supposedly transformers. It also returns the final step so that other functions can do
         something with it.
@@ -351,7 +351,7 @@ class Pipeline(base.Estimator):
 
         return x, next(steps)
 
-    def transform_one(self, x: dict, no_learn: bool=False):
+    def transform_one(self, x: dict, no_learn: bool = False):
         """Apply each transformer in the pipeline to some features.
 
         The final step in the pipeline will be applied if it is a transformer. If not, then it will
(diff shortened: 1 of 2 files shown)


MaxHalford:
> I'm not sure this is the best solution so lets have a discussion on this.

It's good that you're giving this some thought. It is a big decision.

> Firstly without beeing a "clean code" zealot I still feel like adding binary arguments should generally be avoided in favour of separate methods with descriptive naming

I would tend to agree. But I don't like the idea of introducing another method that won't see much usage.

> is there no better name for the process of calculating a predictions while also updating weights than predict_one?

The thing is that updating transformers during the prediction phase is exactly what we want to do in an online setting. Therefore, it should be the default behavior. So keeping the name `predict_one` is fine.

> Or should we name the non-learning methods predict_nolearn_one, for example?
 
I wouldn't go down this road. I think it's a good idea to have this `no_learn` parameter that we document in the docstring of `Pipeline`.

> Secondly, if this is the proper solution - should it rather be named learn or update_transformers and default to true? As an added negation also adds some unnecessary complexity.

I agree, the added negation isn't a good idea. I would call this `learn_unsupervised` and make it default to `True`. `update_transformers` isn't exact as there is a distinction between unsupervised and supervised transformers.

Styren:
> I wouldn't go down this road. I think it's a good idea to have this `no_learn` parameter that we document in the docstring of `Pipeline`.
> I agree, the added negation isn't a good idea. I would call this `learn_unsupervised` and make it default to `True`. `update_transformers` isn't exact as there is a distinction between unsupervised and supervised transformers.

No you're right it might not see much use.

I agree, `learn_unsupervised` is a much better name.

emreguenduez:
Hey there! I recently got the same issue and found that it was discussed in #499, from there I found this link and saw that the `no_learn `parameter was added as stated in [this commit](https://github.com/online-ml/river/pull/503/commits/5afbf72ca054d1efcbf8693123260d4103b1a463).  However, when I try to use it in the `predict_one` method, I get the error `predict_one() got an unexpected keyword argument 'no_learn'`. Was this parameter not added to the method? 

Here's the code to reproduce:

```
for xi, yi in stream.iter_pandas(x_test, y_test):
    y_forecast = model.predict_one(xi, no_learn = True)
    forecasts.append(y_forecast)
forecasts
```

MaxHalford:
Hey @emreguenduez! You have to use [`pure_inference_mode`](https://riverml.xyz/latest/api/utils/pure-inference-mode/) for that

PR #518, opened 2021-03-13
Implementation of Evostream for clustering

hoanganhngo610:
evoStream is a fairly recent streaming algorithm (2018) that is based on genetic algorithms (GA). The paper of this algorithm can be found [here](https://www.sciencedirect.com/science/article/abs/pii/S2214579618300649?via%3Dihub).

More information about genetics algorithms can be found at [the original paper in 2000](https://www.sciencedirect.com/science/article/abs/pii/S0031320399001375?via%3Dihub).


river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+    it will be removed. Then, all clusters whose centers are less than `r` from
+    each other are merged.
+
+    * When the stream is not initiated and the number of cluster is at least equal
+    to the initialization threshold, a cluster population of size `P` will be generated
+    by choosing randomly micro-clusters from the set of currently available micro-clusters.
+
+    **Offline generation of macro clusters (evolutionary step)**
+
+    The offline generation of macro clusters is based upon the original GA clustering algirthm,
```
smastelini, line 54: There is a typo in "algorithm" :)

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+    as follows:
+
+    * Using the roulette wheel algorithm, choose 2 arbitrary cluster solutions proportionally
+    to their fitness from the cluster population.
+
+    * Generate two offsprings using binary crossover.
+
+    * Generate a random number `delta` within the range `[0,1]`. If `delta < P_m`, each gene `g_i`
```
smastelini, line 62: Maybe it's worth mentioning that `P_m` is the mutation probability (so the users can link this variable to the parameter).

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+    within these two offsprings will be mutated with the following rules
+
+        - If `g_i = 0`, `g_i` will be equal to either `-2 * delta` or `2 * delta`, with equal probability
```
smastelini, line 65: If I understood it correctly, in case `g_i = 0`, it is set to `2 * delta`. Please check line 7 of Algorithm 3 in the original paper. Am I wrong?

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+        Size of the cluster population generated from the set of micro clusters.
+    n_clusters
+        Number of clusters generated by the algorithm.
+
+    Attributes
+    ----------
+    clusters
+        A set of final clusters of type `evoStreamMicroCluster`, i.e clusters with structure
+        `(center, last_update, weight)`. These clusters are formed by choosing the cluster solution
+        with highest fitness.
```
smastelini, line 104: "the highest fitness"

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+
+    * Choose the solution with the highest fitness and assign it as the clusters for the whole algorithm.
+
+    Parameters
+    ----------
+    radius
+        Radius threshold to determine neighboring clusters.
+    decay_rate
+        Parameter that controls the importance of historical data to current cluster.
+        Note that `decay_rate` is usually sufficiently small and has to be different from `0`.
```
smastelini, line 80: Suggestion: "Note that `decay_rate` should be sufficiently small and greater than zero."

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+    centers
+        Final clusters' centers.
+
+    References
+    ----------
+    [^1]: Carnein, M., Trautmann, H. (2018). evoStream – Evolutionary Stream Clustering Utilizing Idle Times.
+          Big Data Research, 14, 101-111. DOI: 10.1016/j.bdr.2018.05.005
+    [^2]: Maulik, U., Bandyopadhyay, S. (2000). Genetic algorithm-based clustering technique.
+          Pattern Recognition, 33(9), 1455-1465. DOI: 10.1016/s0031-3203(99)00137-5
+    [^3]: Hahsler, M., Bolanos, M. Clsutering Data Streams Based on Shared Density between Micro-Clusters,
```
smastelini, line 114: Typo in "Clustering" :D

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self._evolution_batch = {}
+        self._micro_clusters = {}
+        self._cluster_population = {}
+        self._fitness_cluster_population = {}
+
+    @staticmethod
+    def _binary_crossover(cluster_sol_1, cluster_sol_2):
+        dim = len(cluster_sol_1[0].center)
+        cutoff = math.floor(dim / 2)
```
smastelini, line 178: It is not clear to me from the paper whether the cut-off point is "always fixed". Keep in mind that I only skimmed the paper, though. Is the mid-point always used or the cut-off is also randomly defined?

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+                            -self.decay_rate
+                            * (self.time_stamp - micro_cluster.last_update)
+                        )
+                    )
+                    + 1
+                )
+                merged_status = True
+
+        if not merged_status:
+            self._micro_clusters[len(self._micro_clusters)] = temp_mc
```
smastelini, line 249: Are the "keys" of the micro-cluster dict used elsewhere? I don't think it is a bottleneck, but a combination of a list and the append method should be slightly faster. Maybe, the same observation applies to other portions of the code.

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _fitness(self, points, clusters):
+        ssq = 0
+        for i, point_i in points.items():
+            closest_cluster_index_i = self._find_closest_cluster_index(
+                point_i, clusters
+            )
+            distance = self._distance(point_i, clusters[closest_cluster_index_i].center)
+            ssq += distance * distance
+        return 1 / ssq
```
smastelini, line 259: What about a try/except block here to avoid division by zero? (This situation should not be common in practice, but who knows...)

river/cluster/evostream.py
smastelini: But then there's the culprit: micro cluster removal, right? Dictionaries could be the simplest/fastest choice in the end.

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _roulette_wheel_selection(self, fitness_clusters):
```
smastelini, line 261: Is this random selection procedure the same one used in the paper?

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+        }
+
+        p_1, p_2 = self._roulette_wheel_selection(self._fitness_cluster_population)
+        o_1, o_2 = self._binary_crossover(p_1, p_2)
+        delta = random.uniform(0, 1)
+
+        def mutate(g):
+            for i in g:
+                if g[i] == 0:
+                    g[i] = 2 * delta if random.uniform(0, 1) < 0.5 else -2 * delta
```
smastelini, line 291: A random number generator should be used instead. The user can pass a seed that will be used to initialize the generator:

```python
# in __init__
...

self.seed = seed
self._rng = random.Random(self.seed)

```

smastelini: The same applies to all the random numbers used. You can use `self._rng.uniform` once the generator is created. By default `seed` is `None`.

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+        self._fitness_cluster_population = {
+            i: self._fitness(self._evolution_batch, self._cluster_population[i])
+            for i in self._cluster_population
+        }
+
+        p_1, p_2 = self._roulette_wheel_selection(self._fitness_cluster_population)
+        o_1, o_2 = self._binary_crossover(p_1, p_2)
+        delta = random.uniform(0, 1)
+
+        def mutate(g):
```
smastelini, line 288: What about using a `@staticmethod` instead?

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+                2
+                ** (
+                    -self.decay_rate
+                    * (self.time_stamp - self._micro_clusters[i].last_update)
+                )
+            )
+            if self._micro_clusters[i].weight < 2 ** (
+                -self.decay_rate * self.cleanup_interval
+            ):
+                self._micro_clusters.pop(i)
```
smastelini, line 352: If the returned value will not be used and `i` is expected to be in the dict, why not using `del` instead?

river/cluster/evostream.py
```diff
@@ -0,0 +1,424 @@
...
+        return self
+
+    def predict_one(self, x, sample_weight=None):
+
+        label = self._find_closest_cluster_index(x, self.clusters)
+
+        return label
+
+
+class evoStreamMicroCluster(metaclass=ABCMeta):
```
smastelini, line 408: Why this class is linked to `ABCMeta`?

Besides that, as a possible performance trick, you check the usage of `__slots__` in this class :)

smastelini (changes requested):
Hi @hoanganhngo610, thanks for your hard work on this PR! The idea behind evoStream is pretty neat and I was happy to read a bit about it. I am already quite happy with the current shape of this PR, but there are a few points that still need to be polished before merging. I left some comments/requests.

My main concern, and that's more of a "philosophical" nature is the concept of "idle" that is proposed. In the original algorithm, the authors keep running the evolutive procedure as long as there are not incoming samples. Here, the evolutive update is called once every so often. See my point? How do these two strategies compare? For example, in the paper, the authors assume a delay of one second per sample in some cases. In my understanding, multiple iterations of the evolutive update could be performed depending on the task at hand (and the machine, of course). 

Would River's evoStream have a lower clustering performance when compared with the original proposal? Imagine, as an example, that we set the evolution interval to 1 sample (which is the minimum we can do) and the original proposal runs 10 times between each arriving instance.

I'll study the evoStream paper in-depth to better understand this situation. Any thoughts @hoanganhngo610 and @jacobmontiel?

river/cluster/evostream.py
hoanganhngo610: Resolved!

river/cluster/evostream.py
hoanganhngo610: Resolved!

river/cluster/evostream.py
hoanganhngo610: Suggestion added!

river/cluster/evostream.py
hoanganhngo610: Resolved!

river/cluster/evostream.py
hoanganhngo610: Yes and no! Indeed, in the paper, it is written that, in case `g_i = 0`, it is set to `2 * delta`. However, when you look at part `3.2.6` of the other paper written in 2000 (the original paper on genetics algorithm), in this case, it is set to either `2 * delta* or *-2 * delta`. In order to maintain randomness in the algorithm, I decided to adopt this version of this part, making it a little bit different from the algorithm in the paper.

river/cluster/evostream.py
hoanganhngo610: Mentioned! :D

river/cluster/evostream.py
hoanganhngo610: You are right, the cut-off point is also randomly defined. Changed to `cutoff = random.randint(dim - 1)`.

river/cluster/evostream.py
hoanganhngo610: In case there is no opposition to this, I would leave a note here so that we can close this later! @jacobmontiel @smastelini

river/cluster/evostream.py
hoanganhngo610: Try/except block added for a ZeroDivisionError!

river/cluster/evostream.py
hoanganhngo610: @staticmethod used. I thought it would have been better to leave the function within the original `_evolution` function, but now thinking back, pulling it outside as a static method would leave the code look nicer!

river/cluster/evostream.py
hoanganhngo610: `del` used (honestly I did not pay attention to this at all while coding)!

river/cluster/evostream.py
hoanganhngo610: Usually, for microcluster structures of all online clustering algorithms, we always link this to `ABCMeta`. Maybe @jacobmontiel can you provide a better explanation on this?

river/cluster/evostream.py
hoanganhngo610: Unfortunately, the paper only says simply the "roulette wheel selection" is used, without explicitly saying what it is. However, upon looking it up, the roulette wheel selection is simply the fitness proportionate selection, whose algorithm can be found (here)[https://en.wikipedia.org/wiki/Fitness_proportionate_selection]. 
Indeed, the roulette wheel selection implemented here does not return one, but *two* independent key chosen by randomizing two independent numbers.

river/cluster/evostream.py
hoanganhngo610: The option of passing a `seed` and the creation of random generators are used! However, this would make all static methods return to normal functions, since all requires the use of a generator whose seed is passed from the user.

river/cluster/evostream.py
smastelini: Let's hear @jacobmontiel's opinion. I'm cool with that.

MaxHalford:
Closing this for tidyness, feel free to reopen if work continues!

PR #539, opened 2021-03-24
Fix issue with KNNClassifier

tachyonicClock:
Classification probabilities were not being wrapped with the softmax
function.


MaxHalford:
Hey, thanks for the PR! What makes you think that the softmax function should be used? It won't necessarily calibrate the predictions. As you can see in [scikit-learn's source code](https://github.com/scikit-learn/scikit-learn/blob/95119c13a/sklearn/neighbors/_classification.py#L606), they don't use the softmax function.

tachyonicClock:
You are correct. Sorry about that I got a little confused about what the issue was. It turns out it was to do with scipy.spatial.cKDTree.query which returns a 1D array when k=1

MaxHalford:
Nice catch!

PR #545, opened 2021-03-29
Pipeline new param

MaxHalford:
I added a `learn_unsupervised` parameter to `Pipeline.learn_one` and `Pipeline.learn_many` following the discussion [here](https://github.com/online-ml/river/discussions/542). It's not the first time users ask about this so I'm convinced it's ok. In any case, the default behavior is still the same. Note that we also have a `learn_unsupervised` parameter for `predict_*`, so it makes sense to add its counterpart.

I also corrected a little bug for `HardSampling*`. I added a test so no need to review in depth.

I also removed the table of contents from the README because GitHub has added their own ToC feature at the top of each README :)


river/imblearn/test_hard_sampling.py
```diff
@@ -0,0 +1,19 @@
...
+        preprocessing.StandardScaler() |
+        imblearn.HardSamplingClassifier(
+            classifier=linear_model.LogisticRegression(),
+            p=0.1,
+            size=40,
+            seed=42,
+        )
+    )
+
+    pickle.dumps(model)
```
smastelini: Is this new test necessary? Previously the pickling test was skipped for this class, right?

Shouldn't be the case to enable this test?

smastelini (changes requested):
Thanks for the changes, @MaxHalford. Just some questions:

- Could you update the PR description to also reflect the changes in the README?
- Why did some docstring tests change? (e.g., hst)

river/imblearn/test_hard_sampling.py
MaxHalford: Damn you're right haha. Done!

MaxHalford:
So the docstrings for HST changed because the pipeline code for `score_one` wasn't exactly correct. I haven't spent too much looking at why it wasn't ok before but now the logic seems fine.

river/imblearn/hard_sampling.py
```diff
@@ -36,9 +29,17 @@ def _wrapped_model(self):
     def predict_one(self, x):
         return self.model.predict_one(x)
 
+    @property
+    def _model_pred_func(self) -> typing.Callable:
+        if isinstance(self.model, base.Classifier):
+            if not self.model._multiclass:
+                return lambda x: self.model.predict_proba_one(x)[True]
```
smastelini, line 36: Cool trick!

smastelini (approved):
This PR reflects plenty of requests from river users. I'm happy to acknowledge this project becoming a useful tool for researchers and practitioners alike.

Let's merge it.

PR #552, opened 2021-04-01
New tree interface

MaxHalford:
Here you go @smastelini! I'm done with the tree structure. I haven't done the tree visualisation part yet. I think it will be easier for me to do it once you've updated your trees to use this structure. What do you think?


jacobmontiel (approved):
The generic approach looks good to me. The example usage on HST is promising. I vote to keep this as a template and work incrementally to standardize everything.

Thanks for your understanding of the delay in my review.

PR #558, opened 2021-04-08
Improve tree coverage and fix bugs in HAT{C, R} and EFDT

smastelini:
Bugs fixed:

- Make EFDT update all nodes in the path from the root to a leaf
- Block HATC and HATR from trying to split inactive nodes


MaxHalford (approved):
LGTM! The failing test is again that OpenML one so no worries.

MaxHalford:
87% :)

PR #568, opened 2021-04-19
Tree refactor

smastelini:
Refactor and streamline decision trees to cope with the new base classes. Since most of the altered inner structures are private, they were not extensively documented.

**Fixed:**
- Bug in the nominal splitter for classification
- Bug in HATC/HATR (related to node deactivation)
- Bug in EFDT (concerning split re-evaluation)

**Enforce:**
- The policy to deal with missing features and emerging categories (in nominal features). Make sure all the trees deal with such situations consistently.

**TODO:**

- ~Fix tests related to nominal features, to reflect the bug fix~
- ~Finish porting the tree visualization/inspection tools: `debug_one` and `draw`~
- ~Update jupyter notebook that explores the tree module~

**Verify:**
- Should we rename?
    * `model_description() -> description` (**Update**: method removed in favor of `to_dataframe`)
    * `model_measurements` -> ~measurements~ `summary`


smastelini:
@MaxHalford suggested removing the property `measurements` (from `tree.HoeffdingTree`) by splitting it into multiple properties. If we decide to follow this path (pun intended), here some candidates for addition:

- `n_nodes`: the total number of nodes (including branches and leaves)
- `n_branches`: the number of decision nodes
- `n_leaves`: the total number of leaves (both active and inactive)
- `n_active_leaves`
- `n_inactive_leaves`

Any other suggestions, @jacobmontiel?

river/tree/base.py
```diff
@@ -28,14 +28,29 @@ def __init__(self, *children):
     def next(self, x) -> Union["Branch", "Leaf"]:
         """Move to the next node down the tree."""
 
-    def walk(self, x) -> Iterable[Union["Branch", "Leaf"]]:
+    @abc.abstractmethod
+    def most_common_path(self) -> Tuple[int, Union["Leaf", "Branch"]]:
```
MaxHalford, line 32: This shouldn't be here. The most common path logic should be specific to each, and should not be part of the `Tree` interface.

river/tree/base.py
```diff
@@ -28,14 +28,29 @@ def __init__(self, *children):
...
-        yield from self.next(x).walk(x)
-
-    def traverse(self, x) -> "Leaf":
+        try:
+            yield from self.next(x).walk(x, until_leaf)
+        except KeyError:
+            if until_leaf:
+                _, node = self.most_common_path()
+                yield node
+                yield from node.walk(x, until_leaf)
```
MaxHalford, line 49: I'm not clear on what this is doing (then again, it's late). Can you remind me why you need to introduce `until_leaf`? Isn't this specific to a particular tree?

jacobmontiel:
`n_branches` sounds weird, what about `n_splitters`?
Also worth considering: `depth` and `size (bytes)`

river/tree/nodes/leaf.py
```diff
@@ -0,0 +1,190 @@
...
+import copy
+import numbers
+import typing
+
+from ..base import Leaf
+from ..utils import BranchFactory
+
+
+class HTLeaf(Leaf, metaclass=abc.ABCMeta):
+    """Base leaf class to used in Hoeffding Trees.
```
jacobmontiel, line 11: "...to be used..."

river/tree/splitter/nominal_splitter_classif.py
```diff
@@ -23,72 +29,61 @@ def update(self, att_val, target_val, sample_weight):
...
-                )
-            try:
-                self._att_val_dist_per_class[target_val][att_val] += sample_weight
-            except KeyError:
-                self._att_val_dist_per_class[target_val][att_val] = sample_weight
-                self._att_val_dist_per_class[target_val] = dict(
-                    sorted(self._att_val_dist_per_class[target_val].items())
-                )
+            self._att_values.add(att_val)
+            self._att_dist_per_class[target_val][att_val] += sample_weight
```
jacobmontiel, line 33: In the current implementation, we have a sorting step that I am not confident is covered with default dictionaries.

river/tree/splitter/nominal_splitter_classif.py
```diff
@@ -23,72 +29,61 @@ def update(self, att_val, target_val, sample_weight):
...
-            branch_mapping = {
-                attr_val: branch_id for branch_id, attr_val in enumerate(att_values)
-            }
-            best_suggestion = SplitSuggestion(
-                NominalMultiwayTest(att_idx, branch_mapping), post_split_dist, merit,
+
+            best_suggestion = BranchFactory(
+                merit,
+                att_idx,
+                sorted(self._att_values),
```
jacobmontiel, line 57: Perhaps this answers my previous comment, but better ask.

river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -174,68 +177,50 @@ def learn_one(self, x, y, *, sample_weight=1.0):
...
-                    if leaf_node is None:
-                        leaf_node = fn.parent
-                    dist = leaf_node.leaf_prediction(x, tree=self)
-                    # Option Tree prediction (of sorts): combine the response of all leaves reached
-                    # by the instance
-                    proba = add_dict_values(proba, dist, inplace=True)
+        if self._root is not None:
+            found_nodes = [self._root]
+            if isinstance(self._root, HTBranch):
+                found_nodes = self._root.traverse(x, until_leaf=True)
```
jacobmontiel, line 183: How is the skipping of alternate tree roots enforced?

jacobmontiel (changes requested):
Awesome work @smastelini 

This is quite a big PR. I am happy with the redundant code removed and the standardization of tree maintenance tasks.

I left a couple of comments. Please let me know if you need any help.

river/tree/base.py
smastelini: I disagree :P

Every branch calculates the most common path differently. That's why I created the abstract method. This is the core functionality to deal with missing split features.

I think this is generic enough to belong to the base class. Do you have other suggestions?

river/tree/base.py
smastelini: Nope, it's not! We follow some conventions to deal with missing split features. When learning or predicting we pass the instance to the most traversed path.

Ok, so in these cases, we always need to reach a leaf. But, there are situations where we have categorical features (or discretized numerical features) and a new category arrives. In these cases, we only pass the instance to the most traversed path while predicting (because the branch does not know what to do with the new category). However, while learning, we need to make sure we create a new path corresponding to this new category.

Hence, depending on the situation, we allow instances to stop at a branch, rather than always reaching a leaf. That's what the `until_leaf` param is designed to control. All the decision trees make calls to `walk` and `traverse` using different values for this parameter. Anyway, its default behavior is to always reach a tree.

smastelini:
> n_branches sounds weird, what about n_splitters?
Also worth considering: depth and size (bytes)

This is worth discussing. However, we should proceed with caution, as all trees are going to be affected. So, here are our contenders:

- depth x height
- branch x split

Let's vote :D

river/tree/extremely_fast_decision_tree.py
```diff
@@ -136,23 +138,26 @@ def _new_learning_node(self, initial_stats=None, parent=None):
...
-            return EFDTLearningNodeNB(initial_stats, depth, self.splitter)
+            return EFDTLeafNaiveBayes(initial_stats, depth, self.splitter)
         else:  # NAIVE BAYES ADAPTIVE (default)
-            return EFDTLearningNodeNBA(initial_stats, depth, self.splitter)
+            return EFDTLeafNaiveBayesAdaptive(initial_stats, depth, self.splitter)
 
-    def _new_split_node(
-        self, split_test, target_stats=None, depth=0, existing_splitters=None
-    ):
+    def _branch_selector(
```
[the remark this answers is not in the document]
smastelini: I will change that once we decide upon the final convention. It could be either `_new_splitter` (but `splitter` is also used to refer to the attribute observers, so maybe it's better to use `_new_split`) or `_new_branch`.

river/tree/extremely_fast_decision_tree.py
```diff
@@ -185,17 +190,62 @@ def learn_one(self, x, y, *, sample_weight=1.0):
...
+        1. Find the node where instance should be.
+        2. If no node have been found, create new learning node.
+        3.1 Update the node with the provided instance
+
+        Parameters
+        ----------
+        x
+            Instance attributes.
+        y
+            The instance label.
```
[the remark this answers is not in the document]
smastelini: This docstring is specific to this classifier. We could change it to target, but it sounds like overkill to me. What do you think?

river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -174,68 +177,50 @@ def learn_one(self, x, y, *, sample_weight=1.0):
...
-
-                for i in range(split_node.n_children):
-                    self.__find_learning_nodes(
-                        split_node.get_child(i), split_node, i, found
-                    )
-                if split_node._alternate_tree is not None:  # noqa
-                    self.__find_learning_nodes(
-                        split_node._alternate_tree, split_node, -999, found  # noqa
-                    )
+    def _branch_selector(
```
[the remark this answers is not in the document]
smastelini: Due to the new base classes, this function does not actually build the branch/split node. It only selects the correct type. The real magic happens in the `BranchFactory` objects, which are returned by the splitters. Basically, the factory assembles the branch node which was picked by `_branch_selector`.

river/tree/hoeffding_tree_classifier.py
```diff
@@ -102,15 +95,18 @@ class HoeffdingTreeClassifier(HoeffdingTree, base.Classifier):
     >>> metric = metrics.Accuracy()
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
-    Accuracy: 86.09%
+    Accuracy: 83.78%
```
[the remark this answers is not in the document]
smastelini: Exactly! In this specific dataset the performance dropped a bit, but that's not always the case :)

river/tree/splitter/nominal_splitter_classif.py
smastelini: Since all the access operations are using the keys, I believe we can avoid this extra sorting step. But I'll check it again, just to be really sure :)

river/tree/splitter/nominal_splitter_classif.py
smastelini: I think so! The access is made by key, but I use the ordered values of the categories to ensure the access is always performed consistently.

river/tree/hoeffding_adaptive_tree_classifier.py
smastelini: You indeed have sharp eyes! Well, first some background:

The "option votes" are not part of the original HATC. In fact, a [recent paper](https://arxiv.org/abs/2010.08199) discusses that this and other changes were applied to the code in MOA, but they are not documented anywhere. In summary, "option votes" are not originally part of the theoretical HATC algorithm.

Some time ago, after trying to understand the HATC code for the zillionth time, I decided to add a comment concerning my understanding of the code. I did that because I kept forgetting what that and other portions did. So, _my understading_ is that alternate trees composed of a single leaf don't vote because they are "too fresh". But again, this is my interpretation of the code.

But well, if we are talking about unspecified behaviors, in my humble opinion, if concept drift was detected at a specific decision node, that poor new alternate tree might do a better job than the current subtree since it is being trained using only instances belonging to the new concept.

So, now we have a war of unspecified behaviors XD. In fact, I intend to add a parameter to switch on/off the "option votes" in the future. For simplicity, for now, I allowed every alternate tree to vote in this refactoring. Again, I mostly did that because of my personal bias regarding tree learning, but also because filtering alternate trees composed of a single leaf was a tricky endeavor.

What do you think? If needed, I can try to mimic the behavior in MOA. But if I'm not wrong, the mentioned paper discusses that sometimes the "option vote" is not beneficial. That case was also mentioned in #532.

In any case, my bad! I forgot to document that. It was lost among the other changes.

river/tree/extremely_fast_decision_tree.py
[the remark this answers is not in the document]
smastelini: My bad, `_branch_selector` does a different job than the previous `_new_split_node`. I got confused :P

I discuss the functionality of this method in another comment.

river/tree/hoeffding_adaptive_tree_classifier.py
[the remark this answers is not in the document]
smastelini: To be more specific: the new tree base classes require that all the descendant nodes are passed as parameters when building the branch (decision) nodes. That was not the case previously. We used to rely on the `set_child` method, which was called inside a loop after the (RIP) `SplitNode` was created. Now, we select the branch node type (numerical and binary, numerical and multiway, nominal and binary, or nominal and multiway), build its descendants, select any other needed initialization parameters and pass everything to the `assemble` method of `BranchFactory`.

`BranchFactory` instances are built by the attribute splitters. They carry all the basic information to build a decision node: the split attribute, threshold (or any other needed information), the children's stats, and so on. But the tree splitters "don't know" which kind of decision node the trees might come up with. For example, most of the decision trees rely on the vanilla branch nodes, but EFDT, HAT{C/R} implement some additional fancy logic in their decision nodes. Therefore, the `BranchFactory` instances take everything they can from the tree splitters; the remaining information is obtained from the actual decision trees. The `assemble` call in the `_attempt_to_split` method joins everything and ensures the decision nodes are properly created.

This whole factory thing was what I was able create to solve the problem with the decision nodes. It took me while to figure out. In fact, I spent most of the time during this refactoring creating and polishing the `BranchFactory` class and making sure it could handle every fancy/alien branch nodes I might throw at it.

MaxHalford:
The depth of a tree doesn't exist. Depth is a notion that is tied to a node. The depth of a node is the distance from the node to the root. The height of a tree is the maximum distance from the root to any of the successors.

Semantically, I would say that a split is a part of a branch. The leaf/branch terminology is more common. But I'm happy to be challenged on that. 

With regards to memory usage, can't you reuse the memory_usage property from base.Base? Each tree now inherits from it.

Cheers!

Sent from my iPhone.

river/tree/extremely_fast_decision_tree.py
```diff
@@ -229,17 +279,22 @@ def _process_nodes(self, x, y, sample_weight, node, parent, branch_index):
 
             if (new_weight - old_weight) >= self.min_samples_reevaluate:
                 # Reevaluate the best split
-                stop_flag = self._reevaluate_best_split(node, parent, branch_index)
+                stop_flag = self._reevaluate_best_split(
+                    node,
+                    parent,
+                    branch_index,
+                    splitter=self.splitter,
```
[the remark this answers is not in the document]
smastelini: I added a comment to clarify their purpose. They could be passed as parameters, rather than kwargs. However, future generations might want to create variations of EFDT, so this gives us flexibility.

river/tree/extremely_fast_decision_tree.py
```diff
@@ -185,17 +190,62 @@ def learn_one(self, x, y, *, sample_weight=1.0):
...
 
         # Sort instance X into a leaf
         self._sort_instance_into_leaf(x, y, sample_weight)
         # Process all nodes, starting from root to the leaf where the instance x belongs.
-        self._process_nodes(x, y, sample_weight, self._tree_root, None, -1)
+        self._process_nodes(x, y, sample_weight, self._root, None, None)
 
         return self
 
+    def _sort_instance_into_leaf(self, x, y, sample_weight):
```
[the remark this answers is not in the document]
smastelini: Renamed in commit [e609435](https://github.com/online-ml/river/pull/568/commits/e609435fed88979341e1f8e245e7cdfeda8230fb).

river/tree/extremely_fast_decision_tree.py
```diff
@@ -185,17 +190,62 @@ def learn_one(self, x, y, *, sample_weight=1.0):
...
+        """Sort an instance into a leaf.
+
+        Private function where leaf learn from instance
```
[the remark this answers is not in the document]
smastelini: Reworded in commit [e609435](https://github.com/online-ml/river/pull/568/commits/e609435fed88979341e1f8e245e7cdfeda8230fb).

river/tree/nodes/leaf.py
smastelini: Fixed in commit [e609435](https://github.com/online-ml/river/pull/568/commits/e609435fed88979341e1f8e245e7cdfeda8230fb). Thanks for pointing that out!

smastelini:
> The depth of a tree doesn't exist. Depth is a notion that is tied to a node. The depth of a node is the distance from the node to the root. The height of a tree is the maximum distance from the root to any of the successors.

You're right, @MaxHalford. This [page](https://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height) mentions exactly your point.

> Semantically, I would say that a split is a part of a branch. The leaf/branch terminology is more common. But I'm happy to be challenged on that.

I'm still not entirely convinced about the branch terminology. To me, "branch" is a synonym for vertex or edge (from graph theory). I do understand that we are talking about branch(ing) nodes, so I might be wrong. I'll research a bit more regarding that.

> With regards to memory usage, can't you reuse the memory_usage property from base.Base? Each tree now inherits from it.

For sure! Just as a heads up, we might want to use an iterative implementation for this one, as skmultiflow did. Recently, while benchmarking tree regressors (+ the binary search tree splitter) the `memory_usage` calls reached the recursion limit. I was using a large dataset :)
Let's put that on our TODO list.

jacobmontiel:
> > The depth of a tree doesn't exist. Depth is a notion that is tied to a node. The depth of a node is the distance from the node to the root. The height of a tree is the maximum distance from the root to any of the successors.

Agree, I also vote to go with `height`.

> > Semantically, I would say that a split is a part of a branch. The leaf/branch terminology is more common. But I'm happy to be challenged on that.
> 
> I'm still not entirely convinced about the branch terminology. To me, "branch" is a synonym for vertex or edge (from graph theory). I do understand that we are talking about branch(ing) nodes, so I might be wrong. I'll research a bit more regarding that.

Same feeling as Saulo here. Although the branch concept is a core element of the tree structure, at the implementation level we deal with nodes ("leaf" and "internal/split"). I prefer the consistency of nodes.

> > With regards to memory usage, can't you reuse the memory_usage property from base.Base? Each tree now inherits from it.

Definitely, the idea is **only** to collect the information. This method should not implement any property.

river/tree/extremely_fast_decision_tree.py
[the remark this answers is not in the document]
jacobmontiel: My bad, you are right. Let's keep it as is.

river/tree/hoeffding_tree_classifier.py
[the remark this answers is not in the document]
jacobmontiel: OK. Thanks for the clarification.

river/tree/hoeffding_adaptive_tree_classifier.py
jacobmontiel: This is indeed a tricky situation. I can ask the author of HATC ;-) for his opinion.

smastelini: That would be awesome!

river/tree/nodes/branch.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.stats = stats
+        self.__dict__.update(attributes)
+
+    @property
+    def total_weight(self):
+        return sum(child.total_weight for child in filter(None, self.children))
+
+    @abc.abstractmethod
+    def branch_no(self, x):
```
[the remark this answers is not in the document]
smastelini: I don't have exactly strong feelings for `branch_no` ("branch number"), but I do like that is compact, simple haha. If we wanna be clear about the operation, `branch_for_instance` might be the way to go

river/tree/base.py
MaxHalford: Ok very clear! How about just handling the "go to the most branch child" in the `next` function? It's what I've been doing for half-space trees

river/tree/base.py
MaxHalford: Right, but as we discussed elsewhere (I can't remember haha), can't you just handle the new leaf creation in the `next` function?

river/tree/base.py
smastelini: The thing is, sometimes we need to do that "manually". Let me give you another example. Suppose two cases while learning:

1. We receive an instance where a split feature is missing
2. We receive an instance where a new category emerges in one of the split features

In the first case, the tree must pick the most traversed path and return the reached leaf. In the second one, the tree must stop the traversal in the internal node and create a new branch (and leaf) to accommodate the new category. Each tree has specific sets of leaves (that also depend on the prediction strategy selected by the user), that require specific sets of parameters to be constructed. That's why they are usually built by the "main tree" rather than by branches.

Therefore, I decided to expose the `most_common_path`, so that the trees could wisely use it. In summary, it's not only a matter of taking the "main road", but also deciding when to travel :D

MaxHalford: Mmm I see! But you could also argue that the leaf create could happen in the `next`, and the branch could just have a point to the main tree, no? I'm just trying to weigh all possibilities.

river/tree/base.py
smastelini: Please check my previous comment for more info. But in summary, it's not that simple. The trees create different types of leaves depending on the user's choice. Some of them are simple, like those that output the majority class or the target mean. Others perform fancy operations. Usually, to build a new leaf, we need information from the main tree structure (e.g., leaf type, prediction model, seed, etc.) and the previous leaf to be promoted to an internal node. We could init the branches with the needed parameters from the main tree, but that doesn't sound much future proof nor easy to document. For instance, this also doesn't solve the situation where we copy the linear model from the leaf to be split to all its children.

river/tree/base.py
smastelini: It could do the job, perhaps. I need to think a little bit about the weirdest trees though. Anyhow, it sounds to me that we will end up delegating too much power to a single method, haha. Besides, all new tree implementations would need to account for this and possible other corner cases in `next`. And yet, `next` would need to know when to take one type of action or another.

Currently, `next` tries to obtain the next node in the path, if it's not possible an error is thrown. Then `walk` decides what to do, depending on `until_leaf` (by the way, any ideas for a better name are welcome). The walk can either stop at the branch or continue by following the most traversed path.

I also do not appreciate very much having to implement `most_common_path` every time. But it still sounds like a good compromise. (Kind of) quoting Jacob: otherwise, we might end up complicating even more the maze that is the trees.

smastelini:
As an update, we decide to keep the Leaf/Branch naming convention. Hence, we maximize the gardening references.

Jokes aside, there are still a few items to cross off from our list. Once @jacobmontiel is satisfied with the changes we merge this PR :D

river/tree/hoeffding_tree.py
```diff
@@ -361,60 +333,30 @@ def debug_one(self, x: dict) -> typing.Union[str, None]:
         -------
             A representation of the path followed by the tree to predict `x`; `None` if
             the tree is empty.
+
+        Notes
+        -----
+        Currently, Label Combination Hoeffding Tree Classifier (for multi-label
```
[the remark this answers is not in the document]
smastelini: I made both `debug_one` and `draw` return `None` in `LabelCombinationHoeffdingTreeClassifier`, as well as also made generating a warning (please, check commit [3a7a2a4](https://github.com/online-ml/river/pull/568/commits/3a7a2a42c01f89b2e0cbcdd6e940bb5cefed832c)).

Initially, I thought about simply raising a `NotImplementedError`, but this action is discouraged in the [exception page](https://docs.python.org/3/library/exceptions.html#NotImplementedError) itself:

> **Note:** It should not be used to indicate that an operator or method is not meant to be supported at all – in that case either leave the operator / method undefined or, if a subclass, set it to `None`.

river/tree/nodes/branch.py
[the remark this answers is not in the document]
jacobmontiel: I am just concerned that "no" could be misinterpreted.

river/tree/hoeffding_tree.py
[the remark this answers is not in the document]
jacobmontiel: Excellent. Thanks!

river/tree/hoeffding_adaptive_tree_classifier.py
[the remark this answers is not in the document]
jacobmontiel: Thanks for the clarification! Just as a comment, we could use `_branch_type_selector` to avoid confusion. Not an actual change request ;-). I leave it up to you.

river/tree/hoeffding_adaptive_tree_classifier.py
jacobmontiel: This falls outside of the scope for this PR so I am not going to block its merge based on it. If I get new information I will update the comment.

jacobmontiel (approved):

PR #570, opened 2021-04-25
Tree improvements

MaxHalford:
I actually didn't manage to find enough time to decently work on tree viz. I did a couple of light improvements. In particular, I added the `to_dataframe` method, which I think we should use in favor of `model_description`.

The way I see it, we have 3 options for tree viz:

- Keep using graphviz: it's ugly but it works.
- Write an HTML viz. It could be pretty and interactive but I need to spend time on it.
- Integrate with [dtreeviz](https://github.com/parrt/dtreeviz), which looks good but is a bit akward because you need to pass around the data along with the model.

### DIFF
--- a/river/anomaly/hst.py
+++ b/river/anomaly/hst.py
@@ -10,7 +10,6 @@ __all__ = ["HalfSpaceTrees"]
 
 
 class HSTBranch(Branch):
-
     def __init__(self, left, right, feature, threshold, l_mass, r_mass):
         super().__init__(left, right)
         self.feature = feature
--- a/river/anomaly/hst.py
+++ b/river/anomaly/hst.py
@@ -11,14 +11,20 @@ __all__ = ["HalfSpaceTrees"]
 
 class HSTBranch(Branch):
 
-    __slots__ = ['feature', 'threshold', 'l_mass', 'r_mass']
-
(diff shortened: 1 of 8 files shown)


smastelini:
Hi @MaxHalford,

Thanks for the changes!

> which I think we should use in favor of model_description.

I agree. It's better to generate a structured output so that the user can use it as they want. Regarding tree viz, I vote to keep the graphviz-based solution for now. We can think about better solutions down the road (option 2 sounds promising).

I think this PR is ready. I like the new scheme where leaves yield themselves. That avoids multiple `isinstance` calls. I'll also change #568 to reflect this new strategy.

By the way, we need to define a method to describe a branch, given its index (for `debug_one` and `draw`). Any suggestions for a name? I can create that on #568, but it might fit the scope of this PR better.

smastelini (approved):

smastelini:
I'll also update the notebook containing the Decision Tree examples, to replace `model_description` by `to_dataframe` (in #568).

MaxHalford:
> By the way, we need to define a method to describe a branch, given its index (for debug_one and draw). Any suggestions for a name? I can create that on #568, but it might fit the scope of this PR better.

That popped up in my mind too. I think `repr_split` fits the bill.

PR #597, opened 2021-05-20
Update batch-to-online.ipynb

joshuagornall:
# Simple spelling error on line 20 - (after 3rd sentence)

This sentence: Like many other models based on numerical weights, ~~logisitc~~ regression is sensitive to the scale of the features.
```diff
- logisitc
+ logistic
```


smastelini (approved):
Thanks for the fix @joshuagornall!

PR #622, opened 2021-06-29
Index fix for CluStream and DBStream

hoanganhngo610:
This PR serves to handle `KeyError` cases at the beginning of each algorithm. 

Whenever the algorithm returns this error when `predict_one` is called, the index is automatically set to `0`, which comes from the assumption that all outlier points are kept in the same cluster.


jacobmontiel (approved):
The fixes are correct. Thanks @hoanganhngo610

PR #644, opened 2021-07-26
updated ewa.py to accommodate time series models

mathco-wf:
Updated the file to enable users to create expert of time series models like SNARIMAX

### DIFF
--- a/river/expert/ewa.py
+++ b/river/expert/ewa.py
@@ -118,7 +118,7 @@ class EWARegressor(base.EnsembleMixin, base.Regressor):
         total = 0
         for i, regressor in enumerate(self):
             #if the regressor is an instance of SNARIMAX
-            if isinstance(regressor, river.time_series.snarimax.SNARIMAX):
+            if 'forecast' in dir(regressor):
                 y_pred = regressor.forecast(horizon=horizon,xs=[x])
                 y_pred = y_pred[0]
             else:
@@ -128,7 +128,7 @@ class EWARegressor(base.EnsembleMixin, base.Regressor):
             self.weights[i] *= math.exp(-self.learning_rate * loss)
             total += self.weights[i]
             #if the regressor is an instance of SNARIMAX
-            if isinstance(regressor, river.time_series.snarimax.SNARIMAX):
+            if 'forecast' in dir(regressor):
                 regressor.learn_one(y,x)
             else:
                 regressor.learn_one(x, y)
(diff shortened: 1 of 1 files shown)


mathco-wf:
This update is provided in reference to the discussion https://github.com/online-ml/river/discussions/641

raphaelsty:
Hi @mathco-wf,

There are two points on which I disagree with your pull request.

```python
 if isinstance(regressor, river.time_series.snarimax.SNARIMAX):
```
Incorporating a model specific logic is not a good idea in an independent and external class IMO. If we modify `snarimax` in the next few weeks, we may also have to modify `river.expert.ewa`. If we add a model to the `river.time_series` module that has the forecast method, we will have to modify `river.expert.ewa` again to integrate the new model. I think we can find a more elegant solution to integrate models that have a `forecast` method. 

Also, you choose to optimize `river.expert.ewa` only on the first prediction (horizon = 1) of `river.time_series.SNARIMAX` model. I don't know if this is ideal or the right way to do it. The EWARegressor model would mechanically give more weight to a model that is good on the near future than in the long term. It would be great to find existing works on this topic. 

I would like to have @MaxHalford opinion, we could then update the pull request together @mathco-wf . 😊

mathco-wf:
Hi @raphaelsty,

Do you think this condition will work:
`if 'forecast' in dir(regressor):`

regarding the horizon=1, We tried to give it as a default value but we would also like to know the views of @MaxHalford on this and will update the code accordingly.

AdilZouitine:
In my opinion, the easiest solution is to make a `SnarimaxToRegressor` wrapper.
I don't think this wrapper should be in the library (because it's a niche use case).
But it would have its place in a future `river-contrib` :smile:

raphaelsty:
This is what a wrapper could look like:

```python
from river import base, time_series

class SnarimaxRegressor(time_series.SNARIMAX, base.Regressor):
    """Snarimax as a regressor for experts models."""
    def __init__(
        self,
        p: int,
        d: int,
        q: int,
        m: int = 1,
        sp: int = 0,
        sd: int = 0,
        sq: int = 0,
        regressor: base.Regressor = None,
    ):
        super().__init__(
            p=p,
            d=d,
            q=q,
            m=m,
            sp=sp,
            sd=sd,
            sq=sq,
            regressor=regressor,
        )

    def predict_one(self, x):
        """Regressor predict_one method."""
        return super().forecast(horizon=1, xs=[x])[0]

    def learn_one(self, x, y):
        """Override learn_one method to avoid switching x and y."""
        return super().learn_one(y, x=x)

    
from river import expert

model = expert.EWARegressor(
    regressors = [
        SnarimaxRegressor(p=17, d=0, q=1), 
        SnarimaxRegressor(p=16, d=0, q=1), 
        SnarimaxRegressor(p=18, d=0, q=1),
        SnarimaxRegressor(p=19, d=0, q=1)]
)

# Creating toy time series:
for _ in range(30):
    for y in range(20):
        x = {}
        model.learn_one(x=x, y=y)
        
model.predict_one(x={})   
# >>> 2.9623816579144098

model.regressors[2].forecast(10)
# >>> [11.8, 7.25, 3.94, 2.0, 1.04, 0.85, 1.26, 2.14, 3.38, 4.89]
```

mathco-wf:
@raphaelsty this looks good, this works. Will this be available in the current version of the library?

MaxHalford:
For me it makes little sense to have a time series model that is designed to forecast one step ahead. That's just a regression model. In other words, adding a `predict_one` method to a model in the `time_series` module is just wrong by design.

In any case, we might be delegating our `time_series` module to the [`sktime` project](https://github.com/alan-turing-institute/sktime), but 🤫

Closing this.

PR #458, opened 2021-01-29
Stochastic Gradient Trees

smastelini:
[Stochastic Gradient Trees (SGT)](https://arxiv.org/abs/1901.07777) are a family of general-purpose decision tree models designed to learn incrementally. Unlike Hoeffding Trees, which learn by minimizing heuristics related to the target impurity (classification) or spread (regression), SGTs directly minimize a loss function to guide tree-growth/update. Hence, one can design different loss functions for different prediction tasks, as long as they are twice differentiable. This characteristic also enables SGTs to be used as processing units of neural networks.

This PR brings the core components of SGTs to `river`. The main author of the SGTs, @henrygouk, kindly agreed to review and improve this PR. Thanks for that, Henry. The SGTs implemented here work slightly differently from the original version available in MOA. Rather than using a fixed buffer to define split points for numerical features, the current implementation uses an adapted version of the upcoming [Quantization Observer (QO)](https://arxiv.org/abs/2012.00083). Thus, they do not need an initial "warm-up" period.

#### EDIT: The QO paper was accepted for publication in Pattern Recognition Letters. The manuscript should be online in the next weeks. :D Soon, I will make QO available for Hoeffding Tree Regressors, as it was originally designed to work with.

Here's the summary of the changes:

- Add SGTs for single-target classification and regression
- Add new folders to the tree submodule to accommodate the inner structures of the SGTs

TODO:

- [x] Improve documentation
- [ ] Perform in-depth testing
- [x] Refactor the code regarding naming convention (if needed). For instance: Streaming/Stochastic Gradient Tree
- [x] Improve the placement of files to reflect the new folder structure (merge the `_tree_utils.py` file into the `_utils` folder)
- [x] ~Add other loss functions and~ make the loss function selection a configurable tree parameter?

Moreover, we could add (now or in the future) the needed tools to build neural networks using SGTs as building blocks.

PR #674, opened 2021-08-20
Add Stochastic Gradient Trees (now for real)

smastelini:
This PR takes upon the work done in #458.

**Summary**

- SGTs now follow the new base classes for decision trees.
- Fixed bug where `repr_branch` and `repr_split` were referred as the same operation
- HTBranch -> DTBranch: this structure is generic enough to be used by both HTs and SGTs
- Created feature quantizers
    - The DynamicQuantizer is inspired by `tree.splitter.QOSplitter`
    - The StaticQuantizer is inspired by the original implementation made by @henrygouk 
- Housekeeping: files renamed for aesthetic purpose


river/tree/__init__.py
```diff
@@ -39,6 +39,10 @@
 * Define properties to access leaf prediction strategies, split criteria, and other
 relevant characteristics.
 
+**2. Stochastic Gradient Trees**
+
+TODO
```
smastelini, line 44: I need to do that :P

@henrygouk, please feel free to give us any suggestions to introduce the SGTs :D

smastelini, line 44: I need to do that :P

@henrygouk, please feel free to give us any suggestions to introduce the SGTs :D

smastelini:
I will run some benchmarks comparing SGTs against the HTs.

river/rules/amrules.py
```diff
@@ -7,7 +7,7 @@
 from river import base, drift, linear_model, stats, tree
 
 from ..tree.split_criterion import VarianceRatioSplitCriterion
-from ..tree.splitter.base_splitter import Splitter
+from ..tree.splitter.base import Splitter
 from ..tree.splitter.nominal_splitter_reg import NominalSplitterReg
```
MaxHalford, line 11: Meta: at Alan we've decided to not do any relative imports. Here we would do:

```py
from river.tree.split_criterion import VarianceRatioSplitCriterion
```

At some point we can decide if we want to do that too!

river/tree/losses.py
```diff
@@ -0,0 +1,52 @@
+import math
+from abc import ABCMeta
+
+from .utils import GradHess
+
+
+class Loss(metaclass=ABCMeta):
```
MaxHalford, line 7: I think inheriting from `abc.ABC` is the more modern way, but not sure.

Other question: is there a reason why you can't use the `Loss` class from `optim.losses`?

river/tree/nodes/sgt_nodes.py
```diff
@@ -0,0 +1,223 @@
...
+                        # Categorical features are treated with a simple dict structure
+                        self._split_stats[idx] = {}
+                    self._split_stats[idx][x_val] = GradHessStats()
+                    self._split_stats[idx][x_val].update(gh, w)
+            else:
+                try:
+                    self._split_stats[idx].update(x_val, gh, w)
+                except KeyError:
+                    # Create a new quantizer
+                    self._split_stats[idx] = sgt.feature_quantizer._set_params(
```
MaxHalford, line 73: Just thinking: wouldn't it be possible to use a `collections.defaultdict` of some sort? I think it would ease readibility to delegate the initialization somewhere else than here. But no biggie!

river/tree/nodes/sgt_nodes.py
```diff
@@ -0,0 +1,223 @@
...
+        ] = {}
+        self._update_stats = GradHessStats()
+
+    def reset(self):
+        self._split_stats = {}
+        self._update_stats = GradHessStats()
+
+    def update(self, x: dict, gh: GradHess, sgt, w: float = 1.0):
+        for idx, x_val in x.items():
+            if not isinstance(x_val, numbers.Number) or idx in sgt.nominal_attributes:
```
MaxHalford, line 57: An `is_categorical` method might ease readability.

river/tree/nodes/sgt_nodes.py
```diff
@@ -0,0 +1,223 @@
...
+        dlms = self._update_stats.delta_loss_mean_var(best_split.merit.delta_pred)
+        best_split.merit.loss_mean = dlms.mean.get()
+        best_split.merit.loss_var = dlms.get()
+
+        for feature_idx in self._split_stats:
+            candidate = BranchFactory()
+            candidate.merit = GradHessMerit()
+            candidate.feature = feature_idx
+
+            if feature_idx in sgt.nominal_attributes:
```
MaxHalford, line 100: I would break this function down into two functions: one for categorical and one for numerical. But again, no biggie :)

river/tree/nodes/sgt_nodes.py
```diff
@@ -0,0 +1,223 @@
...
+    def update_stats(self):
+        return self._update_stats
+
+    @staticmethod
+    def delta_prediction(gh: GradHess, lambda_value: float):
+        # Add small constant value to avoid division by zero
+        return -gh.gradient / (gh.hessian + sys.float_info.min + lambda_value)
+
+    def __repr__(self):
+        return f"{self.prediction()}"
```
MaxHalford, line 223: You can just do `return str(self.prediction())`

river/tree/splitter/__init__.py
```diff
@@ -1,29 +1,37 @@
...
     "Splitter",
+    "Quantizer",
     "ExhaustiveSplitter",
     "GaussianSplitter",
     "HistogramSplitter",
     "EBSTSplitter",
     "QOSplitter",
     "TEBSTSplitter",
+    "DynamicQuantizer",
+    "StaticQuantizer",
```
MaxHalford, line 36: I like it when `__all__` is sorted.

river/tree/splitter/sgt_quantizer.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __len__(self):
+        if self.buckets:
+            return len(self.buckets)
+        return 0
+
+    def __iter__(self):
+        # TODO check that: in this case the null split should be choice
+        if self.buckets is None:
+            return
```
MaxHalford, line 167: To verify but I think you need to add a `yield` before or after the `return` to make this method a generator.

MaxHalford (commented):
Looks good! All this tree stuff goes over my head, I should take the time to understand it well one day. It really makes me happy to be working on River together 🤗

river/rules/amrules.py
smastelini: Nice! We can create a PR to tackle that :)

Is there a specific reason to avoid such imports?

river/tree/losses.py
smastelini: Yeap! You're right. I'll change that.

> Other question: is there a reason why you can't use the `Loss` class from `optim.losses`?

In the previous PR, I had [considered](https://github.com/online-ml/river/pull/458#issuecomment-785095468) this possibility.

But then we reached [another conclusion](https://github.com/online-ml/river/pull/458#issuecomment-798211842).

river/tree/splitter/__init__.py
smastelini: The order here is semantic. But this is not usually visible for the final user. Let's use the alphabetic order.

river/tree/splitter/sgt_quantizer.py
smastelini: Hmm, thanks for the tip! I'll check that :D

river/tree/nodes/sgt_nodes.py
smastelini: Done!

river/tree/nodes/sgt_nodes.py
smastelini: I see your point. The challenge is to use the same kind of structure to initialize a `dict` for categorical data or an instance of `tree.splitter.Quantizer` in the case of numbers. Do you have any suggestions?

river/rules/amrules.py
MaxHalford: They're considered to be messier, see [here](https://realpython.com/absolute-vs-relative-python-imports/). I think the community converged on that in the last few years. One practical thing I find is that I can copy/paste import statements without having to worry where my file is, so I don't have to worry about modifying the import statement 🤷🏼‍♂️

river/tree/losses.py
MaxHalford: Ok perfect 👌, I forgot 👴

river/tree/nodes/sgt_nodes.py
MaxHalford: I mean I see some ways but nothing very elegant. Let's keep it like this for now!

river/tree/nodes/sgt_nodes.py
smastelini: Nice catch! I totally agree. The end result is much simpler to read. You can check it at [this commit](https://github.com/online-ml/river/pull/674/commits/b33ed26aa625562c6df2c6e1bda4ebb59616a72c)

river/tree/nodes/sgt_nodes.py
smastelini: Totally agreed. I added this [method](https://github.com/online-ml/river/blob/b33ed26aa625562c6df2c6e1bda4ebb59616a72c/river/tree/nodes/sgt_nodes.py#L56-L58). I wish it had fewer parameters XD

river/rules/amrules.py
smastelini: Your last argument convinced me haha!

Let's do that soon!

MaxHalford (approved):

PR #697, opened 2021-09-02
Add test for newborn stats

MaxHalford:

### DIFF
--- a/river/stats/mode.py
+++ b/river/stats/mode.py
@@ -62,9 +62,7 @@ class Mode(base.Univariate):
         return self
 
     def get(self):
-        if self.counts:
-            return max(self.counts, key=self.counts.get)
-        return None
+        return max(self.counts, key=self.counts.get, default=None)
 
 
 class RollingMode(base.RollingUnivariate, utils.Window):
@@ -136,6 +134,4 @@ class RollingMode(base.RollingUnivariate, utils.Window):
         return self
 
     def get(self):
-        if self.counts:
-            return max(self.counts, key=self.counts.get)
-        return None
+        return max(self.counts, key=self.counts.get, default=None)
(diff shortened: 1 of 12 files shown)


river/stats/mode.py
```diff
@@ -62,7 +62,9 @@ def update(self, x):
         return self
 
     def get(self):
-        return max(self.counts, key=self.counts.get)
+        if self.counts:
+            return max(self.counts, key=self.counts.get)
+        return None
```
smastelini, line 67: What about using:

`return max(self.counts, key=self.counts.get, default=None)`?

river/stats/quantile.py
```diff
@@ -161,7 +161,7 @@ def get(self):
             length = len(self.heights)
             return self.heights[int(min(max(length - 1, 0), length * self.q))]
 
-        return 0
+        return None
```
smastelini, line 164: Nice catch!

river/stats/test_.py
```diff
@@ -48,6 +51,12 @@ def test_pickling(stat):
...
 
+@pytest.mark.parametrize("stat", load_stats(), ids=lambda stat: stat.__class__.__name__)
+def test_repr_with_no_updates(stat):
+    assert isinstance(repr(stat), str)
+    assert isinstance(str(stat), str)
+
+
 @pytest.mark.parametrize(
     "stat, func",
     [
```
smastelini, line 62: Nice!!

smastelini (changes requested):
Great work @MaxHalford! The new test is a must for pipeline viz! I left one small comment that might result in changes.

The PR is in good shape!

river/stats/mode.py
MaxHalford: Oh nice!

PR #703, opened 2021-09-09
PyTorch integration

kulbachcedric:
Hi,

I have adapted the PyTorch integration as follows:
1. Removed the batch size argument,
2. Added PyTorch2RiverClassifier with adaptive classes,
3. Homogenised NN build functions,
4. Possibility to add arguments to PyTorch2RiverRegressor & PyTorch2RiverClassifier that are required for the build_nn function. 

However, the `check_shuffle_features_no_impact` check still fails, do you have ideas to fix this?

For testing I ran the following:

```Python
import datasets as datasets

from river import compat, datasets, evaluate, metrics
from torch import nn, optim


def build_torch_mlp_classifier(n_features):
    net = nn.Sequential(
        nn.Linear(n_features, 5),
        nn.Linear(5, 5),
        nn.Linear(5, 5),
        nn.Linear(5, 5),
        nn.Linear(5, 1),
        nn.Sigmoid()
    )
    return net

if __name__ == '__main__':

    model = compat.PyTorch2RiverClassifier(
        build_fn=build_torch_mlp_classifier,
        loss_fn=nn.BCELoss,
        optimizer_fn=optim.Adam,
        learning_rate=1e-3
    )

    from river.utils import check_estimator

    check_estimator(model=model)
```

Best regards
Cedric


river/compat/__init__.py
```diff
@@ -38,8 +38,8 @@
     pass
 
 try:
-    from .pytorch import PyTorch2RiverRegressor
+    from .pytorch import PyTorch2RiverClassifier, PyTorch2RiverRegressor
 
-    __all__ += ["PyTorch2RiverRegressor"]
+    __all__ += ["PyTorch2RiverRegressor", "PyTorch2RiverClassifier"]
```
MaxHalford, line 43: Let's try to keep things alphabetical 😬

river/compat/pytorch.py
```diff
@@ -1,53 +1,231 @@
...
+    def __init__(
+        self,
+        build_fn,
+        loss_fn: typing.Type[torch.nn.modules.loss._Loss],
+        optimizer_fn: typing.Type[torch.optim.Optimizer] = torch.optim.Adam,
+        learning_rate=1e-3,
+        seed=42,
+        **net_params,
+    ):
+        self.build_fn = build_fn
```
MaxHalford, line 24: I like that

river/compat/__init__.py
MaxHalford, line 43: Let's try to keep things alphabetical 😬

river/compat/pytorch.py
```diff
@@ -1,53 +1,231 @@
...
+        self.optimizer.step()
 
     def learn_one(self, x, y):
+        if self.net is None:
+            self._init_net(n_features=len(list(x.values())))
+
+        x = torch.Tensor([list(x.values())])
+        y = torch.Tensor([[y]])
+        self._learn_one(x=x, y=y)
+        return self
```
MaxHalford, line 59: I agree that it makes sense to remove this fake mini-batching in `*_one` methods.

However, I think we should also have `learn_many` and `predict(_proba)_many` methods. What do you think?

river/compat/pytorch.py
```diff
@@ -1,53 +1,231 @@
...
+
+    def _filter_torch_params(self, fn, override=None):
+        """Filters `sk_params` and returns those in `fn`'s arguments.
+
+        # Arguments
```
MaxHalford, line 64: Try to respect our docstring conventions.

river/compat/pytorch.py
```diff
@@ -1,53 +1,231 @@
...
-        if self._batch_i == self.batch_size:
 
-            x = self.x_tensor(self._x_batch)
-            y = self.y_tensor(self._y_batch)
+    >>> model = compat.PyTorch2RiverClassifier(
+    ...     build_fn= build_torch_mlp_classifier,
+    ...     loss_fn=nn.BCELoss,
+    ...     optimizer_fn=optim.Adam,
+    ...     learning_rate=1e-3
+    ...     )
```
MaxHalford, line 133: Remove spaces

river/compat/pytorch.py
```diff
@@ -1,53 +1,231 @@
...
+        )
+
+    def learn_one(self, x: dict, y: base.typing.ClfTarget, **kwargs) -> base.Classifier:
+        self.classes.update([y])
+
+        # check if model is initialized
+        if self.net is None:
+            self._init_net(len(list(x.values())))
+
+        # check last layer and update if needed
```
MaxHalford, line 169: I would put this method in a separate method.

MaxHalford (changes requested):
Not bad! I think a couple of unit tests would be cool (checking net size etc.) But no biggie.

It's normal that `check_shuffle_features_no_impact` is failing. The thing is that in River we want all models to not care about the order of features. The thing is that you're passing the values of `x` to a list, without handling the order. You could either decide to sort features, but that's slow. We could also agree that we add a property to our models that indicates if a model should be expected to handle randomized orders or not.

Meta: could you add an entry to the release notes?

river/compat/pytorch.py
kulbachcedric: At least it would boost the tensor conversion, I'll add that.

docs/releases/unreleased.md
```diff
@@ -54,3 +54,8 @@
 ## utils
 
 - Moved `model_selection.expand_param_grid` to `utils.expand_param_grid`.
+
+## compat
+
+- Added `compat.pytorch.PyTorch2RiverClassifier`
+- Refactored `compat.pytorch.PyTorch2RiverRegressor`
```
MaxHalford, line 61: Actually you don't need the `.pytorch.`. It's important to be accurate here because we actually add a link when we build the documentation.

MaxHalford (approved):
LGTM! Thanks a lot 💚

MaxHalford:
Ping @kulbachcedric

kulbachcedric:
> Ping @kulbachcedric

It should work! :-) 
There might be some adaption in the future (Integration of AutoEncoders) as well as better default parameters.

MaxHalford:
Ok sounds good! There's an unspoken that the PR instigator should carry the PR to the finish line. Do you need anything else to get this PR merged?

kulbachcedric:
> Ok sounds good! There's an unspoken that the PR instigator should carry the PR to the finish line. Do you need anything else to get this PR merged?

Sorry, I am new into that, my fault ... 
Unfortunately I have no write access to resolve the last merge conflict, but then it should be ok

smastelini:
LGTM!

MaxHalford:
I merged it. But I just realized the unit tests were not running in the branch. So it breaks the main branch. I'll disable the tests for both estimators that inherit from `PyTorch2RiverBase`. @kulbachcedric can I ask you to take a look at the `check_estimator` function?

Issue #652, opened 2021-08-06
TFIDF clustering pipeline


nlyf:
## Versions

**river version**: 0.7.1
**Python version**: 3.7
**Operating system**: ubuntu 20.04

## Describe the bug

I'm trying to make online tf-idf clustering using DBSTREAM and DenStream. 
Can I use those algorithms with TFIDF feature extraction method?
Or I'am supposed to build whole feature-matrix table and then perform clustering? But this is not streaming clustering...

## Steps/code to reproduce

```python
from river import cluster,compose,feature_extraction
tfidf = compose.Pipeline(("tfidf",feature_extraction.TFIDF()),
                        ("clusterer",cluster.DBSTREAM()))

corpus = [
    'This is the first document.',
    'This document is the second document.',
    'And this is the third one.',
    'Is this the first document?',
]
for sentence in corpus:
    tfidf.transform_one(sentence)
```
Get the following traceback
```python 
    201                     + self._gaussian_neighborhood(x, self.micro_clusters[i].center)
    202                     * (x[j] - self.micro_clusters[i].center[j])
--> 203                     for j in self.micro_clusters[i].center.keys()
    204                 }
    205                 self.micro_clusters[i].last_update = self.time_stamp

KeyError: 'first'

```
And traceback for DenStream respectively
```
    386             self.initial_weight = 2 ** (self.decaying_factor * self.timestamp)
    387             self.LS = x
--> 388             self.SS = {i: (x[i] * x[i]) for i in range(self.dim)}
    389             self.IWLS = {
    390                 i: (2 ** (self.decaying_factor * self.timestamp)) * x[i]

KeyError: 0

```

MaxHalford:
@hoanganhngo610 could you take a look at this? I've tried with `DBStream` and `DenStream` and both don't work. Some initialization stuff is missing somewhere.

nlyf:
Hi, I'm still very curious. Have you got any ideas? @MaxHalford @hoanganhngo610

MaxHalford:
Sorry @nlyf I closed this by mistake. I would like @hoanganhngo610 to take a look at what's going on.

ribab:
@MaxHalford @nlyf @hoanganhngo610 

This is actually something I am very interested at implementing.

I think the issue has something to do with TFIDF outputting data in the format of dictionaries whereas dbstream and denstream expect the input in the form of a 2D array.

I've only just discovered RiverML, so my analysis of this issue might be incorrect if I'm missing something.

[...]

PR #711, opened 2021-09-17
Fix clusterer indexing

jacobmontiel:
This PR introduces fixes for `Clustream` and `DenStream`.

An indexing issue affecting `DenStream` was reported in #652. The same issue was found in `CluStream`.

This PR also includes a refactoring on the `DenStream` implementation, including:
- Simplify code.
- Fix indexing bug.
- Refactor micro cluster class.
- Improve documentation.
- Improve modularity.
- Rename class parameters to match the names used in the original paper.
- Update test case.


river/cluster/denstream.py
```diff
@@ -43,411 +44,424 @@ class DenStream(base.Clusterer):
...
+
+    @property
+    def centers(self):
+        if self.clusters:
+            return {
+                k: cluster.calc_center(self.timestamp)
+                for k, cluster in self.clusters.items()
+            }
+        else:
+            return {}
```
MaxHalford, line 177: No need for the `else`

MaxHalford: In fact, if `self.clusters` is an empty `dict`, you don't need the `if` condition

MaxHalford (approved):
LGTM, just a couple of nits

river/cluster/denstream.py
smastelini: +1 to that

river/cluster/clustream.py
```diff
@@ -305,13 +305,15 @@ def __init__(
         if x is not None and sample_weight is not None:
             # Initialize with sample x
             self.n_samples = 1
-            self.linear_sum = {}
-            self.squared_sum = {}
+            self.linear_sum = defaultdict(float)
+            self.squared_sum = defaultdict(float)
             for key in x.keys():
                 self.linear_sum[key] = x[key] * sample_weight
                 self.squared_sum[key] = x[key] * x[key] * sample_weight
```
smastelini, line 312: In the future we should consider using components from `stats`.

river/cluster/denstream.py
```diff
@@ -43,411 +44,424 @@ class DenStream(base.Clusterer):
...
-            2 ** (-self.current_time * self.decaying_factor)
-        ) * self.initial_weight
-        return weight
-
-    @property
-    def center(self):
-        center = {i: self.weighted_linear_sum[i] / self.weight for i in range(self.dim)}
+        self.N = 1
+        self.feature_keys = set(x.keys())
+        self.liner_sum = x
```
smastelini, line 406: linear_sum?

river/cluster/denstream.py
```diff
@@ -43,411 +44,424 @@ class DenStream(base.Clusterer):
...
         return radius
 
-    def add(self, cluster):
-        assert self.dim == cluster.dim
+    def insert(self, x, timestamp):
+        if set(x.keys()) != self.feature_keys:
+            raise KeyError(
+                f"Features mismatch. Expected {self.feature_keys}",
+                f"got {set(x.keys())}",
+            )
```
smastelini, line 446: Can't the union of both key sets be used?

river/cluster/denstream.py
```diff
@@ -43,411 +44,424 @@ class DenStream(base.Clusterer):
...
+        for k in self.feature_keys:
+            self.liner_sum[k] += x[k]
+            self.squared_sum[k] += x[k] * x[k]
+
+    def merge(self, cluster):
+        if self.feature_keys != cluster.feature_keys:
+            raise KeyError(
+                f"Features mismatch. Expected {self.feature_keys}",
+                f"got {cluster.feature_keys}",
+            )
```
smastelini, line 458: Same here :)

smastelini (changes requested):
It's looking good! I left a couple of minor comments for further discussion. Please, do not hesitate in contacting me if you have any questions.

river/cluster/denstream.py
jacobmontiel: Nice catch. Much better now.

Thanks for the review!

river/cluster/clustream.py
jacobmontiel: Indeed, this will be addressed in a separate PR as it might be possible to replicate for other methods as well.

river/cluster/denstream.py
jacobmontiel: I just renamed it to follwo the convention in CluStream but I agree that it sounds strange. Basically, we are tracking the sum of values and the sum of squared values. Any suggestion is welcomed.

river/cluster/denstream.py
jacobmontiel: Interesting, although I am not completely sure how a new feature would fare alongside the existing metrics? I am not saying no, I am just thinking out loud.

river/cluster/clustream.py
smastelini: Cool!

river/cluster/denstream.py
smastelini: I don't know either haha

But still, I think it's better to keep processing instead of raising an error. What do you think, @MaxHalford?

hoanganhngo610: @jacobmontiel @smastelini IMO, we can have two solutions for this: One is either to raise a 
```python 
KeyError
```
or to simply ignore it and continue, with a 
```python
pass
```
Personally, I would prefer the second way, in the sense that it would not break our stream for one small error.

river/cluster/denstream.py
hoanganhngo610: Same as above!

river/cluster/denstream.py
jacobmontiel: Nice catch. Fixed in c7ec6f9a8ae5cb5ffdbf5d9d6630c6105d8cfc60

river/cluster/denstream.py
jacobmontiel: Added support for emerging features in c7ec6f9a8ae5cb5ffdbf5d9d6630c6105d8cfc60

river/cluster/denstream.py
jacobmontiel: Added support for emerging features in c7ec6f9a8ae5cb5ffdbf5d9d6630c6105d8cfc60

PR #729, opened 2021-10-05
FIX estimator_checks

kulbachcedric:
Hi,
Here a quick fix for the missing tests.
However, I disabled the following tests:
- `check_pickling:` A pickable torch estimator could be added in the future
- `check_shuffle_features_no_impact:` Changing features are quite difficult to handle in NNs
- `check_emerging_features`
- `check_disappearing_features`
- `check_predict_proba_one:` 
- `check_predict_proba_one_binary:` Actually the classifier is able to adapt to the number of classes, what makes it difficult to give directly at the beginning of the dataset `True` and `False` as result.

Best
Cedric

### DIFF
--- a/river/compat/pytorch.py
+++ b/river/compat/pytorch.py
@@ -32,15 +32,35 @@ class PyTorch2RiverBase(base.Estimator):
         torch.manual_seed(seed)
         self.net = None
 
+    @classmethod
     def _unit_test_params(self):
         def build_torch_linear_regressor(n_features):
-            net = torch.nn.Sequential(torch.nn.Linear(n_features, 1))
+            net = torch.nn.Sequential(
+                torch.nn.Linear(n_features, 1), torch.nn.Sigmoid()
+            )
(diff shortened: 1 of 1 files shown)


MaxHalford:
Any news @kulbachcedric? Please tell us if you don't have time to work on this. If so, I can handle this.

kulbachcedric:
> Any news @kulbachcedric? Please tell us if you don't have time to work on this. If so, I can handle this.

Unfortunately, the `check_doc` is still tested, even if it is listed within the `_unit_test_skips` function. Do you know how to skip the doctests?

MaxHalford:
`check_doc` does not disable the doctest. It tests that the class has a doc.

Why do you want to skip the doctests? That doesn't sound like a good idea to me.

Meta: don't hesitate to ask these question by yourself.

MaxHalford:
I think rounding is ok. I'll do the necessary changes!

PR #746, opened 2021-10-28
Finish bandit refactoring

MaxHalford:
This PR is to finish the revamp of model selection based on bandits. Essentially, I refactored the code for the UCB policy. I also added a unit test to check that model order does not matter during model selection. This is to address what @etiennekintzler mentions in #738. I found that adding a burn-in phase worked well, very much like what Etienne did previously.

@etiennekintzler, I didn't find that scaling the reward was necessary. I get it that it's nice to have rewards on the same scale. But I'm doubtful as to if that's really essential. For UCB, you have to pick an exploration parameter anyway, regardless of the scale. For Epsilon Greedy, the scale does not matter. The variance from the initial conditions simply gets resolved with the burn-in phase.

What do you think @etiennekintzler?

PS: sorry there's a lot of noise in this PR, I got carried away 😅


MaxHalford:
Gentle ping @etiennekintzler, any progress on your end? Just asking because we would like to merge this PR before the next release. But seriously, no pressure. I just wanted to reach out :)

river/model_selection/bandit.py
```diff
@@ -0,0 +1,120 @@
...
+                key=lambda arm: arm.metric.get(),
+                reverse=self.metric.bigger_is_better,
+            )
+        ]
+
+    def __repr__(self):
+        return utils.pretty.print_table(
+            headers=["Ranking", self.metric.__class__.__name__, "Pulls", "Share",],
+            columns=[
+                [str(r) for r in self.ranking],
```
etiennekintzler, line 69: I am not sure that you are really computing the ranking here. The ranking would be 
 `np.argsort(self.ranking)` (in numpy) or `sorted(range(len(r)), key=lambda k: r[k])` (in vanilla python). 
If you want to see an example of mismatch between the 2 you can try the docstring example with the following learning rates : 
`lrs = [0.0001, 0.001, 1e-05, 0.01]`

MaxHalford: Good catch! I'll add a unit test.

MaxHalford:
Indeed @etiennekintzler you're right, I forgot to handle the sub-gaussian aspect. From what I understand, standard scaling the target should the trick, correct?

etiennekintzler:
> Indeed @etiennekintzler you're right, I forgot to handle the sub-gaussian aspect. From what I understand, standard scaling the target should the trick, correct?

@MaxHalford this is my understanding as well !

 Semi-related but as `Arm.n_pulls` is used in the exploitation/exploration trade-off, shouldn't it be reset to 0 after the burn in ?

MaxHalford:
> Semi-related but as Arm.n_pulls is used in the exploitation/exploration trade-off, shouldn't it be reset to 0 after the burn in ?

I thought about that too! I would say it's fine as it is, because arms do in fact get pulled during the `burn_in`. I find the contrary confusing. We can revisit this decision later though!

etiennekintzler:
Yes but the exploration part `2*log(n_total)/n_arm`  is much lower after the burn in.

As an example imagine 5 arms bandit and let's say you pull an arm for the first time: 
- without burn-in, after 3 total pulls: 2*np.log(3)/1=**2.2**
- after burn-in, that is after 5*100 total pulls : 2*np.log(5*100)/(100+1)=**0.12**

I don't know to what extent it's important since there is a lot of exploration done in the burn-in phase anyway. That is just something to keep in mind in case.

Issue #359, opened 2020-10-03
Scaler in pipeline not fitting


mtanghu:
## Versions

**`creme` version** : 
0.6.1
**Python version** : 
3.6.9
**Operating System** :
Windows 10

## Describe the bug

It looks like creating a pipeline with a StandardScaler() doesn't fit the StandardScaler for me. When I use the debug_one() it seems that the scaler always scales to 0.

## Steps/Code to Reproduce

Running this sample shows the issue I'm having with the scaler.

```python
from creme import preprocessing as creme_preprocessing
from creme import linear_model

scaled_model = creme_preprocessing.StandardScaler() | linear_model.LinearRegression()

for i in range(1000):
    scaled_model.fit_one({"i": i}, i)
    
print(scaled_model.debug_one({"a": 0}))
print(scaled_model.debug_one({"a": 1000}))
````
produces:

```
0. Input
--------
a: 0 (int)

1. StandardScaler
-----------------
a: 0.00000 (float)

2. LinearRegression
-------------------
Name        Value      Weight       Contribution  
Intercept    1.00000    950.00000      950.00000  
        a    0.00000      0.00000        0.00000  

Prediction: 950.00000
0. Input
--------
a: 1000 (int)

1. StandardScaler
-----------------
a: 0.00000 (float)

2. LinearRegression
-------------------
Name        Value      Weight       Contribution  
Intercept    1.00000    950.00000      950.00000  
        a    0.00000      0.00000        0.00000  

Prediction: 950.00000
```

MaxHalford:
This is not a bug, it's a feature :)

The unsupervised parts of a pipeline are fitted when `predict_one` is called. Basically, this is the optimal behaviour. Performance would be much lower if we updated the unsupervised parts only when the model is trained. Indeed, if the data is available during `predict_one`, then why wouldn't we updated the unsupervised part of the pipeline there and then?

If you think about it, it's weird to train a pipeline without making predictions before anyway. The only case where you would to do that is pre-training (warming up). But even in that case you would want to be making predictions to know the performance of your model.

I've thought about this long and hard and I keep thinking about this. I'm happy to discuss alternative solutions.

mtanghu:
Oh I see, that makes sense! I appreciate the answer!

Issue #130, opened 2019-07-10
Scaler in debug_one returning nan


konradsitarz:
I bumped into the issue when I tried to use Pipeline debug_one method. Scaler wasn't fitting the data using the fit_one method before transforming and the result was dict with all nan values.
```
0. Input
--------
STARTDATETIME: 2018-09-21 22:00:00 (Timestamp)
HUMIDITY: 68.59412222222221 (float)
PM10: 37.111111111111114 (float)
PM25: 32.111111111111114 (float)
TEMPERATURE: 19.59688888888889 (float)

1. ['HUMIDITY', 'PM10', 'PM25', 'TEMPERATURE']
----------------------------------------------
PM10: 37.111111111111114 (float)
HUMIDITY: 68.59412222222221 (float)
PM25: 32.111111111111114 (float)
TEMPERATURE: 19.59688888888889 (float)

2. MinMaxScaler
---------------
PM10: nan (float)
HUMIDITY: nan (float)
PM25: nan (float)
TEMPERATURE: nan (float)

3. LinearRegression
-------------------
0.0 * nan (PM10) +
0.0 * nan (HUMIDITY) +
0.0 * nan (PM25) +
0.0 * nan (TEMPERATURE)

nan
```

I changed slightly the implementation of debug_one in Pipeline class. I added call to fit_one if transformer has this method.
```

        for i, t in enumerate(self.transformers):
           # my change
            if callable(getattr(t, 'fit_one', None)):
                t.fit_one(x)
           # ----
            if isinstance(t, union.TransformerUnion):
                print_title(f'{i+1}. Transformer union')
                for j, (name, sub_t) in enumerate(t.items()):
                    print_title(f'{i+1}.{j} {name}', indent=True)
                    print_features(sub_t.transform_one(x), indent=True)
                x = t.transform_one(x)
                print_features(x)
            else:
                print_title(f'{i+1}. {t}')
                x = t.transform_one(x)
                print_features(x)
```

With this change I was able to see values of scaled parameters.
At the end I would like to thank you for very interesting project. I hope it will raise to be commonly used in incremental learning. Cheers :)

MaxHalford:
Hey, thank you for the kind words!

[...]

PR #768, opened 2021-11-13
Refactor pipeline order of resolution

MaxHalford:
I noticed a weird thing while I was working on one-class SVM.

If you take the current code base, both of the following scripts produce the exact same result:

```py
from river import anomaly
from river import compose
from river import datasets
from river import metrics

model = compose.Pipeline(
    preprocessing.MinMaxScaler(),
    anomaly.HalfSpaceTrees(seed=42)
)

auc = metrics.ROCAUC()

for x, y in datasets.CreditCard().take(2500):
    score = model.score_one(x)
    model = model.learn_one(x)
    auc = auc.update(y, score)

print(auc)
```

```py
from river import anomaly
from river import compose
from river import datasets
from river import metrics

model = compose.Pipeline(
    preprocessing.MinMaxScaler(),
    anomaly.HalfSpaceTrees(seed=42)
)

auc = metrics.ROCAUC()

for x, y in datasets.CreditCard().take(2500):
    score = model.score_one(x)
    auc = auc.update(y, score)

print(auc)
```

Do you notice the difference? The `model = model.learn_one(x)` line got removed. Weird right? That's because `HalfSpaceTrees` is an unsupervised estimator, and therefore gets updated when `score_one` is called. 

I thought about this quite a bit. I don't think it's good. It makes more sense if the last step of the pipeline gets updated when `learn_one` is called. Regardless of if it's unsupervised or not.

I proceeded to make the necessary changes in this pull request. This is what I did:

- Corrected the order of resolution to handle the case where the last step of the pipeline is unsupervised. This slightly changes some unit test outputs because before the unsupervised estimators got updated before producing a prediction.
- Added a `utils.log_method_calls` context manager. It's a debugging utility to track which methods are called and in which order. Very useful for debugging but also for unit testing.
- Added a `utils.warm_up_mode` that replaces the `learn_unsupervised` parameters that we had before. This is much more straightforward and should be more user-friendly. Recall that is a thing where users trip over: see #130, #359, and #499.
- Oh and I also removed the `load_boston` dataset because it's being deprecated.


river/compose/pipeline.py
```diff
@@ -236,9 +316,13 @@ def _set_params(self, new_params: dict = None):
     def _supervised(self):
         return any(step._supervised for step in self.steps.values())
 
+    @property
+    def _last_step(self):
+        return list(self.steps.values())[-1]
+
```
smastelini, line 322: Cool

PR #770, opened 2021-11-16
Fix PyTorch classifier

kulbachcedric:
Hi,

this PR contains a quick fix for `compat.PyTorch2RiverClassifier`, where the Classifier could not adapt to new classes when the last layer of from the PyTorch model was trainable.

Best
Cedric

### DIFF
--- a/river/compat/pytorch.py
+++ b/river/compat/pytorch.py
@@ -196,22 +196,26 @@ class PyTorch2RiverClassifier(PyTorch2RiverBase, base.Classifier):
     def _update_classes(self):
         self.n_classes = len(self.classes)
         layers = list(self.net.children())
+        # Get last trainable layer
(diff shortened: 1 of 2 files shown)


river/compat/pytorch.py
```diff
@@ -43,6 +46,23 @@ def build_torch_linear_regressor(n_features):
...
         }
 
+    @classmethod
+    def _unit_test_skips(self):
+        """Indicates which checks to skip during unit testing.
+
+        Most estimators pass the full test suite. However, in some cases, some estimators might not
+        be able to pass certain checks.
+
+        """
```
MaxHalford, line 56: Weren't these tests working before? 

Meta: you don't need to include the comment docstring for inherited methods.

Issue #776, opened 2021-11-17
New feature suggestion: InteractionsUnion


oegedijk:
Hi all,

I was looking for a way to have more control over interactions in river Pipelines. The `PolynomialExtender` adds interactions of every feature with every feature, which is often a bit more than you want or need. In Vowpal Wabbit you have the concept of namespaces that you can interact with each other, e.g. `|a feature_1 feature_2 |b feature_3 feature_4` and then add `-q ab` to the command line options to generate all interactions between namespace `a` and namespace `b`. 

I was looking for something similar in river, and came up with the following. It's basically a mixture of `TransformerUnion` and `PolynomialExtender`. 

Would this be a useful addition to the library? Or is there maybe already some other way of achieving the same goal that I missed?

```python 

import collections
import itertools
import types

from river import base, utils

from river.compose import func

class InteractionsUnion(base.Transformer):
    
    def __init__(self, *transformers, interactions_only=False):
        self.transformers = {}
        for transformer in transformers:
            self += transformer
        self.interactions_only = interactions_only

    def __getitem__(self, key):
        """Just for convenience."""
        return self.transformers[key]

    def __len__(self):
        """Just for convenience."""
        return len(self.transformers)

    def __str__(self):
        return " + ".join(map(str, self.transformers.values()))

    def __repr__(self):
        return (
            "TransformerUnion (\n\t"
            + "\t".join(
                ",\n".join(map(repr, self.transformers.values())).splitlines(True)
            )
            + "\n)"
        ).expandtabs(2)

    def _get_params(self):
        return {
            name: transformer._get_params()
            for name, transformer in self.transformers.items()
        }

    def _set_params(self, new_params: dict = None):

        if new_params is None:
            new_params = {}

        return TransformerUnion(
            *[
                (name, new_params[name])
                if isinstance(new_params.get(name), base.Estimator)
                else (name, step._set_params(new_params.get(name, {})))
                for name, step in self.transformers.items()
            ]
        )

    @property
    def _supervised(self):
        return any(t._supervised for t in self.transformers.values())

    def _add_step(self, transformer):
        """Adds a transformer while taking care of the input type."""

        name = None
        if isinstance(transformer, tuple):
            name, transformer = transformer

        # If the step is a function then wrap it in a FuncTransformer
        if isinstance(transformer, (types.FunctionType, types.LambdaType)):
            transformer = func.FuncTransformer(transformer)

        def infer_name(transformer):
            if isinstance(transformer, func.FuncTransformer):
[...]

PR #777, opened 2021-11-17
Calculate interactions between feature groups

MaxHalford:
This PR addresses a very good suggestion from @oegedijk in #776.

I'm a bit tired so not 100% sure inheriting from `TransformerUnion` is a good idea as is, but it works. Maybe there's some edge cases I'm not seeing but we can catch those later.

### DIFF
--- a/river/compose/product.py
+++ b/river/compose/product.py
@@ -59,7 +59,7 @@ class TransformerProduct(union.TransformerUnion):
     """
 
     def __repr__(self):
-        return repr(super()).replace("Union", "Product", 1)
+        return super().__repr__().replace("Union", "Product", 1)
 
     def __mul__(self, other):
         return self._add_step(other)
--- a/river/compose/product.py
+++ b/river/compose/product.py
@@ -1,6 +1,6 @@
 import itertools
 
-from river import base, utils
+from river import utils
 
 from . import union
 
--- a/river/compose/product.py
+++ b/river/compose/product.py
@@ -1,0 +1,73 @@
+import itertools
+
+from river import base, utils
+
+from . import union
+
+__all__ = ["TransformerProduct"]
+
+
+class TransformerProduct(union.TransformerUnion):
+    """Computes interactions between the outputs of a set transformers.
+
+    This is for when you want to add interaction terms between groups of features. It may also
+    be used an alternative to `feature_extraction.PolynomialFeatures` when the latter is overkill.
+
+    Parameters
+    ----------
+    transformers
+        Ideally, a list of (name, estimator) tuples. A name is automatically inferred if none is
+        provided.
+
+    Examples
+    --------
+
+    Let's say we have a certain set of features with two groups. In practice these may be different
+    namespaces, such one for items and the other for users.
+
+    >>> x = dict(
+    ...     a=0, b=1,  # group 1
+    ...     x=2, y=3   # group 2
+    ... )
+
+    We might want to add interaction terms between groups `('a', 'b')` and `('x', 'y')`, as so:
+
+    >>> from pprint import pprint
+    >>> from river.compose import Select, TransformerProduct
+
+    >>> product = TransformerProduct(
+    ...     Select('a', 'b'),
+    ...     Select('x', 'y')
+    ... )
+    >>> pprint(product.transform_one(x))
+    {'a*x': 0, 'a*y': 0, 'b*x': 2, 'b*y': 3}
+
+    This can also be done with the following shorthand:
+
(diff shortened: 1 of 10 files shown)


oegedijk:
Looks great! Very nice clean implementation. 

I would probably add a deprecation error to the old `FeatureExtender` and keep it around for a while though...

river/compose/product.py
```diff
@@ -0,0 +1,73 @@
...
+from . import union
+
+__all__ = ["TransformerProduct"]
+
+
+class TransformerProduct(union.TransformerUnion):
+    """Computes interactions between the outputs of a set transformers.
+
+    This is for when you want to add interaction terms between groups of features. It may also
+    be used an alternative to `feature_extraction.PolynomialExtender` when the latter is overkill.
```
oegedijk, line 14: should be changed to PolynomialFeatures

oegedijk:
I was also thinking whether there is a way to make it even more readable with R-style function definitions (i.e. basically automatically insert all the `Select()`). 

So you could do something like `Formula("a + b*x")` which would then generate `Select("a") + Select("b") * Select("x")` under the hood.

river/compose/product.py
MaxHalford: Actually I'm keeping the name `PolynomialExtender`

MaxHalford:
Mmm well I would say we can keep `PolynomialExtender` as is. It's still a nice thing to have around.

MaxHalford:
> So you could do something like Formula("a + b*x") which would then generate Select("a") + Select("b") * Select("x") under the hood.

Indeed! That's what Patsy does in the Python ecosystem. I would handle this in a separate PR though. It requires some string parsing. I'm not 100% on how to get it done properly. But at least now we have some good building blocks to build it.

oegedijk:
> Indeed! That's what Patsy does in the Python ecosystem. I would handle this in a separate PR though. It requires some string parsing. I'm not 100% on how to get it done properly. But at least now we have some good building blocks to build it.

Yeah definitely out of scope for this PR, would be a great readability definition though!

oegedijk:
> Mmm well I would say we can keep `PolynomialExtender` as is. It's still a nice thing to have around.

Ah, hadn't seen the revert renaming commit. I do think `PolynomialFeatures` is a more descriptive name, but also better not to have too many unnecessary breaking changes :)

oegedijk:
Stress testing a bit, this seems to work:

```
group_1 = Select('a', 'b') | OneHotEncoder(sparse=True)
group_2 = Select('x', 'y')
x = dict(a=1, b=2, x=3, y=4, z=5)


tp1 = TransformerUnion(group_1, group_2,TransformerProduct(group_1, group_2))
tp1.transform_one(x)
```

But this does not:

```
tp2 = group_1 + group_2 + group_1 * group_2
tp2.transform_one(dict(a=1, b=2, x=4, y=4, z=5))
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_6101/2982673641.py in <module>
      1 tp2 = group_1 + group_2 + group_1 * group_2
----> 2 tp2.transform_one(dict(a=1, b=2, x=4, y=4, z=5))

/opt/conda/lib/python3.7/site-packages/river/compose/union.py in transform_one(self, x)
    231         return dict(
    232             collections.ChainMap(
--> 233                 *(t.transform_one(x) for t in self.transformers.values())
    234             )
    235         )

/opt/conda/lib/python3.7/site-packages/river/compose/union.py in <genexpr>(.0)
    231         return dict(
    232             collections.ChainMap(
--> 233                 *(t.transform_one(x) for t in self.transformers.values())
    234             )
    235         )

/opt/conda/lib/python3.7/site-packages/river/compose/grouper.py in transform_one(self, x)
     47 
     48     def transform_one(self, x):
---> 49         key = self._get_key(x)
     50         return self.transformers[key].transform_one(x)

/opt/conda/lib/python3.7/site-packages/river/compose/grouper.py in _get_key(self, x)
     39 
     40     def _get_key(self, x):
---> 41         return "_".join(str(x[k]) for k in self.by)
     42 
     43     def learn_one(self, x):

/opt/conda/lib/python3.7/site-packages/river/compose/grouper.py in <genexpr>(.0)
     39 
     40     def _get_key(self, x):
---> 41         return "_".join(str(x[k]) for k in self.by)
     42 
     43     def learn_one(self, x):

KeyError: Pipeline (
  Select (
    a
    b
  ),
  OneHotEncoder (
    sparse=True
  )
)
```

MaxHalford:
Good catch! Fixing.

oegedijk:
Looks good!

Issue #779, opened 2021-11-19
DBSTREAM throws a KeyError when handling sparse data


SeanWeber:
## Versions

**river version**: 0.8.0
**Python version**: 3.9
**Operating system**: Ubuntu 21.10

## Describe the bug

When a micro-cluster's center is updated, the algorithm tries to find all keys in the data point that are in the micro-cluster. If the data point is missing a key, a KeyError is thrown. Instead, it should assume any missing keys have a 0 value. For example: `x.get(j, 0.0)` 

## Steps/code to reproduce

```python
import river

# We choose a clustering_threshold large enough that the second data point updates the first micro-cluster
dbstream = river.cluster.DBSTREAM(clustering_threshold=5)

dbstream.learn_one({1: 1, 2: 2})
dbstream.learn_one({1: 1})  # KeyError: 2
```

smastelini:
Pinging @hoanganhngo610 :)

MaxHalford:
In the meantime you can use an imputation method to manually add 0s to your data. In fact, making it explicit that way might be a good idea. I'm not fully convinced (yet) that features should default to 0 "magically".

hoanganhngo610:
Dear @SeanWeber, thank you for raising the issue. Unfortunately, when we are designing the clustering algorithms, we expect that all data points coming through have to have the same dimension, with the same keys. This assures that the microclusters and macroclusters can be properly formed. As such, sparse data are not covered within our interest. I also think @jacobmontiel agrees with me on this.

MaxHalford:
It's good that we commit to a decision on this, thanks @hoanganhngo610! Imputation it is then.

hoanganhngo610:
If everyone agrees with this, I would proceed to close the issue. However, please do not hesitate to comment if you find any other issue since there are still tons of think to do with the clustering module @SeanWeber!

PR #780, opened 2021-11-19
Modify DBSTREAM to handle sparse data

SeanWeber:
When there are missing keys in clusters and data points, we use the value of 0.0. This way we can process sparse data sets.

This fixes https://github.com/online-ml/river/issues/779


MaxHalford:
In principle I agree with this PR. But I would like us to think about clustering methods in general, and not just DBSTREAM. Would love if @hoanganhngo610 could put some thought into this.

hoanganhngo610:
As mentioned in #779, although the fix is really nice, I would not encourage it as it would turn out to be a bad practice for future users.

hoanganhngo610:
Proceed to close the PR as issue #779 has been resolved.

PR #786, opened 2021-11-29
Add OneHotListEncoder to the preprocessing `one_hot` module

fox-ds:
This commit will add functionality for one-hot encoding variables that are held in a list.  Currently, the `OneHotEncoder` takes features with a single value, and converts them to dummy variables, where in practice we also see the need for passing lists to this encoder. It's implemented as a child class of the regular `OneHotEncoder` class because of the similarities, with the major difference in that it checks for the presence of lists/sets and afterwards, loops over these values to add them.

### DIFF
--- a/river/preprocessing/one_hot.py
+++ b/river/preprocessing/one_hot.py
@@ -1,8 +1,9 @@
 import collections
 
 from river import base
+from typing import Dict, Union
 
-__all__ = ["OneHotEncoder"]
+__all__ = ["OneHotEncoder", "OneHotListEncoder"]
 
 
 class OneHotEncoder(base.Transformer):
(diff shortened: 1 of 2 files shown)


MaxHalford:
Hello there! This feature is a nice suggestion, I like it.

However, I would rather add this functionality as a default behavior to the existing one-hot encoder. I see many advantages to this. What do you think?

fox-ds:
Hi Max,

Ah, that's indeed a good idea of which I hadn't though yet! Let me rewrite the code a bit and resolve some of the code quality issues. Thanks!

river/preprocessing/one_hot.py
```diff
@@ -1,16 +1,18 @@
...
+__all__ = ["OneHotEncoder"]
 
 
 class OneHotEncoder(base.Transformer):
     """One-hot encoding.
 
-    This transformer will encode every feature it is provided it with. You can apply it to a
-    subset of features by composing it with `compose.Select` or `compose.SelectType`.
+    This transformer will encode every feature it is provided it with.
+    If an iterable is provded, this transformer will encode every entry in the iterable.
```
MaxHalford, line 13: `provided` typo

I would say "list or set" rather than "iterable" to be precise. Indeed, a `str` is an iterable.

river/preprocessing/one_hot.py
```diff
@@ -152,71 +113,23 @@ class OneHotListEncoder(OneHotEncoder):
...
-            if type(xi) not in [list, set]:
-                continue
-
-            # Here, it's at least a list or a set.
-            # If it's empty, we don't have any elements and will
-            # effectively skip it. Otherwise, add all the elements of the
-            # list to the defaultdict
-            for xj in xi:
-                self.values[i].add(xj)
+            if isinstance(xi, (list, set, tuple)):
```
MaxHalford, line 121: I wouldn't include tuples. Indeed, those can be singular values in some case.

fox-ds: Okay, I removed the tuple.

river/preprocessing/one_hot.py
fox-ds: Removed the extra 'it' and replaced Iterable with list/set

river/preprocessing/one_hot.py
```diff
@@ -1,16 +1,15 @@
...
 __all__ = ["OneHotEncoder"]
 
 
 class OneHotEncoder(base.Transformer):
     """One-hot encoding.
 
-    This transformer will encode every feature it is provided it with.
-    If an iterable is provded, this transformer will encode every entry in the iterable.
+    This transformer will encode every feature it is provided with.
+    If a list or set is provded, this transformer will encode every entry in the list/set.
```
MaxHalford, line 12: `provided`, not `provded`

fox-ds: Oh wow, sorry, my bad. Fixed it.

PR #796, opened 2021-12-10
Inherit AnomalyDetector from Classifier

kulbachcedric:
Hi all,

based on #794 I changed the `AnomalyDetector` to inherit from a `base.Classifier`.
However, I think there are still some smaller bugs and open questions / discussion points:

Errors / Bugs:
- For the tests, the Pishing dataset is not found in cache - Here, I don't know how to solve this issue.

Discussion / Open Questions:
- the `learn_one` function of the `base.Classifier` takes the arguments `self, x: dict, y: base.typing.ClfTarget, **kwargs`. Since an `AnomalyDetector` is now a `base.Classifier` a work around is to pass `y = None` as default, as suggested in the PR. 
- I added the `predict_proba_one` function, which differs from the `score_one` function in the return type.
- In the future I would suggest to move the `AnomalyDetector` into the base directory?
- In the future a `predict_learn_one`method, could lead to faster Estimators, since in some cases one only need to predict once to update the model. 

What do you think? 

Best
Cedric

### DIFF
--- a/river/anomaly/base.py
+++ b/river/anomaly/base.py
@@ -1,17 +1,24 @@
 import abc
+import typing
 
 from river import base
 
 
-class AnomalyDetector(base.Estimator):
+class AnomalyDetector(base.Classifier):
     """An anomaly detector."""
 
     @property
     def _supervised(self):
         return False
 
+    @property
+    def _multiclass(self):
+        return False
+
     @abc.abstractmethod
-    def learn_one(self, x: dict) -> "AnomalyDetector":
+    def learn_one(
+        self, x: dict, y: base.typing.ClfTarget = None, **kwargs
+    ) -> "AnomalyDetector":
         """Update the model.
 
         Parameters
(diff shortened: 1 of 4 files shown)


MaxHalford:
Actually in #794 I said that I don't think anomaly detectors should inherit from `Classifier`. I really don't think this is a good idea. I would rather think about how to allow metrics to support anomaly detectors.

smastelini:
> Actually in #794 I said that I don't think anomaly detectors should inherit from `Classifier`. I really don't think this is a good idea. I would rather think about how to allow metrics to support anomaly detectors.

I second to that

MaxHalford:
Sorry Cedric, I should have been clearer!

kulbachcedric:
No Problem, Max! I'll work on the `progressive_validation` :-) However, in the long run, I think it makes sense to handle it similarly to sklearn.

PR #797, opened 2021-12-10
Refactor anomaly detection

kulbachcedric:
Hi all,

based on discussion #794 , you find in this PR a quick-fix so that the `anomaly.AnomalyDetector` works with `Tracks`. 
In addition, I have adapted the `metrics.ROCAUC` metric so that it is now also supported by an `anomaly.AnomalyDetector`.

Best

### DIFF
--- a/river/anomaly/base.py
+++ b/river/anomaly/base.py
@@ -1,4 +1,5 @@
 import abc
+import typing
 
 from river import base
 
@@ -42,3 +43,20 @@ class AnomalyDetector(base.Estimator):
         normal observation.
 
         """
+
+    def predict_proba_one(self, x: dict) -> typing.Dict[base.typing.ClfTarget, float]:
+        """Predict the probability of each label for a dictionary of features `x`.
+
+                Parameters
+                ----------
+                x
+                    A dictionary of features.
+
+                Returns
+                -------
+                A dictionary that associates a probability which each label.
+
+                """
+
+        p = self.score_one(x=x)
+        return {True: p, False: 1.0 - p}
--- a/river/anomaly/base.py
+++ b/river/anomaly/base.py
@@ -43,20 +43,3 @@ class AnomalyDetector(base.Estimator):
         normal observation.
 
         """
-
-    def predict_proba_one(self, x: dict) -> typing.Dict[base.typing.ClfTarget, float]:
-        """Predict the probability of each label for a dictionary of features `x`.
-
-                Parameters
-                ----------
-                x
-                    A dictionary of features.
-
-                Returns
-                -------
-                A dictionary that associates a probability which each label.
-
-                """
-
-        p = self.score_one(x=x)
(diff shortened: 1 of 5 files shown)


river/anomaly/base.py
```diff
@@ -42,3 +43,20 @@ def score_one(self, x: dict) -> float:
...
+                    A dictionary of features.
+
+                Returns
+                -------
+                A dictionary that associates a probability which each label.
+
+                """
+
+        p = self.score_one(x=x)
+        return {True: p, False: 1.0 - p}
```
MaxHalford, line 62: I'm not comfortable with this. The output of `score_one` is not a probability, it's a score. Even if it's between 0 and 1, it's not a probability. I would rather stick with the notion of `score_one` and make it work with the rest of River.

river/evaluate/progressive_validation.py
```diff
@@ -48,7 +51,10 @@ def _progressive_validation(
         y_pred = preds.pop(i)
         if y_pred != {} and y_pred is not None:
             metric.update(y_true=y, y_pred=y_pred)
-        model.learn_one(x=x, y=y)
+        if model._supervised:
+            model.learn_one(x=x, y=y)
+        else:
+            model.learn_one(x=x)
```
MaxHalford, line 57: This merits some discussion. What about if we want to do `model.learn_one(x)` before the prediction/scoring? That's actually desirable in some cases. Do you have any thoughts on this?

river/anomaly/base.py
kulbachcedric: You're right. I removed it.

river/evaluate/progressive_validation.py
kulbachcedric: It could be that I'm wrong here, but isn't a progressive validation a test-then-train evaluation?
If we still want to have the `model.learn_one(x)` before the prediction/scoring, I would avert the problem at a higher level. When calling the `run` method of a `Track` we could pass a new argument for a different validation procedure.

MaxHalford: Agreed. I think I'm convinced :)

river/metrics/roc_auc.py
```diff
@@ -58,6 +60,11 @@ def __init__(self, n_thresholds=10, pos_val=True):
         self.thresholds[-1] += 1e-7
         self.cms = [confusion.ConfusionMatrix() for _ in range(n_thresholds)]
 
+    def works_with(self, model) -> bool:
+        return utils.inspect.isclassifier(model) or utils.inspect.isanomalydetector(
```
MaxHalford, line 64: A bit better:

```py
return super().works_with(model) or utils.inspect.isanomalydetector(model)
```

MaxHalford (approved):
Approved with a suggestion.

Can you also add an entry to `docs/releases/unreleased.md`?

Cheers!

MaxHalford (approved):

PR #799, opened 2021-12-11
Remove Lagger and TargetLagger

MaxHalford:
@AdilZouitine and @raphaelsty noticed that the functionality of the `feature_extraction.Lagger` and `feature_extraction.TargetLagger` can actually be performed by `feature_extraction.Agg` and `feature_extraction.TargetAgg` by passing `stats.Shift` to the `how` parameter. The only change to make is to allow `by` to be `None`. Thanks guys!

I took the docstring tests of the deleted classes and put them into a `test_agg.py` file.


river/feature_extraction/agg.py
```diff
@@ -126,26 +120,33 @@ class Agg(base.Transformer):
     """
 
     def __init__(
-        self, on: str, by: typing.Union[str, typing.List[str]], how: stats.Univariate
+        self,
+        on: str,
+        by: typing.Optional[typing.Union[str, typing.List[str]]],
```
AdilZouitine, line 125: `by` should  be set to `None` by default when using `typing.Optional` in annotation ?

river/feature_extraction/agg.py
```diff
@@ -240,19 +242,23 @@ class TargetAgg(base.SupervisedTransformer):
 
     def __init__(
         self,
-        by: typing.Union[str, typing.List[str]],
+        by: typing.Optional[typing.Union[str, typing.List[str]]],
```
AdilZouitine, line 245: Same https://github.com/online-ml/river/pull/799/files#r767171200

AdilZouitine (commented):
LGTM ! 😄

river/feature_extraction/agg.py
MaxHalford: I thought about that too. Not sure. I don't think that `typing.Optional` forces us to put a `None` by default. It just hints that we can pass a `None`. I rather have the user explicitly pass the `None` themselves, because not using a group by key is not a common behavior. Let me know if you disagree!

AdilZouitine: I understand your point of view and I agree with you.
Using `aggregate` without `by` is "unusual" behavior.

PR #821, opened 2022-01-14
Refactored feature renaming

damienwojtowicz:
Following [this discussion](https://github.com/online-ml/river/discussions/820), I implemeted the following changes to the existing `compose.Renamer` :

- Split `Renamer` into `Prefix` and `Suffix` transformers, doing what `Renamer` currently does.
- Added a `Rename` transformer that renames the features following a user-defined mapping. If a key in this mapping does not refer to an existing feature, it is silently ignored.
- Added unit tests for the three aforementioned new transformers.


MaxHalford (approved):
This looks good!

The only thing I would change is that `Rename` becomes `Renamer`, `Suffix` becomes `Suffixer`, and `Prefix` becomes `Prefixer`. It's just an unspoken convention we have.

damienwojtowicz:
Noted for the convention name. I did the changes as you said.

MaxHalford:
Good stuff @damienwojtowicz, thanks for this nice quality of life improvement!

PR #827, opened 2022-01-16
Feature/mini batch support

ColdTeapot273K:
See #818


river/compose/union.py
```diff
@@ -118,6 +120,35 @@ class TransformerUnion(base.Transformer):
...
+    Since we need a transformer with mini-batch support to demonstrate, we shall use the `StandardScaler`
+
+    >>> from river import compose
+    >>> from river import preprocessing
+
+    >>> t_1 = compose.Select("place") | preprocessing.StandardScaler()
+    >>> t_2 = compose.Select("revenue") | preprocessing.StandardScaler()
+
+    >>> agg = compose.TransformerUnion(t_1, t_2)
+    >>> agg.learn_many(pd.DataFrame(X))
```
ColdTeapot273K, line 143: This thing outputs a cumbersome dict representation, I wish there was a way to suppress the output or I could wrap it in a multi-line statement

river/preprocessing/one_hot.py
```diff
@@ -109,6 +112,93 @@ class OneHotEncoder(base.Transformer):
...
+    Keep in mind that ability for sparse transformations is limited in mini-batch case,
+    which might affect speed/memory footprint of your training loop.
+
+    Here's a non-sparse example:
+    >>> oh = river.preprocessing.OneHotEncoder(sparse=False)
+    >>> X_init = [{"c1": "Oranges", "c2": "Apples"}]
+    >>> oh = oh.learn_many(pd.DataFrame(X_init))
+    >>> oh = oh.learn_many(pd.DataFrame(X))
+
+    >>> oh.transform_many(X=pd.DataFrame(X))
```
ColdTeapot273K, line 158: The order of columns is just not idempotent when the test is called via `pytest` cli. Likely due to hashing in feature sets, which affects the column order at the `_encode_1d` level. Need help fixing this

river/preprocessing/one_hot.py
```diff
@@ -141,3 +231,60 @@ def transform_one(self, x, y=None):
                 oh[f"{i}_{xi}"] = 1
 
         return oh
+
+    # Mini-batch methods
+
+    @staticmethod
+    def _encode_1d(data, prefix, categories=None, sparse=False):
+        # REFLECTION: river often descends onto numpy level for computations
```
ColdTeapot273K, line 239: Kinda referes to the `transform_many` situation where multiple `pd.DataFrame` objects are created and written into a list  and then concatenated. It's ok for `pandas` library to be like that but for `river` it would be better to descend to `numpy` level and at the `_encode_1d` level output instead a tuple of `np.ndarray` + column name ordered collection (e.g. list). Then at the `transform_many` concatenate the arrays and create `pd.DataFrame` only then. 

Haven't come up with a pretty way to do this or just didn't have time. Help appreciated.

ColdTeapot273K: Nvm, got an idea, see commits below

river/base/transformer.py
```diff
@@ -32,6 +34,28 @@ def learn_one(self, x: dict, **kwargs) -> "Transformer":
         """
         return self
 
+    def learn_many(self, X: pd.DataFrame, **kwargs) -> "Transformer":
```
MaxHalford, line 37: It would be better here to add `MiniBatchTransformer` class (like `MiniBatchClassifier` and `MiniBatchRegressor`). Indeed, by adding this `learn_many` method to `Transformer`, it makes it look like each transformer has a `learn_many` method.

river/compose/select.py
```diff
@@ -97,6 +99,35 @@ class Select(base.Transformer):
...
+    {'x_1': 9.100117273476478, 'x_2': 8.892842952595291}
+    {'x_1': 10.94588485665605, 'x_2': 10.706797949691644}
+    {'x_1': 11.568718270819382, 'x_2': 8.347755330517664}
+    {'x_1': 9.687687278741082, 'x_2': 8.119188877752281}
+    {'x_1': 8.874551899214413, 'x_2': 10.021421152413449}
+
+    >>> import pandas as pd
+    >>> X = pd.DataFrame.from_dict(X)
+
+    You can then call `transform_many` to scale a mini-batch of features:
```
MaxHalford, line 121: `scale` isn't the correct verb

river/compose/select.py
```diff
@@ -105,6 +136,12 @@ def __init__(self, *keys: typing.Tuple[base.typing.FeatureName]):
     def transform_one(self, x):
         return {i: x[i] for i in self.keys}
 
+    def transform_many(self, X: pd.DataFrame):
+        # INFO: has either side-effects or doesn't have copy - choose your poison
+        # REFLECTION: worth adding `copy=True` parameter to the object constructor to allow both?
+        return X.loc[:, self.keys].copy()
```
MaxHalford, line 142: The expectation is that every `transform_` method in River is pure (with no side-effect). I would default to that behavior. We can think about performance later :)

river/compose/union.py
```diff
@@ -118,6 +120,35 @@ class TransformerUnion(base.Transformer):
     ...     ('# by place', count)
     ... )
 
+    Mini-batch example:
+
+    >>> X = [
+    ... {"place": 2, "revenue": 42},
```
MaxHalford, line 126: Add 4 spaces before each `dict`

river/compose/union.py
```diff
@@ -118,6 +120,35 @@ class TransformerUnion(base.Transformer):
...
+    ... {"place": 2, "revenue": 50},
+    ... ]
+
+    Since we need a transformer with mini-batch support to demonstrate, we shall use the `StandardScaler`
+
+    >>> from river import compose
+    >>> from river import preprocessing
+
+    >>> t_1 = compose.Select("place") | preprocessing.StandardScaler()
+    >>> t_2 = compose.Select("revenue") | preprocessing.StandardScaler()
```
MaxHalford, line 140: For the sake of example, I would use a different scaler for `t_2`

river/compose/union.py
MaxHalford: Yep, you can do `_ = agg.learn_many(pd.DataFrame(X))`

river/compose/union.py
```diff
@@ -118,6 +120,35 @@ class TransformerUnion(base.Transformer):
...
+    >>> agg = compose.TransformerUnion(t_1, t_2)
+    >>> agg.learn_many(pd.DataFrame(X))
+    >>> agg.transform_many(pd.DataFrame(X))
+        revenue  place
+    0  0.441250   -1.0
+    1 -1.197680    1.0
+    2 -0.693394    1.0
+    3  1.449823   -1.0
+    4 -0.945537    1.0
+    5  0.945537   -1.0
```
MaxHalford, line 151: The convention is to have an extra space before end of docstring

river/compose/union.py
```diff
@@ -233,3 +264,36 @@ def transform_one(self, x):
...
         )
+
+    # Mini-batch methods
+
+    def learn_many(self, X: pd.DataFrame, y: pd.Series = None):
+        """Update each transformer.
+
+        Parameters
+        ----------
+        x
```
MaxHalford, line 275: `X`, not `x`

river/compose/union.py
```diff
@@ -233,3 +264,36 @@ def transform_one(self, x):
...
+        """Passes the data through each transformer and packs the results together."""
+        # INFO: not the most optimal but at least it works
+        return pd.DataFrame(
+            dict(
+                collections.ChainMap(
+                    *(t.transform_many(X) for t in self.transformers.values())
+                )
+            ),
+            copy=False,
+        )
```
MaxHalford, line 299: Mmm intuitively I would just have used the [`pandas.concat` function](https://pandas.pydata.org/docs/reference/api/pandas.concat.html). What do you think?

river/preprocessing/one_hot.py
MaxHalford: I would say that you can add an extra step where you sort the dataframe by columns

river/preprocessing/one_hot.py
```diff
@@ -141,3 +231,65 @@ def transform_one(self, x, y=None):
                 oh[f"{i}_{xi}"] = 1
 
         return oh
+
+    # Mini-batch methods
+
+    @staticmethod
+    def _encode_1d(data, prefix, categories=None, sparse=False):
```
MaxHalford, line 238: Before I dive into the logic here, do you know why `pd.get_dummies` isn't sufficient here?

ColdTeapot273K: Albeit I don't remember the reasoning i had exactly... 😅 

Probably it was related to the fact that `pd.get_dummies` will encode every value, regardless of what the transformer has or hasn't seen. Which would require some potentially ugly filtering and whatnot. Also it would likely be slow and memory inefficient because a lot of `pandas` abstractions are carried around in the reference implementation (`pd.get_dummies`)

So here I've kinda reverse-engineered the reference implementation, thrown away unnecessary parts and tailored the rest to our application. Wanted to keep it as close to `numpy` as reasonable (like, `pd.Categorical` seemed too powerful to abandon)

river/compose/union.py
ColdTeapot273K: Sounds reasonable

river/compose/union.py
ColdTeapot273K: `StandardScaler` is the only scaler that supports `learn_many`

To keep it concise, could do instead:
```python
t_1 = compose.Select("place") 
t_2 = compose.Select("revenue") | preprocessing.StandardScaler()
```

What do you think?

river/compose/union.py
ColdTeapot273K: Btw do you know of any easy way to put code into these doctests? Like less manual/error-prone

river/compose/select.py
ColdTeapot273K: Fair :)

river/compose/select.py
ColdTeapot273K: Fair

river/compose/union.py
MaxHalford: I think it's possible to configure black to lint doctests, but I haven't looked into it

river/compose/union.py
MaxHalford: > StandardScaler is the only scaler that supports learn_many

*facepalm*

> What do you think?

Sure that's a good idea :)

river/base/transformer.py
ColdTeapot273K: Fair

ColdTeapot273K:
Unresolved yet:

https://github.com/online-ml/river/pull/827#discussion_r786018444

MaxHalford (approved):
Looks good to me!

MaxHalford:
Hey @ColdTeapot273K, can you give me push access to your fork? I've finished the PR for you, but I can't push my changes :)

ColdTeapot273K:
Thanks @MaxHalford! sent you an invite

MaxHalford:
Ok done! Thanks a lot for your work @ColdTeapot273K, it means a lot :)

PR #842, opened 2022-02-14
refactoring code with chain comparison

zjzh:
refactoring code with chain comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/river/datasets/synth/agrawal.py
+++ b/river/datasets/synth/agrawal.py
@@ -260,19 +260,19 @@ class Agrawal(base.SyntheticDataset):
     ):
         if age < 40:
             if (elevel == 0) or (elevel == 1):
-                return int((25000 <= salary) and (salary <= 75000))
+                return int(25000 <= salary <= 75000)
             else:
-                return int((50000 <= salary) and (salary <= 100000))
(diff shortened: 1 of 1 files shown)


MaxHalford:
Thanks for contributing. But this code isn't valid Python, there are syntax errors. I would suggest trying to run the code before submitting a pull request :)

JosiahWI:
Hi @zjzh. I like your intended change! But you've been making these refactoring changes on a very large number of repositories, and this is not the only one you've done incorrectly. Would you consider taking a little more time to make higher quality commits on fewer repositories?

raphaelsty:
I'm willing to bet it's a bot 😂

JosiahWI:
I've looked at a number of the PRs and I don't think it's a bot. There's both human error and human understanding involved, including for example creating a new Python file in a C operating system that is a dummy and does nothing useful. The user has a legitimate repo containing some homework code written in Swift, but I'm afraid he's cruisin' for a bruisin'.

MaxHalford:
The internet can be a weird place

zjzh:
Thanks for your comment. I am sorry for the syntax error. I just fix it.

PR #848, opened 2022-02-15
adapted robust_soft_learning_vector_quantization from scikit-multiflow

nejox:
Adapted robust_soft_learning_vector_quantization from scikit-multiflow to river.

### DIFF
--- a/river/prototype/__init__.py
+++ b/river/prototype/__init__.py
@@ -1,0 +1,5 @@
+"""Prototype-based learning."""
+
+from .robust_soft_learning_vector_quantization import RobustSoftLearningVectorQuantization
+
+__all__ = ["RobustSoftLearningVectorQuantization"]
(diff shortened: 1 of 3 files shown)


MaxHalford:
Hey, that looks like a lot of work! Thanks for the effort. Did you ask anyone to make sure this was desirable?

The rule with River is that we only accept pure online algorithms. In this case, the algorithm seems to work with mini-batches. For instance, in the `learn_one` method you cast `x` to a numpy array. This brings a lot of overhead and isn't acceptable.

I would suggest either revamping the implementation entirely to make use of dictionaries, or to contribute this to [river-extra](https://github.com/online-ml/river-extra).

nejox:
Hi, 
I will follow your suggestion and put this to river-extra.
As I do not know how much work it'll be to switch that to dictionaries. 
Thanks

PR #870, opened 2022-03-09
Added parameters to EDDM drift detector

TawabG:
PR related to Discussion https://github.com/online-ml/river/discussions/869

### DIFF
--- a/river/drift/eddm.py
+++ b/river/drift/eddm.py
@@ -40,6 +40,16 @@ class EDDM(DriftDetector):
 
     - 1: Error, $y \neq y'$
 
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


river/drift/eddm.py
```diff
@@ -40,6 +40,16 @@ class EDDM(DriftDetector):
 
     - 1: Error, $y \neq y'$
 
+    Parameters
+    ----------
+    eddm_min_num_instances
```
MaxHalford, line 45: I would remove the `eddm_` prefix, as it stutters with the name of the class :)

MaxHalford:
Thanks a lot @TawabG!

PR #872, opened 2022-03-09
updates to dbstream

vsoch:
So I ran into a few issues testing out DBSTREAM - if these changes are all wrong just tell me and I'll close the PR, but I wanted to at least understand what is going on.

1. The first has to do with the first call to predict when we don't have anything yet. We would get back an empty set of labels, and then it would try to take the max of an empty dict. This only happens on the very first call if it's predict so to fix I just do a check if there are labels (line 374).
2. Secondly, on the merge step of a cluster / micro cluster, it is currently required to have perfectly overlapping features (which I do not think has to be the case?) so there I thought I could use a .get() instead of direct index and that will default to a value of 0.0, so a feature that isn't overlapping just is added with no multiplier. I suspect this is developed on a dataset where the features are are known and consistent between entities (and maybe I'm not allowed to have this variable case but I thought I'd try!)
3. The third I started seeing OverflowError when I had added a lot of data. Since we are checking that the value is below some pretty small threshold as a criteria for removal, I just catch the two exceptions and continue, because such a huge number wouldn't be below the threshold anyway (so we should not remove!) I suspect I could tweak the fading factor to make it smaller... but then the model is likely to fail at some point given my choice, so I decided to try catching the error instead and handling it logically.

Anyhoo, this could be entirely incorrect to do in which case just tell me and I'll close the PR! Or if some of it is reasonable (but not all of it) tell me that and I can update. For my use case, I'm trying to use these cluster models with my same errors dataset, so I pass them through BagOfWords first (and maybe I'm not allowed to do that, and provide for this model, heh). 

The [resulting clusters](https://github.com/buildsi/spack-monitor-nlp-dev/tree/main/data/clusters/dbstream) are actually pretty nice (more homogeneous than just Kmeans, at least at first glance) and the exception(s) I can see are for example cluster 4 where there are a lot of single words (that I probably should have filtered out) so I hope there is some goodness in the PR that could get the model working for us, looking forward to chatting about it!
 

smastelini:
Nice proposal, @vsoch! It's nice to see the clustering model being used!

The proposed changes sound good to me (I haven't checked the code yet). 

Just to put my two cents in, for a long time I dream of using elements from the `stats` module to replace some operations in the clustering algorithms. 

For instance, variance calculation. I did something similar with the trees and before this change it was not uncommon to get overflow errors and even negative variance values (!). The Welford's algorithm solved all those issues. 

Not sure if this is the case with the cluster algos, but the third point you mentioned made me think about that.

vsoch:
@smastelini I do really like the idea of having algorithms use a common shared set of functions, when possible!  I think I've seen some of that already with stuffs coming out of utils? I'll keep this in mind as I test other clustering models (after we discuss changes here I'm going to try the same for [DenStream](https://riverml.xyz/latest/api/cluster/DenStream/).

MaxHalford:
@hoanganhngo610 feel free to give your thoughts too :)

hoanganhngo610:
@MaxHalford Thank you so much for pinging me and also a sincere thanks to @vsoch for contributing to DBStream. It feels so great that someone is actually using it, and seems to be useful. I will have a look today and merge if everything proves to be OK!

hoanganhngo610:
Thank you so much for your patience @vsoch . Regarding your questions, I would like to voice my discussion as follows:

1. Usually, with online machine learning algorithms, especially with online clustering, we really do not expect users to call `predict_one` right at the beginning when there has been no data points passing through the algorithm. However, it is true indeed that we should also take that exception into account. For me, I would suggest fixing them as follows:

```python
labels = self._generate_labels(weighted_adjacency_list)

if labels:
    self.n_clusters, self.clusters = self._generate_clusters_from_labels(labels)
    self.centers = {i: self.clusters[i].center for i in self.clusters.keys()}
else:
    self.n_clusters, self.clusters = 0, {}
    self.centers = {}
```

2. When we design clustering algorithms, we particularly want to emphasize that the data points that are fitted into the model has to have identical features (dictionary values). This can also be seen equivalent to two numpy arrays (which, indeed, has no features). However, in order for certain cases when the features of micro-clusters are not identical, I agree that we can have a work-around with the `.get()` function. @smastelini @MaxHalford would need a bit of your opinion on this, but for me, this seems totally fine!
3. To be honest, the datasets that we have been working on are large, in some sense (a csv file with a few million lines). However, each data point that comes into the model are basically one-dimensional array (or 1 x n array). This may be the reason why we have never seen any `OverflowError`. However, if you have encountered this, I have no problem fixing that by what you have been doing with the PR!

Once again, thank you so much for using River and spend time trying to make it more comprehensive and compatible with many more specific situations! I really appreciate it!

Pinging @smastelini @MaxHalford for some of your ideas on this, cause it may not be 100% perfect!

vsoch:
Thank you @hoanganhngo610 ! I've made and pushed the changes. I will note, however, that I'm not sure they were necessary because in the case of not having labels, those values are already set to empty/0 on init, and the centroid "calculation" at the end doesn't change anything (I checked). If there is some case where the values aren't 0 and there aren't labels we'd essentially be resetting the model (but maybe this is not possible). I do like your suggestion better because it is more clear what should happen for the developer user. But I want to double check that we can't have a case of empty labels and something defined for centers, etc. 

To give some background, the reason I call predict before learn for my [server](https://github.com/vsoch/django-river-ml/blob/739920e9556393202bbfca58dd4a0f1fd647ef50/django_river_ml/client.py#L176-L182) is because that's how I saw it done for [chantilly](https://github.com/online-ml/chantilly/blob/6a8e4c3605a830c952d1cc8fd6405e5b4c104633/chantilly/api.py#L286-L293). I like your update and do think it's a case that should be supported, e.g., for whenever a new model is added to a server we just can't know which of learn or predict will be called first.

> Once again, thank you so much for using River and spend time trying to make it more comprehensive and compatible with many more specific situations! I really appreciate it!

The thanks is all mine! I'm loving using river - I've gone from "Oy vey my matrices are too big to do anything nice with in the UI and I need increasingly larger nodes on our HPC to derive different representations" to "oh nice, my entire model with 230K points is like 200kb!" lol.

hoanganhngo610:
@vsoch Oh yes you are absolutely right. This means that the `else` part is extremely unnecessary, and I will delete that right now. Thank you so much for pointing that out!
It's extremely wonderful to see that you are loving River and it brings you a great deal of benefits. However, we have to admit that since we can not always check on how users are using our package, we have to sometimes assume that people are using it the way we assumes them to use. This will also mean that we have to be extremely explicit about it, which we are really trying to do from time to time.
Moreover, a little note that we are quite different from `chantilly` is that while `chantilly` prefers predicting before learning, we are designing `River`, or particularly, the clustering module with the spirit of `learn then predict`. This is also the common sense of online machine learning, generally speaking!

hoanganhngo610:
If there is no other objection, then I would proceed to merge this commit in by now! Thank you so much @vsoch. As discussed above, if you find out any problems with **Denstream**, please do not hesitate to let us know and then create a PR on it!

vsoch:
Just did the same test model generation with denstream - opened an issue with some questions! https://github.com/online-ml/river/issues/874 I'm thinking of just sticking with DBSTREAM since this other one seems like it adds complexity to deal with outliers, and since my data is error messages (that won't be outliers generally since they show up multiple places) I probably don't need that.

PR #873, opened 2022-03-10
Make AMRules show a default rule consequent

smastelini:
The `repr` method from AMRules' rules only shows the literals. Even when we don't pass an instance for debugging purposed (as in `debug_one`) it is nice to have an idea of what the rules output.

This PR makes the regression rules show the target mean as the default consequent of a rule. `debug_one` was slightly updated to explicitly show the prediction strategy.


MaxHalford (approved):
LGTM!

I would suggest reducing the number of significant decimals in the outputs, but it's up to you

smastelini:
Nice catch, done!

PR #885, opened 2022-03-18
Move black, flake8, and isort to dev extra env in setup.py

gbolmier:
Fixes the broken CI with `black` installation from the GitHub repo.

Check the first commit for the real changes of this PR, the second one only has changes from `black` being run on all files.


MaxHalford (approved):
Cool! I guess the ` ** ` -> `**` is due to a black update?

gbolmier:
Yes it is 👍

PR #888, opened 2022-03-20
dbstream: allow microcluster centoid update to have different terms

vsoch:
I was going to open an issue, but figured this was a small enough change that it would be easier to open a pull request to both discuss and show what I mean! I was testing dbstream out more, and hit this case where the microcluster center couldn't update because of a key index error. This issue is a lot like the previous changes that we made to allow for updates only given the overlapping keys. So the solution of course is to add the "if j in x" to mapping of j to the new i center, and then you are allowed to have different keys without throwing an error.

Of course I'm doing this almost at 1am so no promises that my brain is working correctly :laughing:

### DIFF
--- a/river/cluster/dbstream.py
+++ b/river/cluster/dbstream.py
@@ -196,11 +196,13 @@ class DBSTREAM(base.Clusterer):
                     )
                     + 1
                 )
+
+                # Update the center (i) with overlapping keys (j)
                 self.micro_clusters[i].center = {
                     j: self.micro_clusters[i].center[j]
                     + self._gaussian_neighborhood(x, self.micro_clusters[i].center)
                     * (x[j] - self.micro_clusters[i].center[j])
-                    for j in self.micro_clusters[i].center.keys()
+                    for j in self.micro_clusters[i].center.keys() if j in x
                 }
                 self.micro_clusters[i].last_update = self.time_stamp
 
--- a/river/cluster/dbstream.py
+++ b/river/cluster/dbstream.py
@@ -202,7 +202,8 @@ class DBSTREAM(base.Clusterer):
                     j: self.micro_clusters[i].center[j]
                     + self._gaussian_neighborhood(x, self.micro_clusters[i].center)
                     * (x[j] - self.micro_clusters[i].center[j])
-                    for j in self.micro_clusters[i].center.keys() if j in x
+                    for j in self.micro_clusters[i].center.keys()
+                    if j in x
                 }
                 self.micro_clusters[i].last_update = self.time_stamp
 


hoanganhngo610:
Thank you so much for you active participation in the clustering module of River, once again, @vsoch!

In this situation, instead of adding

```python
self.micro_clusters[i].center = {
    j: self.micro_clusters[i].center[j]
    + self._gaussian_neighborhood(x, self.micro_clusters[i].center)
    * (x[j] - self.micro_clusters[i].center[j])
    for j in self.micro_clusters[i].center.keys()
    if j in x
    }
```

I would suggest using the operator `&`, which makes the code a bit cleaner and also makes a bit more sense: 

```python
self.micro_clusters[i].center = {
    j: self.micro_clusters[i].center[j]
    + self._gaussian_neighborhood(x, self.micro_clusters[i].center)
    * (x[j] - self.micro_clusters[i].center[j])
    for j in (self.micro_clusters[i].center.keys() & x.keys())
    }
```

What do you think?

vsoch:
> What do you think?

I don't think that's exactly what we want (meaning not the same thing), because x might have keys, but perhaps not the same set (and it would hit the same index error).

hoanganhngo610:
@vsoch Thank you so much! When you have the tests done, do not hesitate to let us know the results! 
I would also tag @MaxHalford here for his reference over this, that I believe would be quite useful 😉

vsoch:
Okay doing some testing now - will have some metrics to report shortly.

vsoch:
Okay - so I ran both ways (in the same run), and timed the operation, because I suspected needing to take the intersection of two sets of keys would be slower than just checking for each membership in N. The suggested change to keys() was consistently slower for the same operation, across the board:

For a teeny tiny calculation (early in training so probably a small cluster) it looks pretty trivial, e.g.,

```
# this is the suggested change to use keys()
0.00014400482177734375 seconds for keys
# this is my current implementation
0.00013256072998046875 seconds for set memberships
```
A little bit later:

```
# suggested change
5.316734313964844e-05 seconds for keys
# current
4.649162292480469e-05 seconds for set memberships

# suggested change
7.224082946777344e-05 seconds for keys
# current
6.508827209472656e-05 seconds for set memberships
```
But for a larger comparison you can see the differences adding up (or the cost of that keys operation):

```
# this is the suggested change to use keys()
49.017354011535645 seconds for keys
# this is my current implementation
28.747771978378296 seconds for set memberships
```

And then if I sum up total time for each (the entire training)

```python
# suggested change
sum(model["DBSTREAM"].set_diffs)
190.98662066459656

# current
sum(model["DBSTREAM"].membership_diffs)
116.87801384925842
```
So mean and standard deviation for each:

```python
# with suggested change
statistics.mean(model["DBSTREAM"].set_diffs)
0.09436097858922755

statistics.stdev(model["DBSTREAM"].set_diffs)
1.3222071001040605

# current implementation
statistics.mean(model["DBSTREAM"].membership_diffs)
0.05774605427334902

statistics.stdev(model["DBSTREAM"].membership_diffs)
0.7387732902344243
```

So (in my mind) the current implementation is clearly superior if we can get the same result for less time, and the difference is non-trivial. Are there factors I'm not considering? Let me know if you want me to test anything else!

hoanganhngo610:
@vsoch Totally makes sense to me now! In that case, I would approve this PR and merge it in, given that there is no other opposition! :D

hoanganhngo610 (approved):
Changes are perfectly good to merge!

vsoch:
Thank you @hoanganhngo610 !

Issue #891, opened 2022-03-23
Question about centroid changes


vsoch:
heyo! So this might be better suited for chantilly, but it's generally about river so I hope it's okay to put here. I'm creating a server that has a cluster model on the backend to handle clustering types of errors. I was planning on using a cluster model and then assigning new errors (saved as objects in the database) to clusters, but I realize if I have a preference for models with a changing number of centers then my assignments would also likely change over time (and not be valid). This makes the idea of saving the cluster id with the error object not a great idea. But then I was thinking - as long as we have some kind of ability for a model to output changes (e.g.,:

- Cluster 8 no longer exists
- Clusters 10 and 12 were merged into 10
- Cluster 5 was split into 5 and 101

Then it could be feasible to act on those events, and in the case of the above:

- Remove the assignment of any errors to cluster 8 (reclassify if it's computationally easy)
- find objects assigned to 10 and 12, assign all to 10
- Remove the assignment of 5, and reclassify if feasible

So my question - how do we integrate this functionality into river, at least exposing enough for a wrapper of some kind to assess changes? Or is it just a bad design idea on my part?

vsoch:
Here's an idea - what if I could use a KNN model alongside this one, and then for the sliding window here https://github.com/online-ml/river/blob/9ea7707636e825a374cf053d149e3946c746ce2d/river/neighbors/base_neighbors.py#L101-L102 keep a vector of identifiers (of my own choosing) that I could relate back to my own objects or data, and then have a flavor of KNN that returns those full centroids / identifiers instead? It looks like right now the KNN models here always attempt to classify and none of them actually give back a group of meaningful centroids (e.g., here are data points used in the window that you provided earlier, do as  you please with them!) Would this be of interest to others? (it's of interest to me!)

vsoch:
And follow up question (I implemented the model above, am testing) can we make knn flexible to different numbers of features?  E.g., here:

```python
if self._n_features != get_dimensions(x)[1]:
            raise ValueError(
                "Inconsistent number of features in X: {}, previously observed {}.".format(
                    get_dimensions(x)[1], self._n_features
                )
            )
```

smastelini:
Hi @vsoch, indeed the current KNN have some limitations. A direct solution for different sets of features is to always rely on exhaustive search. 

I'm currently working on ways of improve online knn as part of my research internship. Hopefully I should have results soon.

MaxHalford:
[...]

PR #895, opened 2022-03-28
refactoring neighbors models to use simple collections queue

vsoch:
This is a follow up to discussion in https://github.com/online-ml/river/issues/891#issuecomment-1080993289.

The current implementation of nearest neighbors was done for a performance tweak, but it is limited because we cannot be flexible to add slightly varying features (the vector size for X is fixed and you get an error if you vary from the first one) or to add additional metadata about a point in the window such as a UID that can be used to look up a point later. From an online ML standpoint (as I understand it) we should optimize for this kind of flexibility over a small performance tweak (but coming from the world of HPC I totally get the performance bit!)

This refactor takes the base logic from creme.neighbors, and adds some additional features, 

1. a minimum distance to consider adding a new datum to the window, ensuring we have a more variable window that does better at learning. The default for my model was 0.05 because I had ~250K points with quite a bit of repetition so adding redundancy was expensive and led to bad results, but I chose a default of 0.0 here so the default miimcs what you would expect. 
2. Given the window can change (and classes within it) I realized that the prediction dict would potentially have a lot of extra "no longer existing in the window" classes. So I added a class cleanup function for classification to iterate over data in the window and ensure the classes known to the class accurately reflect the current window. The default is False so this will not run (and the model maintains all memory of classes it has seen and returns 0 probability of the class) but it can be set to True to always run after learn (potentially at the cost of performance) or can be run on demand (e.g., I can see an online-ml server setting it to False, and then loading and running once a day or something like that.
3. optional support to add a UID, which is just another variable in the list added to the queue! This I've found incredibly useful  in production cases where you do not just want a prediction back, but you want to look more into the closest points, metadata wise. The data in the window should be useful beyond having x , y, and class, and allowing an additional identifier allows the implementer to put more information elsewhere.

For the reasons in 3, I renamed the BaseNeighbors class to be KNeighbors because people can use it as a very simple model to return the neighbors verbatim. 

This is likely a start (and I have not run tests locally) so we can discuss further changes that need to be done, a strategy for doing them, and what features / additional classes of the old implementation we want to preserve. I had this on another branch, but for some reason running pre-commit made changes outside of my changes (perhaps a different version of black or flake8 or something?) so I created a fresh clone and added the changed files verbatim.


smastelini:
Just playing the devil's advocate here. But what if, we just keep the standard x and y paradigm. But, only but, someone clever does something like that:

```python
...

# learn_one
nn.append((x, uid, "potato", "foo", myobj), y)
```

And we make this implementation receive a dist function as a parameter, which defaults to `None` (which means, Minkowski with p=2). Hence, the user can create a wrapper that simply does something like:

```python
def wrapper(dist_function):
    def wrapped_func(x1, x2, dist_func=dist_func):
        return dist_func(x1[0], x2[0])
    return wrapped_func
```

And voila! Now, much more than an id can be stored and we keep the same API standards and also the inherent simplicity of k-NN (which I made disappear during the merger -- my bad).

By the way, for those that don't now that, this is a prime example of a typical Brazilian strategy called "gambiarra"!

smastelini:
Btw, I like the idea of an optional filter for the addition of new examples!

vsoch:
> Just playing the devil's advocate here. But what if, we just keep the standard x and y paradigm. And someone clever does something like that:

So the idea would be that x is a tuple, with x at 0 always being the data and "the rest" gets added? That would work, but it's different from the typical river design, and I think it would be confusing. I added uid to the base model because I think a simple identifier can handle all use cases of "putting extra data somewhere else." Otherwise we are encouraging people to dump a lot of extra stuff in the model which can get larger faster (especially for a use case like mine). 

> And we make this implementation receive a dist function as a parameter, which defaults to None (which means, Minkowski with p=2). Hence, the user can create a wrapper that simply does something like:

Oh yeah we can totally take the distance function as a parameter - I like that idea a lot. We would remove the check for the range of values that can return back since arguably it could return anything.

vsoch:
Let's wait to hear from @MaxHalford et al - I do think it's somewhat important to not engineer features before they are asked for, especially given that we should be making some effort to maintain the original interface.

vsoch:
> Btw, I like the idea of an optional filter for the addition of new examples!

Haha thanks! That came directly out of my dinosaur brain :P

smastelini:
> So the idea would be that x is a tuple, with x at 0 always being the data and "the rest" gets added? That would work, but it's different from the typical river design, and I think it would be confusing. I added uid to the base model because I think a simple identifier can handle all use cases of "putting extra data somewhere else." Otherwise we are encouraging people to dump a lot of extra stuff in the model which can get larger faster (especially for a use case like mine).

Not at all! The idea is to show that if someone really needs to add some extra stuff there are ways of doing that without changing the current API :D

It was just a tricky example of a workaround. I expect most of people to simply use KNN for prediction with just `x` and `y`.

smastelini:
Although the UID sounds nice, I do believe it deviates from our current design, just like using numpy arrays in k-NN did in my previous implementation. But let's discuss that, maybe I'm not looking at the whole picture!

vsoch:
I suspect tests will fail - I might need some help / pointers for how to work on them because the setup isn't obvious for a new contributor.

vsoch:
> Although the UID sounds nice, I do believe it deviates from our current design, just like using numpy arrays in k-NN did in my previous implementation. But let's discuss that, maybe I'm not looking at the whole picture!

To kind of zoom out "big picture" across online ML, I think many of the models assume that you are just interested in putting data in and getting it out. And maybe for some that's true, but for more research oriented cases I want more understanding than that. E.g., imagine if you are using KNN for some simple image features. Getting another set of features back isn't super useful. But getting back the identifier for those images so you can do further kinds of comparison (likely image processing) or a more even simple case - just wanting to show the person the similar images (think radiology decision support) is something I think would be commonly wanted. Supporting this basic idea to keep an identifier alongside the points that are found to be most similar / chosen (or whatever the algorithm does) I think makes sense for this kind of design, otherwise the person using river is always going to have to implement their own custom model to support that.

smastelini:
I see your point. However, I am now thinking about the whole ecosystem, for instance, how do pipelines (transformers, and so on) play with that? `**kwargs`, maybe? Can't something like what you propose be handled elsewhere? I am not against some kind of temporal buffer that can handle any kind of data. Actually, that's exactly my current research goal. I'm just still not entirely convinced about coupling that in k-NN (I mean, kNNClassifier or Regressor). But I do see some use cases for that, indeed. Let's hear the opinion of the others :)

MaxHalford:
Hey there.

The River API should be consistent for each class inheriting from `base.Estimator`. But we shouldn't be too overzealous regarding other utility classes. For instance, take a look at the stuff in the `misc` module.

Here I would say that the `NearestNeighbors` class should be as flexible as possible. It should take as input a single parameter called `x`. The structure of `x` depends on the task. You could use tuples, dataclasses, dicts, whatever. `NearestNeighbors` should be agnostic of the learning tasks where it used. 

Does that make sense?

river/neighbors/base_neighbors.py
```diff
@@ -1,233 +1,201 @@
...
+        # A minimum distance (similarity) to determine adding to window
+        # The model will perform better with a more diverse window
+        self.min_distance_keep = min_distance_keep
 
-        TypeError: If the buffer type is altered by the user, or is not
-        correctly initialized, a TypeError may be raised.
+        # Save a unique identifier for each neighbor to return later
+        self.save_uid = save_uid
 
+    def add_to_window(self, x, y, uid=None):
```
MaxHalford, line 55: I would call this `append`.

smastelini: +1 to that.

This is what I meant by sending the example with the kDTree from sklearn, maybe I did not express myself effectively (English is not my native language hahah). Although it only takes arrays (X), you can anything with that. It is meant to perform search queries in different flavors. 


I think that `NearestNeighbors` could indeed become part of a collection of sorts, along with (Ordered) Sliding Windows, some types of (biased) reservoir buffers, and stuff like that. Just thinking out loud, we could even have something like `NearestNeighbors` which is agnostic to the data ingestion strategy. For instance:

- FIFO
- Reservoir sampling
- Ordered Values
- Set-esque addition

As long as the underlying data buffers have a `__getitem__` we are good to go.

vsoch: > Does that make sense?

Yes totally! 

And I love your ideas @smastelini - it's not your English being bad, I'm just very dense, haha - I didn't think through the idea that the base class could be general (but the actual classes that the user interacts with could still be consistent with what River generally provides). 

I'll get in some new updates for this PR tomorrow / later today - after I go back to sleep and wake up again :P

vsoch:
@smastelini  @MaxHalford I just added the first part of our discussion, making the NearestNeighbor model more flexible to variable features! 

https://github.com/online-ml/river/pull/895/commits/c8a1cb7a720f3bb70e9b32894acf2db83bfc1ecc

Since this would only make sense if someone is getting the full neighbor back (e.g., and not a regression or classification value) it's now only relevant for the KNeighbors model. I also added a full example in the docstring to add a full UID since people might use that class directly (it's the one I'll be using!). If other KNN derived models want to use this feature they will also need to expose the extra metadata in some way.

My proposal is to discuss this setup, and then we can followup with custom collections possibly in a different PR, just so this one doesn't get too big!

Also, I'm having trouble with pre-commit because when I run it, it changes files extensively outside of my PR directory. I tried running isort / flake8 / black on just the directory but it looks like I still have issues. Do you want me to run it anyway and then make the PR harder to review? I can also accept the lint errors and run it at the end.

hoanganhngo610:
@vsoch Looking at the tests it seems like the code does not have many problems with the code style, you only need to run the `isort` command on the files that you suggested changes and all will be fine. In fact, the reason why it's failing is from the `pytest`, where I think it comes the docstrings that would be needed to change and some assertion errors. Will try to look a bit to it today or tomorrow if you don't mind!
I would also agree that the custom collections should be discussed and set up in a later PR, since this one has already been quite big.
Overall, thank you so much for such a wonderful effort!

vsoch:
> @vsoch Looking at the tests it seems like the code does not have many problems with the code style, you only need to run the isort command on the files that you suggested changes and all will be fine. 

I did this before and there were either changes across the code base. I'll run pre-commit and push to show you - it's going to mess up the PR.

vsoch:
Here is running pre-commit - it does changes across river for files I did not touch. https://github.com/online-ml/river/pull/895/commits/e5e5a3f4524a6cc33e21edeb6cbf23b18ee639e2

hoanganhngo610:
From what I see, basically, what I am seeing is that they basically generate two spaces around any `**` operation. However, this should be extremely weird, since I have never experienced this before. @MaxHalford Do you have any ideas on how to fix this, or should this be merged to the main branch without any concerns?

vsoch:
I do think it's version differences (it may not be the latest) between the PR and my local - I did try installing the "latest" but latest is subjective depending on where you install from. I will try again.

vsoch:
oh I didn't know the docstring examples have to match exactly! That output is correct I just didn't want to muddle the docstrings. Let me see if I can reduce the example and print what I see.

vsoch:
okay I found the tests - KNeighbors needs to be skipped because it returns the full neighbors tuples (and not a dict) and regressors for the same reason - it returns a single value. I noticed other regressors in the ignore list so I think this is what other models are doing (and correct to do). I'll check out errors from this next run and see if I know how to fix.

vsoch:
yay I figured out the tests fixes!! _does a tiny dance_ :t-rex:

vsoch:
> Overall some things still feel a bit weird to me. I wouldn't have a NearestNeighbors and a KNeighbors class. I would drop KNeighbors and expose NearestNeighbors.

Sure, I can try to combine them - conceptually I was trying to mirror the original design - there was a nearest neighbors class and then a "Base" for the models. I think the idea was that you could use the NearestNeighbors outside of the context of a model?

Finishing up another PR for another project - will jump back on this after!

vsoch:
@MaxHalford:

- KNeighbors is now one class, NearestNeighbors
- It takes a custom distance function (optional) and set of kwargs for it. I added two functions to either set the default back, or set a custom function. Both are a courtesy to the user - they might want to change the distance function on the fly (or otherwise after model init) and they shouldn't need to look at the source code and set instance attributes, there should be an obvious function. For the reset to the default, we do this to still be able to do the sanity checks on the variable p for minkowski.

This is the one bit I need more detail on:

> I believe NearestNeighbors should be even more generic than in it. You should be able to pass any kind of object (not just x or y)

Right now we do allow adding any kind of extra metadata to store with the tuple - e.g., this example:

```python
    >>> model = neighbors.NearestNeighbors(window_size=50)
    >>> for x, y in dataset.take(10):
    ...     model = model.learn_one(x, y, extra=[str(uuid.uuid4())])
```
and that means I could add my unique identifier, or another lookup of associated metadata. The only thing we need to maintain is that x is first, y is second, and the distance (will be) last after whatever custom metadata is provided! But I think you are wanting something more general for x and y themselves? I don't totally follow, so if you want to give me more detail/example I can try, or you can definitely just add to the PR yourself since you probably know exactly what you have in mind. Either works for me!

smastelini (changes requested):
Hi @vsoch, I forgot to post my review comments. I believe what @MaxHalford mentions about a generic implementation is also detailed within one of my comments.

river/neighbors/base_neighbors.py
```diff
@@ -1,233 +1,222 @@
...
+        self.p = p
+        self.window = collections.deque(maxlen=window_size)
 
-        TypeError: If the buffer type is altered by the user, or is not
-        correctly initialized, a TypeError may be raised.
+        # A minimum distance (similarity) to determine adding to window
+        # The model will perform better with a more diverse window
+        self.min_distance_keep = min_distance_keep
 
+    def append(self, x, y=None, extra: [Tuple, list] = None):
```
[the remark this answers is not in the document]
smastelini: Regarding passing parameters to the dist function, I believe the way to go is using `functools.partial`. We have used that in other situations in our codebase.

For example:

```python
dist_func=functools.partial(utils.minkowski_distance, p=2)
```

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,174 @@
...
+        Returns:
+            self
+        """
+        self.classes = set()
+        self._nn = NearestNeighbors(
+            window_size=self.window_size,
+            p=self.p,
+            min_distance_keep=self.min_distance_keep,
+        )
+        return self
```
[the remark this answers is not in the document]
vsoch: I think this review was on an old version? self._nn is no longer part of the class.

river/neighbors/base_neighbors.py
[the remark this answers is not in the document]
smastelini: By default:

`dist_func=None`

and then 

```python
# __init__
self.dist_func = dist_func if dist_func is not None else functools.partial(utils.minkowski_distance, p=2)

```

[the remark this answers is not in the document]
vsoch: > In this case, the p parameter and Minkowski distance would be dropped too, and replaced by something like dist_func. Such distance function, besides calculating the distances (duh), would also be in task of selecting what to use from the item.

I also already added this - do you want to review the current version?

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,174 @@
...
 
     Notes
     -----
-    This estimator is not optimal for a mixture of categorical and numerical
-    features. This implementation treats all features from a given stream as
-    numerical.
+    See the KNeighbors documentation for details about the base model. Note that
+    since the window is moving and we keep track of all classes that are added
+    at some point, a class might be returned in a result (with a value of 0)
+    if it is no longer in the window. You can call model.class_cleanup() if
```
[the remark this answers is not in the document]
vsoch: It's not private because the user might choose to have it turned off for automated run, and call on their own.

river/neighbors/knn_regressor.py
```diff
@@ -139,40 +101,29 @@ def predict_one(self, x):
...
-        elif self.aggregation_method == self._MEDIAN:
+
+        if self.aggregation_method == self._MEDIAN:
             return np.median(neighbor_vals)
-        else:  # weighted mean
-            return sum(y / d for y, d in zip(neighbor_vals, dists)) / sum(
-                1 / d for d in dists
-            )
+
+        # weighted mean based on disance
```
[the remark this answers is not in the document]
vsoch: fixed!

river/neighbors/base_neighbors.py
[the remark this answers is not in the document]
vsoch: I'm not familiar with functools partial, and I'm not totally following the "allow any kind of thing" because it would be different than the other river classes. I think this is probably a good splitting point for @MaxHalford to take over and do tweaks to his liking.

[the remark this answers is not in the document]
vsoch: I see. How would that account for:

1. handling custom values of p for the minkowski metric set by the user, and
2. allowing the user to change it on the fly easily OR set the default again easily?

What benefit does functools partial have over just saving a pointer to the function and using it?

river/neighbors/knn_classifier.py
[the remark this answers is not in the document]
smastelini: My bad, I reviewed this morning and forgot to click on that submit button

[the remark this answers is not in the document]
smastelini: Also, `base.Estimator` already has a general purpose `reset` method which is the preferred choice

vsoch:
ping others in this PR because I think we need to discuss some important details about open source contribution:

https://github.com/vsoch/river/pull/1

It's really important when you have a new contributor to be as supportive as possible, and champion their work. What this "refactor" pr does by way of entirely deleting and adding new files is wipe my contribution from the git history. I also find it very confusing because:

1. my custom distance function is gone, which is something we talked about in detail yesterday
2. the effort to consolidate into one class (and separate the nearest neighbors function from the base classifier) is also removed
3. the self.p is again hard coded to the class, which doesn't make sense given the custom distance function.
4. Speaking of, the custom distance function is just removed.

To be totally candid I find this refactor really upsetting. Not only are a lot of the features we discussed (and I implemented yesterday in respond to review) gone, this sends a strong message that not only is my work or contribution in the history not valued, but someone else is taking the snippets I wrote and re-writing them. If this is the kind of standard you set for contributors, and especially new contributors, it's really a deterrent for wanting to be a part of the project. I've spent quite a bit of time here and I want credit for my contributions. Deleting and making entire new files (especially when there isn't a strong function reason and it mirrors the old structure) is something you should be careful about in the future. As a new contributor if this is the only option you are giving me I will not choose to contribute again, because this is a red flag.

Sorry for the rant, I'm feeling really down after seeing the PR.

vsoch:
Gotcha! Sorry for getting upset - we had good conversation there - https://github.com/vsoch/river/pull/1#issuecomment-1084757244. I've done PRs to contributors PRs on the repos I maintain and it's usually "let's merge this in" as opposed to "I want to show you what I mean" but I misunderstood and it's the latter. I'll spend some more time today to make updates here from what you showed me.

vsoch:
Phew green again! Thank you @MaxHalford for the PR to help show me the item idea! I did my best to refactor taking into account the suggestions and we can have further discussion.

- the original BaseKNN and NearestNeighbors are restored. I added more functionality to the NearestNeighbor class because it should work without a hitch and requiring customization. E.g., in the suggested code it required the function to be provided, which means it doesn't work by itself, it's dead in the water. As a new user I'd want to be able to import NearestNeighbors for my own use and have it work without needing to think about distance functions. To fix this, I moved the setting of the default function up to it (Minkowski distance) given that the function is not provided. I also maintained the design to provide exposing the variable p as general default function kwargs instead of hard coding p on the class (and then it might not be used). I tried to use functools.partial as suggested by @smastelini but I think I still like the original way where I allowed using (or changing) the kwargs separately. We probably will disagree here but maybe there is some middle ground.
- I maintained keeping extra metadata separate from x,y (the item). The reason is because we need control over the ordering of keeping things in the window. If someone provided, for example, (x, extra, y) it would break everything. It also doesn't seem like the right design to store metadata and data in separate places, and then need to update both. That sounds like bugs waiting to happen!
- I changed the default input to be "item," although I don't like this design because it increases the complexity of what a person needs to understand to write a custom distance function. They cannot just use distance(a,b), they have to take the first index of a (the items index 0 that is expected to have x). To help this, I clearly wrote the default_distance_function at the top to explicitly show this as an example (and this default is used in the NearestNeighbors when distance is set). Minimally if someone is struggling we can point them to that and it's pretty clear.
- softmax: I think it's more sane to return the original default, and a normalized value, so I made True the default instead of False
- I really like the "cleanup every" idea, but I didn't add it here because I think we should discuss this idea across the library! We should look over all the models that can support some kind of "cleanup" and "rate of cleanup" or even more generally, some kind of hook / callback and think of what would be the common terminology (function args or interactions) to enable that? Not just for KNN, but across models?
- I maintained find_nearest because a 'query' or 'search' is a really general term that isn't obvious to a new person (or someone not familiar with the other libraries). But "find_nearest" is explicitly clear. I want to choose clarity here.

So (my review) on these changes, and this is from the perspective of someone that wants to use river, or wants to extend river, or wants to develop for river but is not familiar with it:

1. I think returning to the original design of a NearestNeighbors class to be used at self.nn is a good choice, and worth the extra work to undo the original change (it's been about 4 hours this morning to do these new changes).
2. I think using "item" over "x,y" is neat and cool (it's simpler and neat/Pythonic to have just one thing) but since we cannot hide the interaction from the user (e.g., if they write a distance function it needs to accept items and not x) it's a move in the wrong direction, because it makes river harder to use / extend. It's not an issue if people use the default distance function, but I suspect will be a pain point (we will see more issues opened because there is confusion).
3. I think I'm okay with any means to derive the distance func, but I do think it's not a great design to hard code p into the class and then if we do it the other way, not validate the kwargs. I think both aren't perfect, but I am feeling less opinionated about this and ok to compromise.

So to fix these, my suggestion is restore the original design of x,y The features should be provided as x, and the class y, and we can remove the complexity of the distance function definition. It's less "neat" to not have the item, but I think it's more expected, more clear, and therefore a better design. But if y'all really like this item thing and are okay with potential issues coming in about why a distance function with x,y doesn't work, then I'd be okay to leave.

smastelini:
From the start I kept giving hypothetical examples but never showed anything concrete. Let me show a possible user case that would illustrate how I envision the design of nearest neighbors. We can converge to that design if needed in the future.

This is how my item could look like:

```python

@dataclass
class Entry:
    name: str
    rate: float
    age: int
    im_out_of_ideas: bool

   def foo(self):
       # Do something important here

   def bar(self):
       # Do another important thing here

```

Then I can define that my distance function as:

```python
def my_dist(a: Entry, b: Entry):
    return (a.rate - b.rate) ** 2
```

Or I could measure the edit distance between the `name` field. Who knows? (well, I should know, as the designer of `Entry`)

I can use default parameter values for extra stuff. For instance:

```python
def my_dist(a: Entry, b: Entry, coefficient: float = 1.0):
    return coefficient * (a.rate - b.rate) ** 2
```

If for some reason I want to change the default parameter value and really don't want to rewrite the distance function, or I want to change the parameter in runtime, then I can rely on `functools.partial`. There are other workarounds, but this is the pythonic way of doing that:

```python
other_param = functools.partial(my_dist, coefficient=42.0)
```

Regardless of the strategy for the distance function, in the end I would use: `buffer = NearestNeighbor(window_size=1000, dist_func=my_dist)` and store my data class `Entry` using FIFO.

In my opinion it is not a matter of whether the user must define a custom function, but whether or not they could if they wanted to do that. Users could even store meta-data in the item itself, as item can be anything. So, I'm cool with other designs if that's what most people want, but still wanted to voice my opinion. I keep thinking this kind of design is really clever, transparent, and simple.

vsoch:
@smastelini I'm almost includes to just hard code a default for the minkowski, and if a user wants to change that - they can use the custom distance function! That will remove a lot of the current checks / complexity that I was adding with the distance_func_kwargs to "make it work!"

smastelini:
What about having something like MinkoskiNN that does everything you say and inherits from NearestNeighbors? (I might be messing up the names as I'm on my phone, so just follow what you're using)

Therefore we keep the base class flexible and simple enough for generic/power-user cases and also provide a simple entry for "everyday usage". 

Also, don't take my example too seriously. It's just an illustration. But a functional version of it could indeed be a guideline of sorts.

vsoch:
That could work - so the idea would be that most people use MinkowskiNN, but NearestNeighbors is there for an empty class to provide general functions? And if the user doesn't provide a distance function there we raise an error?

smastelini:
Yes! I think it's even better to just make dist_func a parameter without a default value. 

Therefore we have a grab and go solution and something that can be tweaked for different item/distance flavors. 

Also, we also have a template for other search-based structures/buffers follow.

vsoch:
okay sounds good! I'll work on these new changes.

vsoch:
Updates from https://github.com/online-ml/river/pull/895/commits/f61ffb92bc2f5ee764e1e02806a0fef2ddb091c4!

- We now have neighbors.py to put `NearestNeighbors` and `MinkowskiNeighbors` in, the idea being that we still have this blank "vanilla" sort of class (NearestNeighbors) and most folks (and the neighbors metrics) that want something that you can instantiate and go (without needing to think about your distance function) can take off running. I also tested the case provided by @smastelini and it totally worked! So this is a nice middle ground I think. It also is really nice because the MinkowskiNeighbors is appropriate to hold the value of p, which we were awkwardly moving around before.

river/neighbors/__init__.py
```diff
@@ -4,9 +4,15 @@
...
+from .base_neighbors import BaseKNN
 from .knn_classifier import KNNClassifier
 from .knn_regressor import KNNRegressor
-from .sam_knn import SAMKNNClassifier
+from .neighbors import MinkowskiNeighbors, NearestNeighbors
 
-__all__ = ["KNNADWINClassifier", "KNNClassifier", "KNNRegressor", "SAMKNNClassifier"]
+__all__ = [
+    "BaseKNN",
+    "MinkowskiNeighbors",
```
MaxHalford, line 14: I don't think we want to make these classes public, right?

MaxHalford:
> softmax: I think it's more sane to return the original default, and a normalized value, so I made True the default instead of False

Be aware that scikit-learn's implementation doesn't use softmax. In fact, the canonical way in the litterature to do KNN normalization is to divide by the total. Not to use softmax.

> I really like the "cleanup every" idea, but I didn't add it here because I think we should discuss this idea across the library! We should look over all the models that can support some kind of "cleanup" and "rate of cleanup" or even more generally, some kind of hook / callback and think of what would be the common terminology (function args or interactions) to enable that? Not just for KNN, but across models?

That's a good idea. But I don't see any obvious way to implement it in a generic way. This is very model specific. I think it's a good choice to keep it for KNN, so at least we keep a trace of it.

river/neighbors/__init__.py
vsoch: I think baseKNN yes, that's the one I want to use, MinkowskiNeighbors probably not, at least until someone complains and says they want to use it and don't understand why they can't. I am more inclined to be conservative and expose both.

MaxHalford: Don't you mean conservative and not expose both? I think the less we expose the better, no? I'm not sure I follow

vsoch:
okay updates for https://github.com/online-ml/river/pull/895/commits/4f5206b4059674e04838e0770b41d76154cc7c13:

- neighbors functions no longer publicly exposed
- softmax default is False
- cleanup_every to replace boolean argument to run a cleanup

river/neighbors/__init__.py
vsoch: I mean conservative in the "develop so users are less likely to open issues on something" but I'm good with removing those from being exposed - see most recent commit!

vsoch:
yay green first try this time! :partying_face:

MaxHalford:
Good job. However some things are still confusing to me.

What's the point of having `BaseKNN` public? What use cases are you thinking of? I still don't see what value `BaseKNN` is bringing. In particular, I find it weird that it inherits from `base.Classifier`, and then `KNNRegressor` inherits from it. That means that `KNNRegressor` is itself a classifier.

Also don't we want to make `NearestNeighbors` public?

Finally, with regards to `MinkowskiNeighbors`, I think it's superfluous. You could simply pass a default distance function when you initialize the `nn` attribute. No need to have a class.

There's also some documentation nitpicks but I can send a PR to fix those.

vsoch:
> What's the point of having BaseKNN public? What use cases are you thinking of? I still don't see what value BaseKNN is bringing.

For my production server example, where I need to return the centroids, this is exactly the model that I need - the others are not useful to me. For anyone that is doing a production setup where they need to get some of the metadata back (e.g., to look up the full datum in a database) this is currently the only model that will work. This is why it needs to be made public.

>  In particular, I find it weird that it inherits from base.Classifier, and then KNNRegressor inherits from it. That means that KNNRegressor is itself a classifier.

Oh, if that's the weirdness we could have KNNRegressor inherit from something else, or BaseKNN called something else. I do think, however, a KNN model that returns the neighbors explicitly without extra fluff is a base KNN model, so the name isn't so far off. 

> Also don't we want to make NearestNeighbors public?

I did but I thought you told me to not make it public?

> Finally, with regards to MinkowskiNeighbors, I think it's superfluous. You could simply pass a default distance function when you initialize the nn attribute. No need to have a class.

This was in response to a review request from @smastelini. So you are saving I should remove it? The reason this is nice is because we can actually include p as a parameter again. It's also the default used in both models.

MaxHalford:
> Oh, if that's the weirdness we could have KNNRegressor inherit from something else, or BaseKNN called something else. I do think, however, a KNN model that returns the neighbors explicitly without extra fluff is a base KNN model, so the name isn't so far off.

Right, but that's the purpose of the `NearestNeighbours` class, no?

I think there's some confusion here. What we're trying to do is what scikit-learn does. They have two classes for what is called "KNN": `KNNRegressor` and `KNNClassifier`. Then they have a `NearestNeighbors` class which just does nearest neighbors search. These are two different things. However, the former make use of the latter internally. That's why I'm saying you don't need `BaseKNN`, because it's already being covered by `NearestNeighbors`.

vsoch:
> Right, but that's the purpose of the NearestNeighbours class, no?

Not exactly - I still want to be running a model (e.g., using predict_one, learn_one) but for predict_one I want the full neighbors back. If we make it "special" then it's not going to plug into, for example, servers like django-river-ml (and the one you are working on) that expect that setup. Conceptually BaseKNN should be just another model type that happens to return neighbors but otherwise is comparable to the others.

I am not confused.

MaxHalford:
> I still want to be running a model (e.g., using predict_one, learn_one) but for predict_one I want the full neighbors back.

But that's not what you're doing. Both your `predict_one` and `predict_proba_one` methods are returning the neighbors and not the prediction. 

I see what you want to do. But I think this is too specific and I'm not in favor of breaking the API. It's so easy for you to override the behavior of the `KNNClassifier` as so:

```py
class MyKNNClassifier(KNNClassifier):

    def predict_one(self, x):
        neighbors = self.nn.find_nearest((x, None), n_neighbors=self.n_neighbors)
        return super().self.predict_one(x), neighbors
```

What do you think?

> I am not confused.

Well, I am.

vsoch:
Okay so you are saying there is no reasonable way to provide a KNN model that just returns the neighbors? I will always have to implement my own class?

vsoch:
Could you explain what you mean by breaking the API? Isn't the current refactor already breaking the API? How does exposing a base model change that?

vsoch:
In the future I think we should discuss how these reviews go - right now there are a lot of cooks in the kitchen, so to speak, and I'm finding I'm adding things just to have a request to remove them.

vsoch:
BaseKNN is removed: https://github.com/online-ml/river/pull/895/commits/7efd9cca3b5b387c522040132462fa1e3ab64a68

vsoch:
Interesting - so it appears that my base class was making the models pickle-able, and now they are not. Since this change was suggested to me, please advise for how to fix it.

vsoch:
okay no idea why it's broken now, sorry.

vsoch:
Picking issue is fixed.

vsoch:
heyo! It's been over a week - just wanted to check in! I was able to fix the pickle error so if y'all have time to take a look again, we can continue discussion.

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
 
 
-class KNNClassifier(BaseNeighbors, base.Classifier):
-    """k-Nearest Neighbors classifier.
-
-    This non-parametric classification method keeps track of the last
-    `window_size` training samples. The predicted class-label for a
-    given query sample is obtained in two steps:
+class KNNClassifier(BaseKNN, base.Classifier):
+    """
```
MaxHalford, line 10: The `K-Nearest Neighbors (KNN) for classification.` should be placed here. Take any other River class as an example for the docstring format

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
-    >>> metric = metrics.Accuracy()
-
-    >>> evaluate.progressive_val_score(dataset, model, metric)
-    Accuracy: 88.07%
+    >>> for x, y in dataset.take(100):
+    ...     model = model.learn_one(x, y)
 
+    >>> for x, y in dataset.take(1):
+    ...     model.predict_one(x)
+    {False: 0.0, True: 1.0}
```
MaxHalford, line 73: The convention is to add an extra empty line before the `"""`

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
+        self.classes = {x for x in self.window if x[0][1] is not None}
+
+    def learn_one(self, x: dict, y=None):
+        """Learn a set of features `x` and optional class `y`.
+        Parameters:
+            x: A dictionary of features.
+            y: A class (optional if known).
+        Returns:
+            self
+        """
```
MaxHalford, line 118: The docstring format is not correct. But anyway, you don't have to specify it, as the docstring is inherited via `base.Classifier`

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
+    def clean_up_classes(self) -> "KNNClassifier":
+        """
+        Classes that are added (and removed) from the window may no longer be valid.
+        This method cleans up the window and and ensures only known classes
+        are added, and we do not consider "None" a class. It is called every
+        `cleanup_every` step, or can be called manually.
+        """
+        self.classes = {x for x in self.window if x[0][1] is not None}
+
+    def learn_one(self, x: dict, y=None):
```
MaxHalford, line 111: You don't have to type hint `x`, it's inherited. As for y, it shouldn't default to `None` because this is a supervised model.

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
+        return self
 
+    def predict_one(self, x: dict):
+        """Predict the label of a set of features `x`.
+        Parameters:
+            x: A dictionary of features.
+        Returns:
+            The neighbors
         """
+        return self.predict_proba_one(x)
```
MaxHalford, line 147: This is wrong. `predict_one` returns a class, not a dict of probabilities. This method is anyway inherited by `base.Classifier`, so you don't have to implement it.

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
 
-        self.classes_.add(y)
-        x_arr = dict2numpy(x)
+    def predict_proba_one(self, x):
+        """Predict the class of a set of features `x`.
+        Parameters:
+            x: A dictionary of features.
+        Returns:
+            Lookup (dict) of classes and probability predictions (normalized)
+        """
```
MaxHalford, line 155: Same as `learn_one`, no need to specify the docstring.

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
+        nearest = self.nn.find_nearest((x, None), n_neighbors=self.n_neighbors)
 
-        self.data_window.append(x_arr, y)
+        # Default prediction for every class we know is 0.
+        # If class_cleanup is false this can include classes not in window
+        y_pred = {c: 0.0 for c in self.classes}
```
MaxHalford, line 160: I don't think this is correct. A probability distribution should sum to 1. I would instead default to `1 / n_classes`.

river/neighbors/knn_classifier.py
```diff
@@ -1,176 +1,193 @@
...
+            y_pred[nearest[0][0][1]] = 1.0
+            return y_pred
 
-        Returns
-        -------
-        proba
-            A dictionary which associates a probability which each label.
+        for neighbor in nearest:
+            distance = neighbor[-1]
+            y = neighbor[0][1]
```
MaxHalford, line 175: You can also do `(x, y), distance = neighbor`, which might be more readable, but it's your call

river/neighbors/knn_regressor.py
```diff
@@ -139,40 +125,34 @@ def predict_one(self, x):
...
-                dist for cnt, dist in enumerate(dists[0]) if cnt < self.data_window.size
-            ]
-        else:
-            neighbor_vals = [target_buffer[index] for index in neighbor_idx[0]]
-            dists = dists[0]
+        # Only include neighbors in the sum that are non None
+        neighbor_vals = [n[0][1] for n in nearest if n[0][1] is not None]
 
         if self.aggregation_method == self._MEAN:
             return np.mean(neighbor_vals)
```
MaxHalford, line 149: I would prefer avoiding numpy as much as possible. You'll find that it's actually slower than pure Python on small volumes of data. I would instead use `statistics.mean`

river/neighbors/knn_regressor.py
```diff
@@ -139,40 +125,34 @@ def predict_one(self, x):
...
-        elif self.aggregation_method == self._MEDIAN:
+
+        if self.aggregation_method == self._MEDIAN:
             return np.median(neighbor_vals)
```
MaxHalford, line 152: Same here, I would rather use `statistics.median`

river/test_estimators.py
```diff
@@ -75,6 +76,8 @@ def iter_estimators_which_can_be_tested():
         imblearn.RandomUnderSampler,
         imblearn.RandomSampler,
         model_selection.SuccessiveHalvingClassifier,
+        neighbors.NearestNeighbors,
+        neighbors.KNNRegressor,
```
MaxHalford, line 80: Any particular reason why `KNNRegressor` can't be tested?

MaxHalford (commented):
Here's a review! In general I would recommend looking at the docstring of `LogisticRegression` to give you guidelines on docstring formatting. I didn't comment every time there was a formatting issue, but hopefully you'll get the hang of the implicit style rules we have.

river/neighbors/knn_classifier.py
vsoch: This is verbatim from creme, so that is wrong too? https://github.com/MaxHalford/creme/blob/0ceffdee9e5afbb22de57f3666417015c71334be/creme/neighbors/knn.py#L158-L160

river/neighbors/knn_classifier.py
vsoch: I originally did the above because we couldn't guarantee the number of items to unwrap. I don't think that's the case anymore because it would be represented in x so happy to change it here, although it could lead to issues I can't anticipate now.

river/test_estimators.py
vsoch: I'll add it back - I forget when I originally removed it!

vsoch:
@MaxHalford all set.

MaxHalford:
Looks good to me! I'll merge. Sorry for the bumpy road, I guess next time will be easier.

vsoch:
Woohoo! Totally okay, it can be bumpy sometimes but we'll be slightly better the next time around :mountain:

Issue #892, opened 2022-03-24
a question on denstream#887


sherrypie077:
I got a question about def _merge: "creat a new o-micro-cluster" is under the condition "if not merged_status and len(self.o_micro_clusters)! = 0", so the point can not creat a new omc if there is no omc before?
Thanks a lot!

smastelini:
Pinging @hoanganhngo610!

hoanganhngo610:
@sherrypie077 Sorry for such a late response! Indeed, we forgot to take this situation into account, and given that the function `_get_closest_cluster_key()` only works when the `self.o_micro_clusters` are not empty. Would make an update on the code ASAP!
@smastelini @MaxHalford pinging for acknowledgement!

hoanganhngo610:
@sherrypie077 This problem has been solved in PR #898 as seen above. Please feel free to leave any comments or concerns remaining!

hoanganhngo610:
Close issue as it is fully solved!

sherrypie077:
Thanks for solving this!

PR #898, opened 2022-03-30
[DenStream] Handle the case when point p is not merged and o-micro-cluster is empty

hoanganhngo610:
This PR fixes the question raised in #892.

### DIFF
--- a/river/cluster/denstream.py
+++ b/river/cluster/denstream.py
@@ -121,13 +121,13 @@ class DenStream(base.Clusterer):
     1
 
     >>> denstream.predict_one({0:5, 1:4})
-    0
+    2
 
     >>> denstream.predict_one({0:1, 1:1})
-    1
+    0
 
     >>> denstream.n_clusters
-    2
+    3
 
     """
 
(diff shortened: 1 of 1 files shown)


river/cluster/denstream.py
```diff
@@ -236,6 +236,15 @@ def _merge(self, point):
...
                 )
                 self.o_micro_clusters[len(self.o_micro_clusters)] = mc_from_p
+            merged_status = True
+
+        # when p is not merged and o-micro-cluster set is empty
+        if not merged_status and len(self.o_micro_clusters) == 0:
+            mc_from_p = DenStreamMicroCluster(
+                x=point, timestamp=self.timestamp, decaying_factor=self.decaying_factor
+            )
+            self.o_micro_clusters = {0: mc_from_p}
```
smastelini, line 246: Hi @hoanganhngo610, what `self.o_micro_cluster` is before being set here? If it is `None`, calling `len` over it will result in a fail, right?

smastelini: An empty dictionary, just found it. Okay, it will work :D

hoanganhngo610: Yep this handles when the length is equal to 0, so basically, an empty dictionary. As I said, the function that queries the closest o-micro-cluster assumes that the set is not empty, so this case is basically left out!

PR #901, opened 2022-03-31
Implement iter_progressive_val_score

MaxHalford:


smastelini (commented):
I love that, and my PhD experiments are grateful for that!

Farewell, string parsing days!

PR #917, opened 2022-04-22
Update KSWIN and remove unnecessary numpy use in drift detectors

smastelini:
Changes:
- KSWIN
    - Remove the numpy usage in favor of `collections.deque`
    - Add seed parameter
    - Suppress convergence warning by using the default parameter of Scipy's two-sample Kolmogorov-Smirnov test
- Other detectors: remove numpy from examples


river/drift/kswin.py
```diff
@@ -107,31 +122,29 @@ def update(self, value) -> tuple:
...
-                self.window = self.window[-self.stat_size :]
+                self.window = collections.deque(most_recent, maxlen=self.window_size)
             else:
                 self._in_concept_change = False
         else:  # Not enough samples in sliding window for a valid test
             self._in_concept_change = False
 
-        self.window = np.concatenate([self.window, [value]])
-
         return self._in_concept_change, self._in_warning_zone
```
MaxHalford, line 150: From what I can tell `self._in_warning_zone` is never `True`, correct?

MaxHalford (commented):
Devil's advocate question: what's the rationale behind returning `self._in_concept_change` and `self._in_warning_zone`? Wouldn't it be simpler to just return whether or not a drift is detected?

smastelini:
To be really fair, I really don't know. I've never delve into the drift module much. I don't have any attachments to that. Maybe @jacobmontiel could shed some light in that matter.

smastelini:
The warning part is just to provide a centralized design for detectors that operate in two levels: warning and drift.

river/drift/kswin.py
smastelini: For ADWIN no. It only detects drifts

smastelini: We could check if these private variables are used in other detectors. If that is not the case, I vote for explicitly returning `drift, False`.

smastelini: [This](https://github.com/online-ml/river/blob/main/river/drift/ddm.py) detector seems to operate on two levels. There might be others.

I always considered drift detectors taken for granted, so I don't know much about this research field.

smastelini:
What I've learned during this PR:

- PageHinkley, ADWIN, and KSWIN support real-valued inputs and do not issue warnings (only drifts)
- DDM, EDDM, HDDM_A, and HDDM_W support only binary inputs (either 0 or 1) and produce both warning and drift signals

Although I feel the 0s and 1s for the second group of detectors are not ensured anywhere (either via `typing` or checks), the documentation states the expected input type. Maybe we can change the inputs of the second group to boolean values in the future.

PR #918, opened 2022-04-22
Streamline SRP{Classifier,Regressor}, remove unneeded numpy usage and add PeriodicTrigger

smastelini:
It started with the idea of removing the numpy-based random number generators. 

I ended up finding bugs (for instance, if the Poisson weight was zero for some base model, the next ones would not be updated) and making a buried utility accessible (PeriodicTrigger). I also made SRP somewhat robust to missing feature values.


smastelini:
Hey @MaxHalford, anything against merging this one? I will proceed to tackle the remaining ensembles.

PR #927, opened 2022-05-11
[ARF] Make the wrapper and base models share the same rng

smastelini:


smastelini:
There is still something odd with some pseudo-random portion of the codebase. In this PR, I unified the rng usage to a single instance, i.e., all the forest components share the same rng: trees, leaves, and the forest per see. Although the tests still fail, I like this design decision better.

All the tests pass successfully in my local OSX setup, so it's hard to reproduce the cause of the failure. I'll keep searching for the possible root cause. My current targets to check:

- the Poisson-distributed variable generator (possibly not the root cause, but making ARF rely on it was one of the few things I had changed before the failures started appearing).
- The synthetic data generator

MaxHalford:
I think it might be due to a new version of scikit-learn. The CI always uses the latest version. I'll investigate :)

MaxHalford:
Ok that's not it, but I'm still going to push small changes to adapt to the new scikit-learn version. I'll keep digging as to your issue in particular :)

MaxHalford:
It fails also, so the issue is not coming from the dataset. It's something specifc to `AdaptiveRandomForestClassifier`

MaxHalford:
@smastelini I made an interesting find: the adaptive random forest classifier tests fail even on a different branches, for instance see [here](https://github.com/online-ml/river/runs/6461193148?check_suite_focus=true).

smastelini:
Interestingly, the ARF tests are passing on Ubuntu (and on my local setup now). The failing tests are related to the modified anomaly detectors.

PR #944, opened 2022-06-02
Fix for protect_anomaly_detector logic

friskin:
The anomaly_dector should be updated if we are not protecting the detector or if it is not an anomaly.  I.e. the only time it should not be updated is when we are protecting the detector and it is an anomaly.


MaxHalford:
Good catch 🤦

Issue #948, opened 2022-06-04
Problem with ThresholdFilter and GaussianScorer


dberardo-com:
i have got this signal: 

![image](https://user-images.githubusercontent.com/65530457/172017334-35c2d50a-9200-4845-8f44-64c2dea33cc8.png)

and i am applying a gaussianScorer  + ThresholdFilter like in the timeseries Waterflow example.

The problem is that at the end of the execution, the gaussian filter has not learnt anything that happened after the jump at the beginning of April. Here is how the plot of the score_one(x,y) function looks like:

![image](https://user-images.githubusercontent.com/65530457/172017418-0021458d-2a63-4377-9c02-d705227d5a9d.png)

1 means anomaly ... so you can see that at the end of the execution, the gaussian scorer has not recognized the dirft ... although it had quite a few days to do so:

![image](https://user-images.githubusercontent.com/65530457/172017514-1e25e119-4969-48ce-a61b-31810ba954dc.png)

here you can see that the mean is stuck at 161, which is close to the initial value for the first couple of weeks

dberardo-com:
as an extra information, if i change the parameter window_size to 48, this is what i get:

![image](https://user-images.githubusercontent.com/65530457/172017883-5be94d34-8382-423f-b6a9-a40ea7006f18.png)

![image](https://user-images.githubusercontent.com/65530457/172017908-f61c4032-660c-4698-9c02-921445d9270e.png)


which means a slightly improved situation, but the adpatation is too slow here.


So my question here is: is the thresholder updating correctly or why is that change so slow?

dberardo-com:
and here is what happens for a smaller window_size of 30:

![image](https://user-images.githubusercontent.com/65530457/172018017-0b3babb3-7c9d-43a1-81a5-ed31369d9c8b.png)

![image](https://user-images.githubusercontent.com/65530457/172018039-314d091d-83cd-46fc-8195-6854bdb1a1b9.png)


which makes no sense, as i would expect an reduced number of anomalies detected

--> indeed if i set the window_size to 1 i get always 0 as anomaly score, which is correct

actually i always get 0 score for values of window_size which are below 30 ... is there a strange relation between grace_period and window_size?

dberardo-com:
alright, so to answer some of my own questions:

> actually i always get 0 score for values of window_size which are below 30 ... is there a strange relation between grace_period and window_size?

[...]

PR #949, opened 2022-06-05
Improvements on GaussianScorer

dberardo-com:
This PR adds 2 features and 3 parameters:

### Params

1. `window_size_t=None`
2. `timestamp_label=None` ("de-facto" mandatory if window_size_t != None)
3. `timestamp_format='%Y-%m-%dT%H:%M:%S%z `

### Features

1. GaussianScorer supports `grace_period > window_size`. This was added here: https://github.com/online-ml/river/commit/3163ea3bb67961b23fd7850ace437ef820258e5f
2. GaussianScorer now give options to pass `window_size` as a time_delta. To do this users have to set the `window_size_t` attribute. The .learn_one function will then try to guess the time value from a variable `t` following this priority list:

    a. `t` is a datetime object
    b. `t` is an integer (epoch timestamp)
   c. `t` is something else and will be parsed using the `datetime.strptime` function taking `timestamp_format` as formatting string
   
   The variable `t` is detived by the `_` dictionary using `timestamp_label` as an index

### DIFF
--- a/river/anomaly/gaussian.py
+++ b/river/anomaly/gaussian.py
@@ -1,8 +1,10 @@
 from river import anomaly, proba
-
+import datetime as dt
+import pytz
 
 class GaussianScorer(anomaly.base.SupervisedAnomalyDetector):
-    """Univariate Gaussian anomaly detector.
+
+     """Univariate Gaussian anomaly detector.
 
     This is a supervised anomaly detector. It fits a Gaussian distribution to the target values.
     The anomaly score is then computed as so:
@@ -46,11 +48,14 @@ class GaussianScorer(anomaly.base.SupervisedAnomalyDetector):
 
     """
 
-    def __init__(self, window_size=None, grace_period=100):
+    def __init__(self, window_size=None, window_size_t=None, timestamp_label=None, timestamp_format='%Y-%m-%dT%H:%M:%S%z', grace_period=100):
         self.window_size = window_size
+        self.window_size_t = window_size_t
+        self.timestamp_label=timestamp_label
+        self.timestamp_format=timestamp_format
         self.gaussian = (
-            proba.Rolling(proba.Gaussian(), window_size=self.window_size)
-            if window_size
+            proba.Rolling(proba.Gaussian(), window_size=self.window_size) if window_size 
+            else proba.TimeRolling(proba.Gaussian(), period=self.window_size_t) if window_size_t 
             else proba.Gaussian()
         )
         self.grace_period = grace_period
@@ -64,7 +69,17 @@ class GaussianScorer(anomaly.base.SupervisedAnomalyDetector):
         )
 
     def learn_one(self, _, y):
-        self.gaussian.update(y)
+        if isinstance(self.gaussian, proba.TimeRolling):
+            t = _[self.timestamp_label] if isinstance(_, dict) else _
+            parsed_t = (
+                t if isinstance(t,dt.datetime)
+                else dt.datetime.utcfromtimestamp(t)  if isinstance(t,int)
+                else dt.datetime.strptime(t,self.timestamp_format)
+            )
+            parsed_t =  parsed_t if parsed_t.tzinfo else pytz.utc.localize(parsed_t)
+            self.gaussian.update(y, parsed_t)
+        else:
+            self.gaussian.update(y)
         return self
 
     def score_one(self, _, y):
--- a/river/anomaly/gaussian.py
+++ b/river/anomaly/gaussian.py
@@ -4,7 +4,7 @@ import pytz
 
 class GaussianScorer(anomaly.base.SupervisedAnomalyDetector):
 
-     """Univariate Gaussian anomaly detector.
+    """Univariate Gaussian anomaly detector.
 
     This is a supervised anomaly detector. It fits a Gaussian distribution to the target values.
     The anomaly score is then computed as so:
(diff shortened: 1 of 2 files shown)


dberardo-com:
This PR addes to features requested here https://github.com/online-ml/river/issues/948#issuecomment-1146805351 and here https://github.com/online-ml/river/issues/948#issuecomment-1146808722

dberardo-com:
> Thanks for the pull request :)
> 
> > GaussianScorer supports grace_period > window_size. This was added here: [3163ea3](https://github.com/online-ml/river/commit/3163ea3bb67961b23fd7850ace437ef820258e5f)
> 
> Cool, I'll review the logic.
> 
> > GaussianScorer now give options to pass window_size as a time_delta. To do this users have to set the window_size_t attribute. The .learn_one function will then try to guess the time value from a variable t following this priority list:
> > a. t is a datetime object
> > b. t is an integer (epoch timestamp)
> > c. t is something else and will be parsed using the datetime.strptime function taking timestamp_format as formatting string
> 
> That's nice. However, I think there's way to do this in terms of user interface. Would you mind if I sent you a pull request with some edits?
> 
> As you can see there are some CI issues. One important thing is that you've forgotten to document your changes. It's not very helpful to document what you do in the pull request but not in the docstring. Ideally you also want to add an entry to the release notes. You have to put yourself in a user's shoe (other than yourself) when you develop a library :)

alright thanks for the tips, i will definitely try to document the changes then. Do i find guidelines on the CONTRIBUTE.md file?

i am kind of new to github workflows as i am used to gitlab, so if you aim for a fast release of the feature, i might need some guidance

river/anomaly/gaussian.py
```diff
@@ -64,10 +71,24 @@ def _dist(self):
...
+            )
+            parsed_t =  parsed_t if parsed_t.tzinfo else pytz.utc.localize(parsed_t)
+            self.gaussian.update(y, parsed_t)
+        else:
+            self.gaussian.update(y)
         return self
 
     def score_one(self, _, y):
-        if self._dist.n_samples < self.grace_period:
+        if self._dist.n_samples < self.grace_period and not self.grace_period_passed:
```
MaxHalford, line 92: Sorry but I don't understand. In what cases is `and not self.grace_period_passed` bringing something to the table? Isn't checking for `self._dist.n_samples < self.grace_period` sufficient?

MaxHalford (commented):
Hey there @dberardo-com. I took some time to review mostly because I was busy, but also because I had to think about it.

As you've probably seen, we've just added support for a generic rolling mechanism. Likewise for time-rolling. Our ambition is to push this further and enable it to be used with estimators, such as `GaussianScorer`. With that in mind, I would prefer to not add explicit support for a time-rolling mechanism in `GaussianScorer`. It feels to me like 1) we're too immature 2) the anomaly detection module is even more immature. I feel like adding support for a time-rolling mechanism in the anomaly detection module is simply too premature.

Sorry if this is disappointing. I do appreciate your energy however! I'm sure you can still use this in your own code :)

river/anomaly/gaussian.py
dberardo-com: well if you have a grace period of 200 and a time window of 30, then the condition `self._dist.n_samples < self.grace_perion` will never get true, and thus the model will never update

dberardo-com:
> Hey there @dberardo-com. I took some time to review mostly because I was busy, but also because I had to think about it.
> 
> As you've probably seen, we've just added support for a generic rolling mechanism. Likewise for time-rolling. Our ambition is to push this further and enable it to be used with estimators, such as `GaussianScorer`. With that in mind, I would prefer to not add explicit support for a time-rolling mechanism in `GaussianScorer`. It feels to me like 1) we're too immature 2) the anomaly detection module is even more immature. I feel like adding support for a time-rolling mechanism in the anomaly detection module is simply too premature.
> 
> Sorry if this is disappointing. I do appreciate your energy however! I'm sure you can still use this in your own code :)

oh i see, so you mean here that you would prefer to go for a more generic / reusable approach to rolling statistics, and TimeRolling might be to hacky to bring in?

Where do you see problems with the TimeRolling approach? Why do you say too immature?

MaxHalford:
Sorry, I wasn't clear.

We want to go with a generic approach. So we now have `Rolling` and `TimeRolling`. Those two are wrappers that can be wrapped around classes which implement `update` and `revert`. It doesn't yet work for estimators which have a `learn_one` and `{predict|score}_one` API.

dberardo-com:
> Sorry, I wasn't clear.
> 
> We want to go with a generic approach. So we now have `Rolling` and `TimeRolling`. Those two are wrappers that can be wrapped around classes which implement `update` and `revert`. It doesn't yet work for estimators which have a `learn_one` and `{predict|score}_one` API.

ok, this PR simply adds more attributes to the existing GuassianScore, i understand therefore that if you want to go for a wrapper, this would not be a natural step.

Feel free of scrapping this PR if it does not fit the roadmap, however i will keep using it in my code especially because i have irregular stream of data coming in, and thus i cannot rely on a "count" based approach for the GaussianScorer (see https://github.com/online-ml/river/issues/948#issuecomment-1155777301 )

i see that there are already conflicts, so i hope that the TimeRolling wrapper gets implemented ASAP so that i dont have to maintain my own version of it

MaxHalford:
> i see that there are already conflicts, so i hope that the TimeRolling wrapper gets implemented ASAP so that i dont have to maintain my own version of it

It's the dev version, see [here](https://riverml.xyz/dev/api/utils/TimeRolling/).

> ok, this PR simply adds more attributes to the existing GuassianScore, i understand therefore that if you want to go for a wrapper, this would not be a natural step.

If that's ok with you I'll close this PR

dberardo-com:
> > i see that there are already conflicts, so i hope that the TimeRolling wrapper gets implemented ASAP so that i dont have to maintain my own version of it
> 
> It's the dev version, see [here](https://riverml.xyz/dev/api/utils/TimeRolling/).

i see ... but i guess this is not the way to do it for SNARIMAX. I think in that case we need a Tumbling Window approach, and nor a Sliding window, like the one in the example.

Can the TimeRolling wrapper perform Tumbling stats already? Or should i try to implement it myself in it?

dberardo-com:
https://github.com/online-ml/river/discussions/971

PR #956, opened 2022-06-07
Fix errors, refactor and make the neighbors module future-proof

smastelini:
As part of the efforts in #954 to fix all mypy errors, the `neighbors` module was refactored.

Apart from mypy, we also want to make neighbors future-proof. Soon, we intend to add nearest neighbor methods that do not perform exact searches as our current solutions. Instead, they use index structures to perform approximate searches. So, we trade off search accuracy for speed.

This PR drops redundant classes, improves documentation and examples, and simplifies the class hierarchy in the neighbors' module.

In the future, the `NearestNeighbor`  class might be renamed to `ExactNearestNeighbors` and a more generic `NearestNeighbor` be created (with only abstract methods). In addition, an `ApproximateNearestNeighbor` might pop up (inheriting from `NearestNeighbor`).


river/neighbors/base.py
```diff
@@ -1,35 +1,48 @@
...
-    """A basic data structure to hold nearest neighbors.
+    The k-NN-based classifier and regressor store tuples with `(x, y)`, but only
+    `x` is used for distance calculations. This wrapper makes sure `x` is accessed
+    when calculating the distances.
 
     Parameters
     ----------
+    distance_function
+        The custom distance function to be wrapped.
+    """
```
MaxHalford, line 23: Style: you're supposed to add a carriage return before the `"""`

river/neighbors/base.py
```diff
@@ -1,35 +1,48 @@
...
+    def __init__(self, distance_function: DistanceFunc):
+        self.distance_function = distance_function
 
+    def __call__(self, a, b):
+        # Access x, which is stored in a tuple (x, y)
+        return self.distance_function(a[0], b[0])
+
+
+class NearestNeighbors:
+    """A basic data structure to hold nearest neighbors.
```
MaxHalford, line 34:
```suggestion
    """Exact nearest neighbors search data structure.
```

river/neighbors/base.py
```diff
@@ -1,35 +1,48 @@
...
         The minimum distance (similarity) to consider adding a point to the window.
-        E.g., a value of 0.0 will add even exact duplicates. Default is 0.05 to add
-        similar but not exactly the same points.
-
+        E.g., a value of 0.0 will add even exact duplicates.
     distance_func
-        An required distance function that accept two input items to compare
-        and optional parameters. It's recommended to use functools.partial.
+        A distance function that accept two input items to compare and optional
+        parameters.
```
MaxHalford, line 45:
```suggestion
        A distance function which accepts two input items to compare.
```

river/neighbors/knn_classifier.py
```diff
@@ -1,75 +1,82 @@
...
+    Accuracy: 84.55%
 
-    >>> for x, y in dataset.take(1):
-    ...     model.predict_one(x)
-    True
+    When defining a custom distance function you can rely on `functools.partial` to set default
+    parameter values. For instance, let's use the Manhattan function instead of the default Euclidean distance:
+
+    >>> import functools
+    >>> from river.utils.math import minkowski_distance
```
MaxHalford, line 70:
```suggestion
    >>> from river import utils
```

river/neighbors/knn_classifier.py
```diff
@@ -1,75 +1,82 @@
...
+    >>> model = (
+    ...     preprocessing.StandardScaler() |
+    ...     neighbors.KNNClassifier(
+    ...         window_size=50,
+    ...         distance_func=functools.partial(minkowski_distance, p=1)
```
MaxHalford, line 75:
```suggestion
    ...         distance_func=functools.partial(utils.math.minkowski_distance, p=1)
```

MaxHalford (approved):

river/neighbors/base.py
smastelini: Done!

river/neighbors/knn_classifier.py
smastelini: Also done!

PR #963, opened 2022-06-10
Drift refactor

smastelini:
As part of our efforts to standardize River's API, I refactored the drift detection module.

Summary of changes:

- `update` now returns `self` instead of a tuple
    - The drift state is accessed via a property
- Only methods that support warnings have a dedicated property for this end (`warning_detected`)
- The main entry point to check for a drift after updating the detector is the property `drift_detected`
- `value` -> `x`


river/base/drift_detector.py
```diff
@@ -10,47 +9,32 @@ class DriftDetector(base.Base):
 
     def __init__(self):
         super().__init__()
-        self._in_concept_change = False
-        self._in_warning_zone = False
+        self._drift_detected = False
 
     def reset(self):
```
MaxHalford, line 14: I would make this method private

river/base/drift_detector.py
```diff
@@ -10,47 +9,32 @@ class DriftDetector(base.Base):
 
     def __init__(self):
         super().__init__()
```
MaxHalford, line 11:
```suggestion
```

river/drift/ddm.py
```diff
@@ -37,7 +36,7 @@ class DDM(DriftDetector):
 
     * if $p_i + s_i \geq p_{min} + 3 * s_{min}$ -> Change detected
 
-    **Input:** `value` must be a binary signal, where 1 indicates error.
+    **Input:** `x` must be a binary signal, where 1 indicates error.
```
MaxHalford, line 39:
```suggestion
    The input `x` must be a binary signal, where `True` indicates an error.
```

river/drift/eddm.py
```diff
@@ -33,7 +32,7 @@ class EDDM(DriftDetector):
 
     $\alpha$ and $\beta$ are set to 0.95 and 0.9, respectively.
 
-    **Input:** `value` must be a binary signal, where 1 indicates error.
+    **Input:** `x` must be a binary signal, where 1 indicates error.
```
MaxHalford, line 35:
```suggestion
    The input `x` must be a binary signal, where `True` indicates an error.
```

river/drift/hddm_a.py
```diff
@@ -12,7 +11,7 @@ class HDDM_A(DriftDetector):
     values and returns the estimated status of the stream: STABLE, WARNING or
     DRIFT.
 
-    **Input:** `value` must be a binary signal, where 1 indicates error.
+    **Input:** `x` must be a binary signal, where 1 indicates error.
```
MaxHalford, line 14: Same here. But now that I read this again, I don't think I understand the explanation. As in, what's an error? 😅

smastelini: As far as I understand, these detectors were developed having a classifier in mind.

They always monitor a Bernoulli random variable, where 0 indicates that a classifier has not made a mistake.

The authors of the paper [Combining univariate approaches for ensemble change detection in multivariate data](https://www.sciencedirect.com/science/article/abs/pii/S1566253517301239) place EWMA, EDM, EDDM, HDDM* under the category of "Control Charts". These are the ones that have three states:

- control
- warning
- drift

Some, more recent, drift detectors such as ADWIN and KSWIN can work with continuous data. Page-Hinckley also accepts continuous data. The three of them only have a "normal" and "drift" state.

Ideally, we could differentiate these detectors via different base classes, but I don't know how to keep allowing them to be used in the ensembles and trees, due to the different types of inputs.

The easy solution is to keep things as they are and focus on documenting the update method.

Pinging @MaxHalford and @jacobmontiel!

river/base/drift_detector.py
smastelini: Done!

Commit [063ca61](https://github.com/online-ml/river/pull/963/commits/063ca61d06f5335c05cf78234dc9549e70fc7a72) also ensures that every non-random detector calls the `_reset` method internally after a drift is detected.

smastelini:
The test error happens because Hoeffding Adaptive Tree (HAT) actually expects that ADWIN is not automatically reset when drift is encountered. HAT accesses the window size and the error estimation (mean value stored in the window). So, resetting messes up with that.

I will think about a proper solution:

1. Add an optional parameter to the drift detectors (`reset_on_drift: bool = True)`
2. Refactor the tree algorithms to handle things internally

Opinions?

river/drift/hddm_a.py
MaxHalford: > The easy solution is to keep things as they are and focus on documenting the update method.

Sounds good to me!

river/tree/nodes/hatr_nodes.py
```diff
@@ -2,8 +2,7 @@
 import random
 import typing
 
-from river.drift.adwin import ADWIN
-from river.stats import Var
+from river.stats import Mean, Var
```
MaxHalford, line 5: As a general rule I would encourage `from river import stats`

smastelini: Me too. However, ironically and historically, we have `stats` as an init-param of the nodes. So, when I try `stats.Mean,` the issued dictionary parameter is accessed rather than the module. I might do something like: `from river import stats as sts`.

Currently, I am scanning every possible paper to try to understand this line:

https://github.com/online-ml/river/blob/46f8e874e328d7f37e8461eb700a428fb98ab1a7/river/tree/nodes/hatc_nodes.py#L277-L279

This portion is the core aspect of the algorithm that I can't find in the available formulae. The idea of resetting the detectors after each drift breaks the conditional on replacing the alternate trees. So I am trying to refactor the tree and make it more generic.

This mysterious bound is on my way :P

river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -60,7 +60,7 @@ class HoeffdingAdaptiveTreeClassifier(HoeffdingTreeClassifier):
         potential replacement to the current one.
     drift_detector
         The drift detector used to build the tree. If `None` then `drift.ADWIN` is used.
-    switch_delta
+    switch_signif
```
MaxHalford, line 63: @smastelini I would prefer full parameter name, so `switch_significance` in this case

smastelini: Done!

docs/releases/unreleased.md
```diff
@@ -36,10 +36,14 @@
 - Removed `proba.Rolling`, due to the addition of `utils.Rolling`.
 - Removed `proba.TimeRolling`, due to the addition of `utils.Rolling`.
 
+## rule
+- The default `splitter` was changed to `tree.splitter.TEBST` for memory and running time efficiency.
```
MaxHalford, line 40: Add spaces after titles :)

smastelini: Good catch! :)

smastelini:
Some updates and replies:

- I decided to leave "**Input:** `x` is an entry in a stream of bits, where 1 indicates error/failure and 0
    represents correct/normal values." (with a **bold** font) to call the user's attention.
     - Right now, we don't differentiate base classes for the drift detectors. So it's up to the user to pass the correct input type. This might change in the future, but I don't feel I have enough expertise in drift detection to group the detectors correctly and design different API entry points for them.

- I have refactored all drift detectors but HDDM_W, ADWIN, and KSWIN. EDM, EDDM, and Page-Hinkley were rewritten from scratch to leverage the `stats` module (and, quite frankly, try to fix some bugs).
    - I have refactored HDDM_A too. But since the paper description does not closely match the MOA code (provided by the original authors), I limited myself to using `stats.Mean` where I could and kept following the MOA logic.
    - HDDM_W is scarier and falls into the same problem as HDDM_A. I will try to do the same thing and use `stats.EWMean`, but let's see.

The getting started entry was also updated. Can you check it, @MaxHalford? You had left some comments that I addressed!

Almost there!

river/drift/ddm.py
smastelini: I left a comment about that in the main thread!

smastelini:
There we go! I was able to streamline HDDM_W too!

I'm finally finished!

PR #966, opened 2022-06-12
Postpone pandas import in base module

MaxHalford:
Before, on my laptop:

```sh
python -c "from river import stats"  0.91s user 0.22s system 268% cpu 0.419 total
```

Afterwards:

```sh
python -c "from river import stats"  0.36s user 0.09s system 263% cpu 0.174 total
```

I did specifically to make importing the `stats` module faster. There are many other places in the code base where we import pandas and scipy, so this doesn't have an impact in all modules. But it's a nice idea to have in mind.

### DIFF
--- a/river/base/classifier.py
+++ b/river/base/classifier.py
@@ -5,6 +5,9 @@ from river import base
 
 from . import estimator
 
+if typing.TYPE_CHECKING:
+    import pandas as pd
+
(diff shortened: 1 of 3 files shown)


gbolmier (approved):
Sweet! Getting the same results locally

PR #967, opened 2022-06-12
Generic rolling

MaxHalford:
Ok so here we go! I did a bit of meta-programming and managed to make a generic `Rolling` wrapper class. It supersedes the specific `Rolling` wrapper classes we had before. I did the same for `TimeRolling`.


smastelini:
I like the idea of having a unified set of tools to "roll" things! This is extremely powerful and should help standardize our API as well as future contributions :)

I left some comments related to typos and other small things. 

@MaxHalford, please let me know if you need help debugging stuff or fighting mypy. I'll be happy to help!

Issue #906, opened 2022-04-06
`RollingQuantile` performance with large windows


fox-ds:
## Versions
**`river` version**: 0.10.1 
**Python version**: 3.10.2

## Describe your task
Keeping track of a rolling quantile over a large window. Once the RollingQuantile reaches the windows size, it removes the oldest value and appends the new value.  For 'small' windows ( ~ < 100k ) this is no problem and the code runs really fast, but with larger window sizes (say, 500k) once the window is full, there is a significant performance drop when updating the `RollingQuantile`. I think the issue is primarily in the underlying `utils.SortedWindow`.

Furthermore, though the `RollingQuantile` is indeed a streaming algorithm, but there may be some downsides on the memory and computational complexity side when handling large windows (`O(n)`). Because the entire window is kept in memory and being used to calculate the quantile, it's really accurate. However, maybe we could trade in some of this accuracy for performance, e.g. by estimating the quantile just like the regular `Quantile` class, but then windowed (although I currently don't know of a specific algorithm that solves this issue, there are a few papers out there though).

## What kind of performance are you expecting?
Until the window is full: 347195 iterations/s
Once the window is full: **279** iterations/s

## Steps/code to reproduce

```python
from river.stats import RollingQuantile
from tqdm import tqdm

q = 0.997

rolling_q = RollingQuantile(window_size=500_000, q=q)

for x in tqdm(range(500_000)):
    rolling_q.update(x)

for x in tqdm(range(500_000)):
    rolling_q.update(x)

```

smastelini:
Hi @fox-ds, this is indeed problematic from a computational point of view. However, there is not much we can do currently.

IMHO, the main problem is that although multiple quantile estimators enable adding new points, removing old data is problematic. For instance, the P2 algorithm, implemented under `stats.Quantile`, does not allow "reverting". I am not sure if it is possible to modify the algorithm to enable that, nor if other more robust solutions (e.g., t-digest) support such operation.

In the end, `SortedWindow` does what the name implies. So every new element addition/removal implies re-sorting the elements in the window. I would love to see a "reversible" incremental estimator of quantiles, and I have to admit that I am totally noob in that matter. Maybe this solution is out there and it is simple.

smastelini:
Hey @AdilZouitine, would it be possible to keep a full-size binary tree in memory, even if the window is not full? Then instead of removing elements, we could mark them for removal. The tree would need to be rebalanced/rebuilt from time to time, which might become a problem. 

[...]

PR #968, opened 2022-06-13
Optimized append sorted window

AdilZouitine:
This pull request is an optimization for rolling methods.

When the window size reaches its maximum capacity to insert a new element, it is necessary to find the "oldest" element in the sorted list to delete it.

Usually, this search is done using a linear search with a complexity of o(n); however, in our case, the elements of the list are sorted, so it is possible to do a binary search o(log(n)). 
This optimization is a simple solution for large sliding windows https://github.com/online-ml/river/issues/906.

### DIFF
--- a/river/utils/sorted_window.py
+++ b/river/utils/sorted_window.py
@@ -47,7 +47,7 @@ class SortedWindow(collections.UserList):
     def append(self, x):
 
         if len(self) >= self.size:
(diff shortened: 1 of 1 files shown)


smastelini (approved):
LGTM!

I believe @fox-ds might be working on something too, but this solution ought to give us some nice speedups already.

PR #969, opened 2022-06-13
Implement average forecast metric

MaxHalford:
This is handy for being able to compare forecasting models. Indeed it's not as clear-cut to compare forecasting models if we only have `HorizonMetric` in our toolbox.

Hey @dberardo-com, could you let me know what you think?

### DIFF
--- a/docs/overrides/home.html
+++ b/docs/overrides/home.html
@@ -24,7 +24,7 @@
             linear-gradient(
                     to bottom,
                     var(--md-primary-fg-color),
-                    hsl(210, 100%, 60%) 99%,
+                    hsl(210.3, 100%, 81%) 99%,
                     var(--md-default-bg-color) 99%
             )
         }
@@ -210,6 +210,10 @@
                 margin-left: 1.0rem;
                 margin-right: 4.0rem;
                 align-items: center;
+            }
+
+            .tx-hero__image {
+                display: none;
             }
         }
 
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -18,6 +18,10 @@
 
 - Removed dependency to `optim`.
 
+## time_series
+
+- Added `time_series.HorizonAggMetric`.
+
 ## utils
 
 - Removed dependencies to `anomaly` and `compose`.
--- a/river/time_series/__init__.py
+++ b/river/time_series/__init__.py
@@ -2,7 +2,7 @@
 from . import base
 from .evaluate import evaluate, iter_evaluate
 from .holt_winters import HoltWinters
-from .metrics import ForecastingMetric, HorizonMetric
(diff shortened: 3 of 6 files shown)


smastelini:
Quoting @dberardo-com:

> how was this working before? which metric value was taken? the last in the "horizon" vector?

@MaxHalford, was the performance for each step up to `horizon` reported individually?

MaxHalford:
> how was this working before? which metric value was taken? the last in the "horizon" vector?

> @MaxHalford, was the performance for each step up to horizon reported individually?

Before, and still after this pull request, the metric at each step is calculated. I simply added an optional `agg_func` to `time_series.progressive_val_score` so that you aggregate the results instead. For this I had to implement a new metric.

> is it possible to "skip timesteps" when defining the horizon. In my example i would like to evaluate a forecasting model for 1 hour, 24 hours, 7 days, and 4 weeks ahead forecasts, so of course i could use the custom aggregation metric to filter out unwanted components from the vector, but the .forecast method will still need to be called 4724 times per iteration ... which is very time consuming. Is there a way around it?

Mmm that sounds a bit too specific. I'm sure there's a way to do this, but we would need to heavily refactor the `_iter_with_horizon` function in `time_series.evaluate.py`. And I don't see any way to proceed right now, sorry :/

> Also: i understand that anomalies are still included (not filtered out) in the evaluation right?

Indeed, they're not filtered out.

dberardo-com:
> > how was this working before? which metric value was taken? the last in the "horizon" vector?
> 
> > @MaxHalford, was the performance for each step up to horizon reported individually?
> 
> Before, and still after this pull request, the metric at each step is calculated. I simply added an optional `agg_func` to `time_series.progressive_val_score` so that you aggregate the results instead. For this I had to implement a new metric.
> 
> > is it possible to "skip timesteps" when defining the horizon. In my example i would like to evaluate a forecasting model for 1 hour, 24 hours, 7 days, and 4 weeks ahead forecasts, so of course i could use the custom aggregation metric to filter out unwanted components from the vector, but the .forecast method will still need to be called 4724 times per iteration ... which is very time consuming. Is there a way around it?
> 
> Mmm that sounds a bit too specific. I'm sure there's a way to do this, but we would need to heavily refactor the `_iter_with_horizon` function in `time_series.evaluate.py`. And I don't see any way to proceed right now, sorry :/
> 
> > Also: i understand that anomalies are still included (not filtered out) in the evaluation right?
> 
> Indeed, they're not filtered out.

thanks for the answers. For now i have implemented my own gridSearch loop around the .evalute function, so i do aggregation around it. I need to do this, because i want to filter out anomalies while grid searching.

I will give a try to the new aggregation function later on, when anomalies will be taken into consideration.

For now looking at the code, it seems to me that it should work.

PR #980, opened 2022-07-05
Fix SNARIMAX formula for seasonal features

MaxHalford:
A graduate student from the university of Hyogo (Japan) called Shogo Tanaka sent me an email enquiring about SNARIMAX. He correctly pointed out there was a bug in our formula for seasonal components. This PR fixes that. I've added unit tests to push the envelope.

Screenshot 2022-07-05 at 16 55 41

Shogo preferred that I make the change rather than contributing himself.

Shogo and I have an ongoing discussion about SNARIMAX. We'll see what fruits this discussion will yield :)


MaxHalford:
There's a unit test failing error but I think it's something transient with scikit-learn.

PR #981, opened 2022-07-07
Protect AdaptiveStandardScaler against numerical inconsistencies

smastelini:
Sometimes `stats.EWVar` can output negative values due to numeric imprecision. The underlying variance formulae are known to be unstable and sometimes called the "naive" approach in textbooks. For the time being, we don't have a version leveraging the Welford algorithm.

This PR protects `preprocessing.AdaptiveStandardScaler` against the numerical instability of EWVar.

### DIFF
--- a/river/preprocessing/scale.py
+++ b/river/preprocessing/scale.py
@@ -543,9 +543,13 @@ class AdaptiveStandardScaler(base.Transformer):
 
     def transform_one(self, x):
         return {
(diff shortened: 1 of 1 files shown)


MaxHalford (approved):

smastelini:
The failing test is not due to this PR. I think it is sklearn-related.

PR #984, opened 2022-07-11
Refactor covariance matrix

MaxHalford:
Ok this PR didn't end where I expected it to haha. My first intention was to contribute the inverse covariance matrix. But then I fell into the covariance rabbit hole. I'll do the inverse covariance matrix in another PR. Here are the contributions of this PR:

- Added a `revert` method to `Cov` and `Pearson`, meaning `RollingCov` and `RollingPearson` are not needed anymore
- Created a `covariance` module to mimic scikit-learn. My belief is that covariance matrix estimation is a topic of itself. I have a hunch finance people will like this (e.g. see [this](https://github.com/microprediction/precise))
- Improved the speed of the covariance matrix for `update_many`. The perf went down a little for `update`, I'm not really sure why, but it's negligible. Note that I found much faster methods at some point, but they didn't work if the features were shuffled or sampled between updates, which is a deal-breaker for us.
- Made `Cov` store the covariance instead of the sum of squares (but didn't manage to do the same for `Var`)
- Added new tests and tidied up existing ones

On the diabetes dataset (442 rows, 10 columns):

```diff
# Covariance matrix update
- 84.3 ms ± 251 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
+ 93 ms ± 5.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Covariance matrix update_many (for _ in range(100))
- 383 ms ± 2.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
+ 72.9 ms ± 1.59 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


river/stats/cov.py
```diff
@@ -48,23 +48,20 @@ def __init__(self, ddof=1):
...
-        self._C = 0
         self.cov = 0
 
     def update(self, x, y, w=1.0):
         dx = x - self.mean_x.get()
         self.mean_x.update(x, w)
         self.mean_y.update(y, w)
-        self._C += w * dx * (y - self.mean_y.get())
-        self.cov = self._C / max(1, self.mean_x.n - self.ddof)
+        self.cov += (w * dx * (y - self.mean_y.get()) - self.cov) / max(1, self.mean_x.n - self.ddof)
```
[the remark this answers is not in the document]
smastelini: Cool! I was looking into the code and thinking about this change. It totally makes sense :)

river/covariance/emp.py
```diff
@@ -64,93 +110,99 @@ class CovMatrix:
...
-                except KeyError:
-                    column.append("")
-            columns.append(column)
+        for i, j in itertools.combinations(sorted(X.columns), r=2):
+            try:
+                self[i, j]
+            except KeyError:
+                self._cov[i, j] = stats.Cov(self.ddof)
+            self[i, j]._iadd(
+                other_mean_x=mean[i],
```
smastelini, line 192: This kind of utility is more valuable than it appears. We might think about creating a standardized way of merging and building new stats objects from precomputed statistics (for instance, initializing an incremental stats estimator from a batch computation).

river/covariance/test_emp.py
```diff
@@ -0,0 +1,162 @@
...
+        for ddof in [0, 1]
+    ],
+)
+def test_covariance_update_many(ddof):
+
+    cov = covariance.EmpiricalCovariance(ddof=ddof)
+    p = 5
+    X_all = pd.DataFrame(columns=range(p))
+
+    for _ in range(5):
```
smastelini, line 102:
```suggestion
    for _ in range(p):
```

river/covariance/test_emp.py
```diff
@@ -0,0 +1,162 @@
...
+        for ddof in [0, 1]
+    ],
+)
+def test_covariance_update_many_shuffled(ddof):
+
+    cov = covariance.EmpiricalCovariance(ddof=ddof)
+    p = 5
+    X_all = pd.DataFrame(columns=range(p))
+
+    for _ in range(5):
```
smastelini, line 131:
```suggestion
    for _ in range(p):
```

river/covariance/test_emp.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_covariance_update_many_sampled():
+
+    # NOTE: this test only works with ddof=1 because pandas ignores it if there are missing values
+    ddof = 1
+    cov = covariance.EmpiricalCovariance(ddof=ddof)
+    p = 5
+    X_all = pd.DataFrame(columns=range(p))
+
+    for _ in range(5):
```
smastelini, line 152:
```suggestion
    for _ in range(p):
```

river/stats/cov.py
```diff
@@ -47,57 +66,71 @@ def __init__(self, ddof=1):
...
         self.cov += (
-            (old_mean_x - other.mean_x.get())
-            * (old_mean_y - other.mean_y.get())
-            * ((old_n * other.mean_x.n) / self.mean_x.n)
+            (old_mean_x - other_mean_x)
+            * (old_mean_y - other_mean_y)
+            * ((old_n * other_n) / self.mean_x.n)
         )
         # Reapply scale
         self.cov /= self.mean_x.n - self.ddof
```
smastelini, line 126: Is there a corner case where that `max(self.mean_x.n - self.ddof, 1)` would be needed here too?

river/covariance/emp.py
smastelini: `Var` used to have a `_from_state` method.

In this example, by following the same design pattern, we would do the following:

`other_cov = stats.Cov._from_state(...)`

and then

`self[i, j] += other_cov`

I like this design decision better than having both a `__iadd__` and `_iadd`. It is also generic enough to handle `__sub__` and `__isub__` (or any other stuff we might come up with in the future).

The `_from_state` name is up to debate, of course. We could make almost every stats object have this utility.

river/stats/test_parallel.py
```diff
@@ -9,39 +8,72 @@
...
+
+    for i, (x, y, w) in enumerate(zip(X, Y, W)):
+        A.update(x, w)
+        B.update(y, w)
+        C.update(x, w).update(y, w)
+
+        D = A + B
+        assert math.isclose(C.get(), D.get())
+
+        # TODO: clone should work instead of deepcopy
```
smastelini, line 61: `clone()` creates a fresh copy, right?

We have `clone` and `mutate`. Maybe we need to change the semantics a little bit and create a third utility:

- `clone`: creates a copy just like deepcopy
- `mutate`: mutates an existing object in place
- `??`: creates a fresh copy, just like `Class(**params)`

What do you think, @MaxHalford ?

smastelini (changes requested):
Great work @MaxHalford! I left some comments for further discussion.

About the frequency x reliability weights. Have you thought about a strategy to handle both cases? I believe the equations are slightly different for both cases.

river/stats/test_parallel.py
MaxHalford: Good idea! I added a `include_attributes` parameter to the `clone` method. I think that's preferable to adding a third method.

river/covariance/emp.py
MaxHalford: That's a very good idea! I'll look into it

river/stats/cov.py
MaxHalford: Better be safe. Good catch!

river/stats/test_parallel.py
smastelini: I also think this is a nice compromise :)

river/stats/cov.py
smastelini: XD

MaxHalford:
> About the frequency x reliability weights. Have you thought about a strategy to handle both cases? I believe the equations are slightly different for both cases.

Nope I didn't look into that.

river/stats/cov.py
```diff
@@ -68,12 +68,16 @@ def __init__(self, ddof=1):
         self.mean_y = stats.Mean()
         self.cov = 0
 
+    @property
```
smastelini, line 71: I got so used to `var.mean.n` that I didn't realize that `var.n` is much clearer! Good stuff

smastelini (approved):
LGTM!

smastelini:
I believe the failing tests are unrelated to this PR. Am I right, @MaxHalford?

MaxHalford:
Mmm yes weird; I'll look into it! :)

Thanks for the review mate!

PR #986, opened 2022-07-13
Remove unused attributes in stats.Cov

AdilZouitine:
Hi, 
While recoding the covariance in another language, I realized that the `n` attribute of `Cov` is never used.


MaxHalford:
🦅

PR #1002, opened 2022-08-24
Update setup.py

kulbachcedric:
Adding the [river-torch](https://github.com/online-ml/river-torch) repo to the river extras.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,7 @@ setuptools.setup(
             "watermark",
         ],
         "extra": ["river_extra"],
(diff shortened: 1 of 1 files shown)


MaxHalford:
Nice!

MaxHalford:
@kulbachcedric is `pip install "river[torch]"` working for you? If so I would add it as an option to your installation instructions.

kulbachcedric:
Works locally, but we should move the PyPi Package of `river_torch` for the next releases.

PR #1009, opened 2022-08-30
Bayesian linear regression

MaxHalford:
🥁🥁🥁

Ok I've finally done it! Looking back, I should have done this way earlier, seeing how simple it is.

I've compared the results to a full numpy implementation and it produces exactly the same (good) results. I'm very happy satisfied.

Next steps (but not for now):

- Implement a mini-batch version. I have the code ready, but I have to do some numeric checks to be safe. In theory it should produce exactly the same results.
- Reproduce the charts in Bishop's book in a notebook. I'll re-use my blog post for this.
- Implement the drift-resilient version of this, which shouldn't be difficult because I've already done it on my blog.


MaxHalford:
@LeonardAukea finally 😅

LeonardAukea:
> @LeonardAukea finally 😅

Awesome @MaxHalford I'll make sure to try it out shortly.

PR #1010, opened 2022-08-31
Textclust

Dennis1989:
Integration of a text-based stream clustering algorithm (textClust) into river, as discussed here: https://github.com/online-ml/river/discussions/1006


MaxHalford:
Thanks for starting a PR Dennis. There are quite a few style issues to fix, but nothing we can't solve together. Do you mind if I nit-pick and add a lot comments? 🤗

We're going to release a new version of River sometime this week or the next. I think it's better to take our time with this PR, and include it in the next version River (we make a new release roughly every 3 months, sometimes less). In that regard, we can take our time with this PR :)

Dennis1989:
No worries Max, go ahead. I am not really a software engineer but am certainly interested in fixing everything that is not adequate.

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.cluster import AgglomerativeClustering
+
+from river import base
+
+__all__ = ["textclust"]
+
+
+class textclust(base.Clusterer):
```
MaxHalford, line 13: It's more common in Python to use `PascalCase` for class names. So I would suggest calling this `TextClust`. Else some people are going to think this is a function.

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
+import math
+import typing
+
+import numpy as np
+import pandas as pd
+from sklearn.cluster import AgglomerativeClustering
```
MaxHalford, line 6: An issue here is that scikit-learn is not one of River's dependencies, and we would like to keep it like that. Agglomerative clustering is not a difficult algorithm to implement, so maybe we could implement a version ourselves. Also that might work better with the fact that we're using dictionaries.

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+
+__all__ = ["textclust"]
+
+
+class textclust(base.Clusterer):
+    r"""
+
+    textClust [^1][^2]
+
+    textClust is a stream clustering algorithm for textual data that can identify and track topics over time in a stream of texts.
```
MaxHalford, line 18:
```suggestion
    r"""textClust is a clustering algorithm designed for text.

    textClust[^1][^2] is a stream clustering algorithm for textual data that can identify and track topics over time in a stream of texts.
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    The algorithm uses a widely popular two-phase clustering approach where the stream is first summarised in real-time.
+    The result is many small preliminary clusters in the stream called `micro-clusters`. Micro-clusters maintain enough information to update and efficiently calculate the cosine similarity between them over time,
+    based on the TF-IDF vector of their texts. Upon request, the miro-clusters can be reclustered to generate the final result using any distance-based clustering algorithm, such as hierarchical clustering.
+    To keep the micro-clusters up-to-date, our algorithm applies a fading strategy where micro-clusters that are not updated regularly lose relevance and are eventually removed.
```
MaxHalford, line 22: The line lengths are a bit long here. We use a maximum of 100. Normally black fixes this automatically, but not for comments I believe (you have to do it yourself).

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+
+    Parameters
+    ----------
+    radius
+        Distance threshold to merge two micro-clusters. Must be within the range `(0,1]`
```
MaxHalford, line 27:
```suggestion
        Distance threshold to merge two micro-clusters. Must be within the range `(0, 1]`
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    _lambda
```
MaxHalford, line 28: It's preferable not to have parameters with a leading `_`. But of course `lambda` is a reserved keyword. In this case maybe you can use `fading_factor`.

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        Fading factor of micro-clusters
+    tgap
+       Time between outlier removal
+    termfading
```
MaxHalford, line 32: `term_fading` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        Determines whether individual terms should also be faded
+    realtimefading
+        Parameter that specifies whether natural time or the number of observations should be used for fading
+    micro-distance
```
MaxHalford, line 36:
```suggestion
    micro_distance
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    ----------
+    radius
+        Distance threshold to merge two micro-clusters. Must be within the range `(0,1]`
+    _lambda
+        Fading factor of micro-clusters
+    tgap
+       Time between outlier removal
+    termfading
+        Determines whether individual terms should also be faded
+    realtimefading
```
MaxHalford, line 34: `real_time_fading` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    auto_r
+        Parameter that specifies if  `radius` should be automatically updated
+    auto_merge
+        Determines, if close observations shall be merged together
+    sigma
+        Parameter that influences the automated trheshold adaption technique
+
+     Attributes
+    ----------
+    microclusters
```
MaxHalford, line 53: `micro_clusters` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    ...    {"text":'Is this the first document?',"idd":4,"cluster": 1},
+    ...    {"text":'This is super unrelated as well',"idd":5,"cluster": 2}
+    ... ]
+
+    >>> stopwords = [ 'stop', 'the', 'to', 'and', 'a', 'in', 'it', 'is', 'I', 'that', 'had', 'on', 'for', 'were', 'was']
+
+    >>> metric = metrics.AdjustedRand()
+
+    >>> model = compose.Pipeline(
+    ...     feature_extraction.BagOfWords(lowercase=True, ngram_range=(1,2), stop_words=stopwords),
```
MaxHalford, line 84:
```suggestion
    ...     feature_extraction.BagOfWords(lowercase=True, ngram_range=(1, 2), stop_words=stopwords),
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+    ...     cluster.textclust(realtimefading=False, _lambda=0.001, tgap=100, auto_r=True, radius=0.9)
+    ... )
+
+    >>> for x in corpus:
+    ...     y_pred = model.predict_one(x["text"])
+    ...     y = x["cluster"]
+    ...     metric = metric.update(y,y_pred)
+    ...     model = model.learn_one(x["text"], id = x["idd"])
```
MaxHalford, line 92:
```suggestion
    ...     model = model.learn_one(x["text"], id=x["idd"])
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        self.realtimefading = realtimefading
+        self.min_weight = min_weight
+        self.auto_r = auto_r
+        self.auto_merge = auto_merge
+        self.sigma = sigma
+
+        # Initialize important values
+        self.num_merged_obs = 0
+        self.t = None
+        self.lastCleanup = 0
```
MaxHalford, line 130: `last_cleanup` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        self.n = 1
+        self.omega = 2 ** (-1 * self._lambda * self.tgap)
+
+        self.assignment: typing.Dict[int, int] = {}
+        self.microclusters: typing.Dict[int, "textclust.microcluster"] = {}
+        self.clusterId = 0
+        self.microToMacro = None
```
MaxHalford, line 137: `micro_to_macro` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        self.upToDate = False
```
MaxHalford, line 138: `up_to_date` is preferable

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+        self.realtime = None
+        self.dist_mean = 0
```
MaxHalford, line 140: Shouldn't a lot of these attributes be private? I don't think you want to expose to users. If such is the case, I would add a leading `_` to mark them as private.

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+
+        # create a new distance instance for micro and macro distances.
+        self.micro_distance: textclust.distances = self.distances(micro_distance)
+        self.macro_distance: textclust.distances = self.distances(macro_distance)
+
+    def learn_one(self, x, time=None, sample_weight=None, **kwargs):
```
MaxHalford, line 146: I would call the `time` parameter `t`

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+
+            # if no close cluster is found we check if embeddings should be verified and maybe create a new one
+            else:
+                self.dist_mean += min_dist
+                clusterId = self.clusterId
+                self.assignment[self.n] = clusterId
+                self.microclusters[clusterId] = mc
+                self.clusterId += 1
+        else:
+            print("error - nothing to process")
```
MaxHalford, line 208: This happens if the input dictionary is empty right? I would ignore such a case, it's up to the users (or some other class) to have this responsibility of checking inputs. Also, you're not supposed to `print` errors, but to `raise` them

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+
+        # cleanup every tgap
+        if self.lastCleanup is None or self.t - self.lastCleanup >= self.tgap:
+            self._cleanup()
+
+        ## increment observation counter
+        self.n += 1
+        return clusterId
+
+    def predict_one(self, x, sample_weight=None, type="micro"):
```
MaxHalford, line 218: You have to document the `type` parameter here. Also wouldn't it be preferable to have this in the `__init__`?

river/cluster/textclust.py
```diff
@@ -0,0 +1,595 @@
...
+                smallest_key = key
+
+        ## if auto threshold is set, we determine the new threshold
+        if self.auto_r:
+            ## if we at least have two close micro clusters
+            if counter > 1:
+
+                ## our threshold
+                mu = (sumdist - min_dist) / (counter - 1)
+                treshold = mu - self.sigma * math.sqrt(squaresum / (counter - 1) - mu**2)
```
MaxHalford, line 265:
```suggestion
                threshold = mu - self.sigma * math.sqrt(squaresum / (counter - 1) - mu**2)
```

MaxHalford (commented):
Here's a first review pass :)

There's still a lot of stuff I would change, but this should get the ball rolling. Feel free to push back and ask for questions if I haven't been clear enough.

Overall, I'm ok merging this PR if we sort out the `sklearn.AgglomerativeClustering` issue.

Once this is merged, I don't mind taking some time to clean up the code myself. We have a bit of a coding style at River and we want to make sure all our code is somewhat homogeneous. It's a lot to ask of you to adapt to this style, so I really don't mind jumping in and doing the necessary changes myself (with your authorization of course).

hoanganhngo610:
Dear @Dennis1989! I see that this PR is inactive for quite some time. Would you mind if I take over and finish what we're currently having? It seems like most of them would be code-style related, and a fresh implementation of Agglomerative Clustering.

Dennis1989:
@hoanganhngo610 @MaxHalford Thank you both for your suggestions and support. I was on vacation the last two weeks and just returned yesterday. This week I plan to incorporate the suggested changes (which I think are all valid and necessary). If I encounter problems, I will return to both of you, if this is okay.

river/cluster/textclust.py
Dennis1989: The question here is, where do we place this implementation? Utils? Also, in what detail should this be implemented: 

1) Textclust provides a pre-computed distance matrix. It is possible to implement a version that only uses this mode. Otherwise, it would be possible to additionally allow different distance metrics (e.g., euclidean, cosine, etc.). 

2) Which linkage modes should we offer (single, complete average?)

Overall we should decide if we only treat agglomerative clustering as util for textclust or if we should implement it with the intention to be used by other algorithms as well.

Dennis1989:
I think I addressed most of the issues mentioned by @MaxHalford. Feel free to iterate over the code again, and don't hold back with criticism :)

river/cluster/textclust.py
```diff
@@ -0,0 +1,643 @@
...
+
+__all__ = ["TextClust"]
+
+
+class TextClust(base.Clusterer):
+    r"""
+
+    textClust [^1][^2]
+
+    textClust is a stream clustering algorithm for textual data that can identify and track topics
```
MaxHalford, line 16:
```suggestion
    r"""textClust, a clustering algorithm for text data. 

    textClust[^1][^2] is a stream clustering algorithm for textual data that can identify and track topics
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,643 @@
...
+    over time in a stream of texts. The algorithm uses a widely popular two-phase clustering
+    approach where the stream is first summarised in real-time.
+    The result is many small preliminary clusters in the stream called `micro-clusters`.
```
MaxHalford, line 19:
```suggestion
    approach where the stream is first summarised in real-time.
    
    The result is many small preliminary clusters in the stream called `micro-clusters`.
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,643 @@
...
+        self.micro_distance = self.distances(self.micro_distance)
+        self.macro_distance = self.distances(self.macro_distance)
+
+    def learn_one(self, x, t=None, sample_weight=None, **kwargs):
+
+        ## if id is specified
+        if kwargs["id"]:
+            id = kwargs["id"]
+        else:
+            id = None
```
MaxHalford, line 163: You have to document this behavior somewhere for users to understand.

I think we should look into consistently having an `id` parameter in the `learn_one` method of every clustering method (cc @hoanganhngo610)

river/cluster/textclust.py
```diff
@@ -0,0 +1,643 @@
...
+            localdict[new_key]["ids"] = [None]
+
+        ngrams = localdict
+        ngrams = dict(ngrams)
+
+        return self.get_assignment(ngrams, type=type)
+
+    # finds the closest micro cluster
+    def _get_closest_mc(self, mc, idf, distance):
+
```
MaxHalford, line 246:
```suggestion
    def _get_closest_mc(self, mc, idf, distance):
        """Return the closest micro cluster."""
```

river/cluster/textclust.py
```diff
@@ -0,0 +1,643 @@
...
+        # initialize all distances to 0
+        distances = pd.DataFrame(np.zeros((numClusters, numClusters)), columns=ids, index=ids)
+
+        for idx, row in enumerate(ids):
+            for col in ids[idx + 1 :]:
+                # use the macro-distance metric to calculate the distances to different micro-clusters
+                dist = self.macro_distance.dist(clusters[row], clusters[col], idf)
+                distances.loc[row, col] = dist
+                distances.loc[col, row] = dist
```
MaxHalford, line 388: It's not efficient to call `.loc` many times like this. You should instead use simply Python lists, or a numpy array. And then convert the data to a pandas DataFrame. But I suppose we don't have to look into performance in this PR :)

MaxHalford (approved):
Here's some more comments.

There's a few thing I would still change, such as using dataclasses. But I think we can merge this as it is, so you can start working on the example notebook, and then I can do some refactoring work on the side if that's ok with you. I'll make a pull request to show you what I would do, and we can agree/disagree :)

river/cluster/textclust.py
Dennis1989: I agree. Not super efficient, but we should keep in mind that this method is only called when macro-clusters are generated from micro-clusters (which usually does not happen during the online phase). Therefore it is (a) really infrequently executed and (b) only used on the significantly smaller number of micro clusters.

MaxHalford: Understood. We could run some profiling to understand what could be improved anyway. But it's not important right now.

river/cluster/textclust.py
Dennis1989: So do I need to fix your comments first or how do we proceed?

MaxHalford: Yes you can fix these first I would say :)

river/cluster/textclust.py
hoanganhngo610: @MaxHalford  I also agree, since CluStream also depends a lot on distinguishing different micro clusters with unique IDs. My main concern is that how do we choose the format for the ID when two micro-clusters are formed?

MaxHalford: I don't know, I'll have to look into it. Let's not mix sample IDs with cluster IDs, I'm not sure which ones you're talking about

Dennis1989: Yes, @MaxHalford is right. These are two different things. 

So why do we need IDs for observations? Especially for text-based input data, a lot of metadata is associated (thinking of Tweets that hold user objects, locations, and other important information). This information can be used within post hoc analyses to find additional characteristics of clusters that are only built based on the text (e.g., which users belong to a certain cluster). Therefore it is necessary to store the IDs. The question here is if IDs should also be subject to concept drift to understand the current importance of observations. What I think is super important is that we do not "fade away" cluster IDs. It should always be possible to know which ID belongs to which cluster. However, we also have to consider what happens if a cluster itself is removed from the current clustering (maybe cluster - ID assignment should be persisted in secondary memory?)

So why do we need IDs for clusters? To understand how clusters were formed over time (e.g., merged together, etc.)

MaxHalford: I understand that we need to identify samples with identifiers. But I'm still a bit doubtful as to why this would have to be handled by the clustering model. Can't we just store the cluster assignments outside the clustering model? Indeed, it sounds like a feature you would want to do for any clustering task.

> So why do we need IDs for clusters? To understand how clusters were formed over time (e.g., merged together, etc.)

That sounds like it should be handled internally.

Dennis1989: > Can't we just store the cluster assignments outside the clustering model?

Yes, we could do this. It would be up to the user to ensure storage. If we decide on this, we should provide some copy&paste examples as I think this is something that people want to do frequently.

MaxHalford: Yes it's a good excuse to document it properly. Something in the user guide would be nice. You can kill two birds with one stone when you work on the notebook in a future PR :).

I'll let you remove this part from the code, then we can merge and move on.

PR #1013, opened 2022-09-02
Check mutable attributes exist

MaxHalford:
Thanks @smastelini for adding mutable attributes in #1012 :)

I'm just adding a unit test for making sure each mutable attribute actually exists.


smastelini (commented):
Perfect!

MaxHalford:
Aaaaand a bug was already caught 😄

smastelini:
My bad 😅

PR #1014, opened 2022-09-02
Add update many for metrics

kulbachcedric:
Hi,

based on #995, we could start by adding the 'update_many' function within the metrics base class that simply calls the update function.
@MaxHalford, @smastelini  What do you think?

Best
Cedric

### DIFF
--- a/river/metrics/base.py
+++ b/river/metrics/base.py
@@ -54,6 +54,7 @@ class Metric(base.Base, abc.ABC):
         assert len(y_true) == len(y_pred)
         [self.update(y_t,y_p) for (y_t,y_p) in zip(y_true,y_pred)]
         return self
+
     def is_better_than(self, other) -> bool:
         op = operator.gt if self.bigger_is_better else operator.lt
         return op(self.get(), other.get())
--- a/river/metrics/base.py
+++ b/river/metrics/base.py
@@ -3,6 +3,7 @@ import collections
 import numbers
 import operator
 import typing
+
 import pandas as pd
 
 from river import base, stats, utils
@@ -52,7 +53,7 @@ class Metric(base.Base, abc.ABC):
     def update_many(self, y_true: pd.Series, y_pred: pd.Series) -> "Metric":
         """Update the metric with multiple instances"""
         assert len(y_true) == len(y_pred)
-        [self.update(y_t,y_p) for (y_t,y_p) in zip(y_true,y_pred)]
+        [self.update(y_t, y_p) for (y_t, y_p) in zip(y_true, y_pred)]
         return self
 
     def is_better_than(self, other) -> bool:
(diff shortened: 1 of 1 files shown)


MaxHalford:
I'm sorry but I'm not really keen on this. If you do this, it means every metric now has a `update_many` method. That means users will use it. That then means we can go back and break the API. It's also not adding a lot of value, as it's just doing a for loop.

For river-torch, I get it that you're working a lot with mini-batches. I think it's fine to have a utility method in river-torch for this.

I'm sorry for not being enthusiastic about this, I just am not really convinced it adds significant value. Adding methods is a big decision, and I rather not do without being 100% convinced. It already took me a lot of convincing to accept having mini-batch methods in River 😅

By the way, `Mean` and `Var` have `update_many` methods. But they're not used anywhere. I would actually possibly  think of removing those methods.

kulbachcedric:
Hi Max,
> I'm sorry but I'm not really keen on this. If you do this, it means every metric now has a `update_many` method. That means users will use it. That then means we can go back and break the API. It's also not adding a lot of value, as it's just doing a for loop.
> 
No, worries! i thought it would be useful, as some estimators support the method `learn_many` method. The idea was precisely that all metrics have an `update_many` method. If there is a more efficient method (should be in some cases easy in comparison of iterating a loop) for a specific metric, it can simply be overwritten.
> For river-torch, I get it that you're working a lot with mini-batches. I think it's fine to have a utility method in river-torch for this.
> 
> I'm sorry for not being enthusiastic about this, I just am not really convinced it adds significant value. Adding methods is a big decision, and I rather not do without being 100% convinced. It already took me a lot of convincing to accept having mini-batch methods in River 😅
> 
I think we should either fully support mini-batches or just allow incremental learning. Here the `update_many` method would support mini-batches for the metrics. At least it was my idea behind this PR.

> By the way, `Mean` and `Var` have `update_many` methods. But they're not used anywhere. I would actually possibly think of removing those methods.
I would then remove them as it would be consistent. 
A minor thing that I have noticed in river 0.12.1 is that using the `stats` classes within the `Rolling` class  (e.g. `utils.Rolling(Mean(), window_size=250)`) leads to type hint errors.

smastelini:
Hi @MaxHalford and @kulbachcedric. 

I'm with @MaxHalford on that. I believe `*_many` should apply when we can take advantage of vectorization. 

Maybe this is the case with metrics. If you can, for instance, update the confusion matrix using batch operation then I'm in. Or error metrics. 

Still, this leads to creating a specialized method for each metric. 

What do you think?

MaxHalford:
I'll give this a think. I do prefer the idea of only adding `update_many` to metrics that support vectorization. We could also have `MiniBatchMetric` wrapper class that adds these methods to metrics that don't support vectorization. But I rather not add it by default to each metric.

kulbachcedric:
Valid point 👍
Closing this one.

PR #1023, opened 2022-09-09
Refactor code and use Welford's algorithm to CluStream

hoanganhngo610:
This new implementation of CluStream serves two purposes:
- One, this version is a refactor of the previous one with certain improvements, such as shortening variable names, removing redundant portions/loops and ensuring that private parts can not be accessed by users.
- Two, this implementation is an improvement of the original version by using Welford's algorithm to calculate the incremental variance of the data points and time stamps instead of having to save the number of observations, linear sum and sum of squares as per the traditional convention. This is facilitated through `stats.Var` within `River`.

### DIFF
--- a/river/cluster/clustream.py
+++ b/river/cluster/clustream.py
@@ -241,7 +241,9 @@ class CluStream(base.Clusterer):
         if re_cluster is True or self._kmeans_mc is None:
             self._mc_centers = {i: mc.center for i, mc in self.micro_clusters.items()}
 
-            self._kmeans_mc = cluster.KMeans(n_clusters=self.n_macro_clusters, seed=self.seed, **self.kwargs)
+            self._kmeans_mc = cluster.KMeans(
+                n_clusters=self.n_macro_clusters, seed=self.seed, **self.kwargs
+            )
             for center in self._mc_centers.values():
                 self._kmeans_mc = self._kmeans_mc.learn_one(center)
 
--- a/river/cluster/clustream.py
+++ b/river/cluster/clustream.py
@@ -32,6 +32,11 @@ class CluStream(base.Clusterer):
     the use of Welford's algorithm [^2] to calculate the incremental variance, facilitated
     through `stats.Var()` available within `River`.
 
+    Since `River` does not support an actual "off-line" phase of the clustering algorithm (as data points
+    are assumed to arrive continuously, one at a time), we introduce a new parameter `time_gap`. After each
+    `time_gap`, an incremental K-Means clustering algorithm will be initialized and applied on currently available
+    micro-clusters to form the final solution, i.e. macro-clusters.
+
     Parameters
     ----------
     n_macro_clusters
@@ -45,6 +50,9 @@ class CluStream(base.Clusterer):
     time_window
         If the current time is `T` and the time window is `h`, we only consider
         the data that arrived within the period `(T-h,T)`.
+    time_gap
+        After each `time_gap`, the incremental K-Means will be applied on the current set of micro-clusters
+        to form the final macro-clusters (final solution).
     seed
        Random seed used for generating initial centroid positions.
     kwargs
@@ -88,6 +96,7 @@ class CluStream(base.Clusterer):
     >>> clustream = cluster.CluStream(
     ...     n_macro_clusters=3,
     ...     max_micro_clusters=5,
+    ...     time_gap=3,
     ...     seed=0,
     ...     halflife=0.4
     ... )
@@ -112,6 +121,7 @@ class CluStream(base.Clusterer):
         max_micro_clusters: int = 100,
         micro_cluster_r_factor: int = 2,
         time_window: int = 1000,
+        time_gap: int = 100,
         seed: int = None,
         **kwargs,
     ):
@@ -120,6 +130,7 @@ class CluStream(base.Clusterer):
         self.max_micro_clusters = max_micro_clusters
         self.micro_cluster_r_factor = micro_cluster_r_factor
         self.time_window = time_window
+        self.time_gap = time_gap
         self.seed = seed
 
         self.kwargs = kwargs
@@ -130,8 +141,8 @@ class CluStream(base.Clusterer):
         self._timestamp = -1
         self._initialized = False
 
-        self._kmeans_mc = None
         self._mc_centers: typing.Dict[int, typing.DefaultDict] = {}
+        self._kmeans_mc = None
 
     def _maintain_micro_clusters(self, x, w):
         # Calculate the threshold to delete old micro-clusters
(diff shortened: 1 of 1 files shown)


hoanganhngo610:
@MaxHalford Would you mind having a really quick pass over this PR? I and @smastelini have been working on this together, so we both feel that this is already in good shape!

MaxHalford:
Hey! Sure I will, but right now I'm going to a music festival for the weekend 🙃

hoanganhngo610:
@smastelini FYI, I have managed to move the calculation of macro-clusters to learn_one. Basically speaking, the model will now calculate micro cluster centers and apply K-Means on them after each time_gap, just as what we have discussed previously!

smastelini:
Hi @hoanganhngo610, thanks for the PR. Since I am not an expert in clustering, I decided to read at least the core portion of the CluStream paper, and some things struck me:

1. CluStream relies heavily on saving snapshots of the model, possibly to the disk. We don't do that.
2. Micro clusters (MC) have unique ids. Merged MC's ids are a list of their constituent MC's ids. We also do not implement this strategy.
3. When clustering, the user can supply two timestamps, and CluStream will do its best to retrieve a clustering corresponding to the state of the clusters during the supplied time interval. In River, we only implement the strategy to delete old data, not serializing snapshots of the models. I believe that is fine for now, but we can revisit this decision in the future if necessary. One thing, though. We could use the same id scheme of the original algorithm and supply`__sub__` and `__add__` methods to manipulate clustering models. But let's save it for a rainy day. We can discuss that another time.

I think I have an idea of how to compromise between creating a new k-Means model every time and using a time gap. What about defining `def predict_one(self, x, re_cluster=True)`?

When `re_cluster` is True (default), a new incremental k-Means will be defined for each call of `predict_one`. If set to False, we leverage an existing cluster model. This way is up to the user to balance the cost x accuracy trade-off.

hoanganhngo610:
@smastelini Thank you so much for your review! After discarding the problems of unique IDs and the heavy dependance on snapshots of model, I totally agree of using `def predict_one(self, x, re_cluster=True)` so that it allows user to have more flexibility towards when to initiate the re-clustering process of micro-clusters to macro-clusters.
I will update the changes accordingly!

river/cluster/clustream.py
```diff
@@ -45,6 +50,9 @@ class CluStream(base.Clusterer):
     time_window
         If the current time is `T` and the time window is `h`, we only consider
         the data that arrived within the period `(T-h,T)`.
+    time_gap
+        After each `time_gap`, the incremental K-Means will be applied on the current set of micro-clusters
```
MaxHalford:
```suggestion
        An incremental k-means is applied on the current set of micro-clusters after each `time_gap`.
```

river/cluster/clustream.py
```diff
@@ -32,6 +32,11 @@ class CluStream(base.Clusterer):
     the use of Welford's algorithm [^2] to calculate the incremental variance, facilitated
     through `stats.Var()` available within `River`.
 
+    Since `River` does not support an actual "off-line" phase of the clustering algorithm (as data points
```
MaxHalford:
```suggestion
    Since River does not support an actual "off-line" phase of the clustering algorithm (as data points
```

river/cluster/clustream.py
```diff
@@ -32,6 +32,11 @@ class CluStream(base.Clusterer):
...
+    are assumed to arrive continuously, one at a time), we introduce a new parameter `time_gap`. After each
```
MaxHalford:
```suggestion
    are assumed to arrive continuously, one at a time), a `time_gap` parameter is introduced. After each
```

river/cluster/clustream.py
```diff
@@ -229,20 +243,27 @@ def learn_one(self, x, w=1.0):
         # Otherwise, closest micro-clusters are merged with each other.
         self._maintain_micro_clusters(x=x, w=w)
 
-        return self
+        # Apply incremental K-Means on micro-clusters after each time_gap
+        if self._timestamp % self.time_gap == self.time_gap - 1:
```
MaxHalford: I'm not sure I like the "timestamp" because it's not really a timestamp. I would use something like `_n_obs_seen`

river/cluster/clustream.py
hoanganhngo610: We have decided to get rid of the use of `time_gap` from the discussion above. As such, this would be marked as resolved!

river/cluster/clustream.py
hoanganhngo610: Marked as resolved since `time_gap` is now omitted.

river/cluster/clustream.py
hoanganhngo610: Mark as resolved since `time_gap` is now omitted.

river/cluster/clustream.py
hoanganhngo610: We inherited this name directly from the paper so that it makes more sense to the users. Since this variable will also be seen and taken into account by micro clusters, we prefer it to be kept that way in order to make more sense word-wise.

MaxHalford: Ok no problem then if it's the terminology used in the paper

MaxHalford:
> I think I have an idea of how to compromise between creating a new k-Means model every time and using a time gap. What about defining def predict_one(self, x, re_cluster=True)?

I rather not do this and would go with what @hoanganhngo610 is suggesting. The issue I have with putting parameters in `predict_one` is that it's a slippery slope. Ideally we want to put the less stuff there possible. I find it makes more sense to have a `reclustering_strategy` parameter in the `__init__` or something like that.

Also, taking a step back, I would argue that this clustering algorithm is not truly online. But let's say it's fine to keep it :)

river/cluster/clustream.py
hoanganhngo610: Thank you so much @MaxHalford!

smastelini:
What troubles me is that our design deviates a lot from the original CluStream. But it is close enough, I guess. 

I see your point @MaxHalford. Having different re-clustering strategies might do the trick. 

What about (for now) either re-clustering every time predict_one is called or after time_window examples are observed?

hoanganhngo610:
@MaxHalford @smastelini I have just returned to the use of `time_gap`, and finalized the changes that Max suggested above. Although I agree that this is not totally the same as the original CluStream, it is close enough (in my personal opinion) and also among the closest versions that we currently have available within the world of open source online ML. However, the snapshot and use of unique IDs for different micro-cluster would be something that we may have to consider in the future.
@MaxHalford I also agree (and maybe we all) that CluStream is not a fully online algorithm, since it utilizes the concept of online and offline phase. However, we decided to keep it since it is the foundation of a huge class of online clustering algorithms (you can have a look at the paper of Matthias Carnein [here](https://doi.org/10.1007/s12599-019-00576-5)), to diversify and to facilitate benchmarking for future online algorithms implemented into River.

river/cluster/clustream.py
MaxHalford: @hoanganhngo610 not for this change right?

hoanganhngo610: Not for this change. This has been fixed in the latest commit below!

smastelini:
LGTM!

MaxHalford:
Yes I agree that this online/offline paradigm is necessary. We can see it in action in textClust too.

Let's merge!

PR #1029, opened 2022-09-14
Lossy counting

smastelini:
Add the Lossy Counting with a fast forgetting factor for frequent itemsets.

Thanks, @BrunoMVeloso for the base code :)

### DIFF
--- a/.github/workflows/branch-tests.yml
+++ b/.github/workflows/branch-tests.yml
@@ -41,7 +41,7 @@ jobs:
           key: ${{ runner.os }}
 
       - name: Download datasets
-        run: python -c "from river import datasets; datasets.CreditCard().download(); datasets..TREC07().download()"
+        run: python -c "from river import datasets; datasets.CreditCard().download(); datasets.TREC07().download()"
 
       - name: pytest
         run: pytest


smastelini:
I am pinging @BrunoMVeloso as the algorithm author to check that.

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
+import math
+import typing
+from operator import itemgetter
```
MaxHalford, line 3: I would `import operator` and do `operator.itemgetter`

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+
+                if freq + delta <= current_bucket:
+                    prune.append(key)
+
+            for key in prune:
+                del self._entries[key]
+
+            self._delta = self._bucket_width + self._delta * self.forgetting
+
+    def get(self):
```
MaxHalford, line 113: Here I would suggest doing like what `collections.Counter` does. Add a `k` parameter that defaults to 1. Also you might just want to return the keys and add a `__getitem__` method to get the count. But these are just wild suggestions, don't feel obliged to do it this way.

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+    - Estimated frequencies are less than the true frequencies by at most `epsilon * N`.
+
+    Parameters
+    ----------
+    support
+        The support threshold used to determine if an item is frequent. The value of `support` must
+        be in $[0,1]$.
+    epsilon
+        Error parameter to control the accuracy-memory tradeoff. The value of `epsilon` must be
+        in $(0,1]$ and typically `epsilon` $\\ll$ `support`. The smaller the `epsilon`, the more
```
MaxHalford, line 32:
```suggestion
        in $(0, 1]$ and typically `epsilon` $\\ll$ `support`. The smaller the `epsilon`, the more
```

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+        accurate the estimates will be, but the count sketch will have an increased memory
+        footprint.
+    forgetting
+        Forgetting factor applied to the frequency estimates to reduce the impact of old items.
+        The value of `forgetting` must be in $(0,1]$.
```
MaxHalford, line 37:
```suggestion
        The value of `forgetting` must be in $(0, 1]$.
```

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+    Considering a data stream where `N` elements were observed so far, the Lossy Counting
+    algorithm has the following properties:
+
+    - All item(set)s whose true frequency exceeds `support * N` are output. There are no
+    false negatives;
+
+    - No item(set) whose true frequency is less than `(support - epsilon) * N` is outputted;
+
+
+    - Estimated frequencies are less than the true frequencies by at most `epsilon * N`.
```
MaxHalford, line 23:
```suggestion
    - All item(set)s whose true frequency exceeds `support * N` are output. There are no
    false negatives
    - No item(set) whose true frequency is less than `(support - epsilon) * N` is outputted
    - Estimated frequencies are less than the true frequencies by at most `epsilon * N`
```

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+from river import stats
+
+
+class LossyCount(stats.base.Univariate):
+    """Lossy Counting with Forgetting factor[^1].
+
+    Keep track of the frequent itemsets in a data stream and apply a forgetting factor to
+    discard previous frequent items that do not often appear anymore.
+
+    Considering a data stream where `N` elements were observed so far, the Lossy Counting
```
MaxHalford, line 14: I would prefer `n` to `N`, but it's your call

river/stats/lossy.py
```diff
@@ -0,0 +1,119 @@
...
+from operator import itemgetter
+
+from river import stats
+
+
+class LossyCount(stats.base.Univariate):
+    """Lossy Counting with Forgetting factor[^1].
+
+    Keep track of the frequent itemsets in a data stream and apply a forgetting factor to
+    discard previous frequent items that do not often appear anymore.
```
MaxHalford, line 12: If I were I would expand this part a little bit to explain what are frequent itemsets, and how this differents to using `collections.Counter`.

MaxHalford (commented):
Very cool! Added some suggestions :)

river/stats/lossy.py
smastelini: Done!

river/stats/lossy.py
smastelini: The paper used `N`, but in this context `n` will do the trick.

river/stats/lossy.py
smastelini: I am not sure I follow the first part :P

Lossy Count is univariate in the sense that the input data is univariate. But the output is not. I have fixed the typing annotation and also other related errors.

I also renamed forgetting to alpha, as the authors also used that in the paper. Maybe `alpha` is the way to go with everything forgetting-related.

river/stats/lossy.py
smastelini: Done!

river/stats/lossy.py
```diff
@@ -108,12 +112,13 @@ def update(self, x: typing.Hashable):
...
-    def get(self):
+    def get(self) -> typing.Optional[typing.List[typing.Hashable]]:  # type: ignore
         res = []
         for key in self._entries:
             freq, _ = self._entries[key]
             if freq >= (self.support - self.epsilon) * self._delta:
                 res.append((key, freq))
-        return sorted(res, key=itemgetter(1), reverse=True)
+        if res:
+            return [elem[0] for elem in sorted(res, key=operator.itemgetter(1), reverse=True)]
```
MaxHalford, line 124: I would always return a list, that will simply the return type annotation
```suggestion
        return [elem[0] for elem in sorted(res, key=operator.itemgetter(1), reverse=True)]
```

MaxHalford (approved):
Approved with one suggestion!

Good job :)

river/stats/lossy.py
smastelini: I had to do that because of a failing test :)

I think it was the test that ensures `repr` is valid when the stat was not yet updated.

smastelini: When returning `[]` the test fails so I made it return `None` when no data was observed. What are your thoughts about that @MaxHalford? 

Keep `None` or should we update the test?

MaxHalford: Mmm I see. Let's go with `None`, it's fine :)

PR #1035, opened 2022-09-21
Update README to add instructions on Rust installation before installing River >= 0.13.0

hoanganhngo610:


hoanganhngo610:
Tagging @MaxHalford for a quick pass!

MaxHalford:
Actually users don't need to install Rust. We build wheels where the Rust part is precompiled into an .so file. However, developers have to install Rust, and those instructions are in the `CONRTIBUTING.md` file. Is that clearer?

hoanganhngo610:
My bad! I was installing on my development environment using `pip`, which is maybe the reason why it caused such problems. When I tested on a totally fresh environment it worked fine. Will close the PR now!

PR #1042, opened 2022-10-02
Fix syntax pypi.yml

AdilZouitine:
It works with this syntax error but it's cleaner now.


MaxHalford:
@AdilZouitine can you redo the PR? I think something went wrong because I'm cleaning the repo

PR #1044, opened 2022-10-02
Fix syntax pypi.yml

AdilZouitine:
It works with this syntax error but it's cleaner now.


AdilZouitine:
@MaxHalford is Ok now ?

MaxHalford:
No I don't think so, because it's deleting a notebook. Did you reclone the repo from scratch?

PR #1045, opened 2022-10-02
Fix syntax pypi.yaml

AdilZouitine:
The GitHub action worked, but the syntax was not good; now, it's cleaner.


MaxHalford:
Cheers :)

PR #1049, opened 2022-10-03
fix docstrings

abonte:
Fix latex math in docstrings.


MaxHalford:
Thanks a lot @abonte :)

PR #1052, opened 2022-10-06
Fix bug that DDM reports drift on data stream with only ones

3tothe6:
Consider a data stream with only ones and no zeros. When the code block inside `if n > self.warm_start:` in method `river.drift.DDM.update` is entered the first time, `p_i`, `s_i`, `self._p_min`, `self._s_min` will be 1.0, 0.0, 1.0, 0.0 respectively, then `p_i + s_i >= self._p_min + self.drift_threshold * self._s_min` will hold and a drift will be reported. I fix this bug by replacing `>=` with `>` and add the corrosponding testcase to `river/drift/test_drift_detectors.py`.

### DIFF
--- a/river/drift/ddm.py
+++ b/river/drift/ddm.py
@@ -141,12 +141,12 @@ class DDM(DriftDetector):
                 self._s_min = s_i
                 self._ps_min = self._p_min + self._s_min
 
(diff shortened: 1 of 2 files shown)


smastelini:
Let's wait for the CI and then I'll merge the PR. Thanks for the fix!

PR #1057, opened 2022-10-08
Make bandit module public

MaxHalford:
Ok, here's a rather big PR to make the `bandit` module public.

Until now, we had some bandit method in the `_bandit` method. We used those in `model_selection` to implement model selection based on bandit logic.

I refactored a bunch of stuff to make the bandits generic. They should be usable in all sorts of situation, with model selection being a special case. In particular, the `reward_obj` within each bandit can be a `proba.base.Distribution`, a `stats.base.Statistic`, or a `metrics.base.Metric`. To make this possible, I made it so that each one of these implements a `>` operator to compare two instances. That was pretty straightforward for metrics and statistics because both have a `get` method. For distribution I added a `mode` property to each distribution.

I also added a (tested) gym environment to reproduce the Candy Cane Competition from Kaggle. Right now it only works for one player, but that can be changed in a future PR.

I added logic to unit test all the (policies x environments) to make sure each policy performs better than a random policy.

The next big steps are:

- Implement a multiplayer version of the candy cane contest, and show how to use bandits in an adversarial situation.
- Implement a contextual bandit, such as LinUCB

I'll add these to the roadmap :)

@reviewers I don't expect you to review all the code. It would be just be cool if you can skim over the bandit API and raise any flags if you see any. Cheers!


smastelini:
Nice work @MaxHalford ! I can give it a look tomorrow. If you are happy with it and in a hurry you can merge and any problems we might find can be fixed later :)

MaxHalford:
There's no hurry at all :)

river/bandit/thompson.py
```diff
@@ -0,0 +1,86 @@
...
+    @property
+    def dist(self):
+        return self.reward_obj
+
+    def _pull(self, arms):
+        return max(arms, key=lambda arm: self._rewards[arm].sample())
+
+    @classmethod
+    def _unit_test_params(cls):
+        yield {"dist": proba.Beta()}
```
AdilZouitine, line 86: I found it confusing to have a private class method only useful for unit testing. Is there an alternative?

river/bandit/ucb.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+            if (reward := self._rewards.get(arm)) is not None
+            else math.inf
+            for arm in arms
+        }
+        return max(arms, key=lambda arm: upper_bounds[arm])
+
+    @classmethod
+    def _unit_test_params(cls):
+        yield {"delta": 1}
```
AdilZouitine, line 80: Same comment as other bandits

river/metrics/base.py
```diff
@@ -52,6 +52,9 @@ def is_better_than(self, other) -> bool:
         op = operator.gt if self.bigger_is_better else operator.lt
         return op(self.get(), other.get())
 
+    def __gt__(self, other):
+        return self.is_better_than(other)
+
```
AdilZouitine, line 57: Love it a lot!

river/model_selection/bandit.py
```diff
@@ -0,0 +1,113 @@
...
+        return self
+
+    @classmethod
+    def _unit_test_params(cls):
+        for params in super()._unit_test_params():
+            yield {**params, "policy": bandit.EpsilonGreedy(0.2)}
+
+    def _unit_test_skips(self):
+        return (
+            {"check_model_selection_order_does_not_matter"} if hasattr(self.policy, "seed") else {}
```
AdilZouitine, line 112: Same comment as the bandit unit testing class method

setup.py
```diff
@@ -57,6 +57,7 @@
             "black>=22.1.0",
             "flake8>=4.0.1",
             "graphviz>=0.10.1",
+            "gym>=0.26.1",
```
AdilZouitine, line 60: Gym has been quite unstable for the last couple of months. We should freeze the version for the following month to wait for the gym team to stabilize the API.

AdilZouitine (commented):
It's a great pull request, and I think we can quickly expand it for multiple players. 
I made some minor remarks!
I can't wait to see this pull request merge 😄

setup.py
MaxHalford: Sure! What version do you suggest?

river/bandit/base.py
```diff
@@ -0,0 +1,106 @@
...
+    ----------
+    reward_obj
+        The reward object used to measure the performance of each arm. This can be a metric, a
+        statistic, or a distribution.
+    burn_in
+        The number of steps to use for the burn-in phase. Each arm is given the chance to be pulled
+        during the burn-in phase. This is useful to mitigate selection bias.
+    seed
+        Random number generator seed for reproducibility.
+
```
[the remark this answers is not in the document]
MaxHalford: Good catch

river/bandit/base.py
```diff
@@ -0,0 +1,106 @@
...
+RewardObj = Union[stats.base.Statistic, metrics.base.Metric, proba.base.Distribution]
+
+
+class Policy(base.Base, abc.ABC):
+    """Bandit policy base class.
+
+    Parameters
+    ----------
+    reward_obj
+        The reward object used to measure the performance of each arm. This can be a metric, a
```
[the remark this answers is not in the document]
MaxHalford: I think so too! A nice trick is that you pass a rolling object if you want your bandit to handle drift. I'll document that at some point :)

river/bandit/epsilon_greedy.py
```diff
@@ -0,0 +1,87 @@
...
+    def _pull(self, arms):
+        return (
+            self._rng.choice(arms)  # explore
+            if self.best_arm is None or self._rng.random() < self.current_epsilon
+            else self.best_arm  # exploit
+        )
+
+    @classmethod
+    def _unit_test_params(cls):
+        yield {"epsilon": 0.2}
```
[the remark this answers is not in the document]
MaxHalford: Why is that?

We need some mechanism to specify unit test params and unit tests to skip. Would you rather define them all in a single file? I did this it this way following scikit-learn, which does this for skipping unit tests.

gbolmier (commented):
I got confused by the `arm` naming convention because it actually is the arm "id", "index", "label" (or whatever we want to call it) and not the arm itself. Otherwise LGTM, great work!

setup.py
gbolmier: Let's cap and not freeze then @AdilZouitine 👍

MaxHalford:
Yes good point @gbolmier. I added a type hint to make this a bit clearer

gbolmier:
Yeah, the type hint actually saved my sanity lol

> Would you rather call it `arm_index`?

That would be clearer IMO. I'm fine with `arm_index` but why not `arm_id` or `arm_ix` if people find the former too verbose.

MaxHalford:
I'll think about it. The thing about `index` is that it implies (to me) that the value is an integer. But the arm can also be a `str`. In fact, the arm can be anything hashable.

gbolmier:
Yep, that's why I mentioned `arm_id`.

river/bandit/thompson.py
```diff
@@ -0,0 +1,86 @@
...
+
+    """
+
+    def __init__(self, dist: proba.base.Distribution, burn_in=0, seed: int = None):
+        super().__init__(dist, burn_in)
+        self.seed = seed
+        self._rng = random.Random(seed)
+        self._rewards.default_factory = self._clone_dist_with_seed
+
+    def _clone_dist_with_seed(self):
```
smastelini, line 74: I did that in the Periodic drift detector :D

smastelini: Soft gambiarra ;)

smastelini:
I don't know much about bandits and reinforcement learning, so I'll leave the technical bit for the experts :)

But design-wise, LGTM! I really like the `BanditRegressor` class and I think a lot of people will like that too.

setup.py
AdilZouitine: Yes I think is a good option!

AdilZouitine:
> Ok, I took into account your suggestion, @gbolmier. You're right I think it looks better :)
> 
> @AdilZouitine feel free to suggest if you come up with a better design for test default parameters, we can address it in a future PR :)

Yes, I'll think about it 😄

PR #1058, opened 2022-10-11
Fix docs

smastelini:

### DIFF
--- a/river/bandit/thompson.py
+++ b/river/bandit/thompson.py
@@ -15,11 +15,15 @@ class ThompsonSampling(bandit.base.Policy):
     distribution, but should rather be defined in the policy parametrization. In other words, you
     should do this:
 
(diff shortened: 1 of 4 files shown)


river/bandit/thompson.py
```diff
@@ -15,11 +15,15 @@ class ThompsonSampling(bandit.base.Policy):
     distribution, but should rather be defined in the policy parametrization. In other words, you
     should do this:
 
-    >>> policy = ThompsonSampling(dist=proba.Beta(1, 1), seed=42)
+    ```
```
MaxHalford, line 18: nit: I think ```py is better :)

smastelini: I don't know why, but it didn't work, at least locally. "py" was rendered in the docstring :/

MaxHalford: No worries. The docs looks cool now!

PR #1059, opened 2022-10-12
Implement bandit.evaluate

MaxHalford:
I called it `evaluate` to be consistent with the existing method in the `time_series` module.

I'll make use of this function in a future user guide to reproduce the examples from Sutton and Barto :)

@AdilZouitine can you take a quick pass? Not sure I'm using/copying Gym environments correctly.

### DIFF
--- a/river/bandit/evaluate.py
+++ b/river/bandit/evaluate.py
@@ -1,0 +1,144 @@
+import copy
+import random
+from typing import Callable, List
+
+try:
+    import gym
+except ImportError:
+    ...
+from river import bandit
+from river import stats
+
+def evaluate(policies: List[bandit.base.Policy], env: "gym.Env", pull_func: Callable[[bandit.base.Policy, "gym.Env"], bandit.base.ArmID], reward_stat: stats.base.Univariate = None, n_episodes: int = 20, seed: int = None):
+    """Benchmark a list of policies on a given Gym environment.
+
+    This is a high-level utility function for benchmarking a list of policies on a given
+    Gym environment. For example, it can be used to populate a `pandas.DataFrame` with the contents
+    of each step of each episode.
+
+    Parameters
+    ----------
+    policies
+        A list of policies to evaluate. The policy will be reset before each episode.
+    env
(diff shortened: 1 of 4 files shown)


river/bandit/evaluate.py
```diff
@@ -0,0 +1,153 @@
...
+    pull_func
+        A function that takes a policy and an environment as arguments and returns the arm that was
+        pulled. This function is called at each time step for each policy. This is required because
+        there is no standard way to pull an arm in Gym environments.
+    reward_stat
+        A univariate statistic to keep track of the rewards. This statistic will be reset before
+        each episode. Note that this is not the same as the reward object used by the policies.
+        It's just a statistic to keep track of each policy's performance. If `None`, `stats.Sum`
+        is used.
+    n_episodes
```
AdilZouitine, line 42: I think it is the best choice, and `Sum` will be used 99% of the time for sure

river/bandit/evaluate.py
```diff
@@ -0,0 +1,153 @@
...
+
+        while not done:
+            terminated, truncated = False, False
+            for i, (_policy, _env, _reward_stat) in enumerate(
+                zip(episode_policies, episode_envs, episode_reward_stats)
+            ):
+
+                action = pull_func(_policy, _env)
+                observation, reward, terminated, truncated, info = _env.step(action)
+                _reward_stat.update(reward)
```
AdilZouitine, line 141: You have several times the same environment instance sequentially.

In this case, one of the environments may end before the others.
This is not handled; the finished environment is often reset, and the steps are continued.

AdilZouitine (commented):
The Pr is excellent! There is a point that concerns me is the fact of not managing the rest of the episodes independently because it is possible that the episodes do not end at the same time

river/bandit/evaluate.py
MaxHalford: Ah yes good point. I've only worked with envs that end after a fixed number of episodes. But you're right, we should handle the general case.

Issue #1061, opened 2022-10-14
Question/proposal regarding ADWIN clock


mvieth:
Hi, I have a question about the clock implemented into the ADWIN drift detector (https://github.com/online-ml/river/blob/main/river/drift/adwin_c.pyx#L215). As I understand it, it will only check for drifts every 32nd time. My guess would be that this is done to increase the speed/performance of the detector, but I haven't found any mention about this in the original ADWIN paper. If you know more about this or can point me to a paper where the reason for this clock is given, that would be great.
My concern is that this clock period (32) will likely not fit for all scenarios, and there is no (easy) way for a user to change it. So I would propose to make this configurable, similar to the `delta` parameter. If you agree, I can create a pull request.
Thanks.

MaxHalford:
I believe we already discussed moving back to a Python implementation for the sake of maintainability. But we decided the performance gain is worth keeping.

I would say that the required change is small enough to justify adding it to the Cython implementation. Especially considering the OP is offering to make a pull request :)

smastelini:
I think the best compromise is to set `clock` based on the `bucket_size` then. I don't remember the details now, but I have a gut feeling these parameters should walk together :)

Feel free to open a PR @mvieth. I had missed that part hahaha

mvieth:
Are you sure that there is a connection between the clock and the max number of buckets? [In the MOA code](https://github.com/Waikato/moa/blob/master/moa/src/main/java/moa/classifiers/core/driftdetection/ADWIN.java), there is no apparent connection, and the clock is set to 32 (independently from `MAXBUCKETS`).
Let's say the max number of buckets is 5, then the 32 new samples since the last check for concept change would fill all 5 buckets of size 1, all 5 buckets of size 2, and 4 of the 5 buckets of size 4 (assuming I understood the algorithm correcty :smile: ).
Another example, let's say the max number of buckets is 3 (then `2 ** 3 = 8`). The 8 samples since the last check for concept change would fill all 3 buckets of size 1, and almost but not completely all buckets of size 2.
I just don't see any logic in this. Could it be that it was just a coincidence that `MAXBUCKETS` was set to 5 and the clock to 32 in the MOA code, and they don't have any connection?

smastelini:
Hi @mvieth. No, I'm not sure, TBH. Just recently I started delving a bit around the concept drift world, out of necessity 😅.

What I know is that both variables are hardcoded in the MOA code, and the paper does not defines`clock`.
[...]

PR #1062, opened 2022-10-18
[drift] Expose more parameters of adwin

mvieth:
As discussed in https://github.com/online-ml/river/issues/1061

# Benchmarks
I used the example from the adwin.py documentation (drift at index 1000).

## Varying `clock`
Fixed: `delta=0.002, max_buckets=5, min_window_length=5, grace_period=10`
| clock                    | 1    | 2    | 4    | 8    | 16   | 32   | 64   |
|--------------------------|------|------|------|------|------|------|------|
| processing time (ms)     | 49   | 29   | 16   | 11   | 8    | 7    | 6    |
| change detected at index | 1005 | 1005 | 1007 | 1007 | 1007 | 1023 | 1023 |

This shows nicely that the processing time decreases but the change detection delay increases with a higher `clock`.

## Varying `max_buckets`
Fixed: `delta=0.002, clock=16, min_window_length=5, grace_period=10`
| max_buckets              | 2    | 3    | 4    | 5    | 6    | 7    | 8    | 9    |
|--------------------------|------|------|------|------|------|------|------|------|
| processing time (ms)     | 5    | 6    | 8    | 9    | 10   | 11   | 12   | 12   |
| change detected at index | 1039 | 1023 | 1039 | 1007 | 1023 | 1023 | 1007 | 1007 |

This shows that processing time increases slightly but change detection has less delay/is more fine-grained with higher `max_buckets`.

I couldn't think of any sensible experiments for `min_window_length` and `grace_period`, but the effects should be as I described in the documentation.

The four new parameters should all be at least 1, but I am not sure if it is necessary to check the user-specified value since that should IMO be clear from the documentation.

### DIFF
--- a/river/drift/adwin.py
+++ b/river/drift/adwin.py
@@ -19,6 +19,14 @@ class ADWIN(DriftDetector):
     ----------
     delta
         Significance value.
+    clock
+        How often ADWIN should check for change. 1 means every new data point, default is 32. Higher values speed up processing, but may also lead to increased delay in change detection.
+    max_buckets
+        The maximum number of buckets of each size that ADWIN should keep before merging buckets (default is 5).
+    min_window_length
+        The minimum length of each subwindow (default is 5). Lower values may decrease delay in change detection but may also lead to more false positives.
+    grace_period
+        ADWIN does not perform any change detection until at least this many data points have arrived (default is 10).
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


MaxHalford:
Great work! Thanks for adding docstrings.

> The four new parameters should all be at least 1, but I am not sure if it is necessary to check the user-specified value since that should IMO be clear from the documentation.

I agree. It's better to document properly than to test user inputs. The latter can become a nightmare 😱

river/drift/adwin.py
```diff
@@ -19,6 +19,14 @@ class ADWIN(DriftDetector):
...
     delta
         Significance value.
+    clock
+        How often ADWIN should check for change. 1 means every new data point, default is 32. Higher values speed up processing, but may also lead to increased delay in change detection.
+    max_buckets
+        The maximum number of buckets of each size that ADWIN should keep before merging buckets (default is 5).
+    min_window_length
+        The minimum length of each subwindow (default is 5). Lower values may decrease delay in change detection but may also lead to more false positives.
+    grace_period
+        ADWIN does not perform any change detection until at least this many data points have arrived (default is 10).
```
MaxHalford, line 29: Could you make these 100 characters max? It's a soft rule we have in the codebase.

MaxHalford (approved):

PR #1073, opened 2022-10-27
fix docstring

abonte:


MaxHalford:
Actually we don't record FLOPS yet, so the best would be to remove this :) (or implement FLOPS counting hehe)

abonte:
I just wanted to fix this formatting error :)

docs

PR #1077, opened 2022-11-04
Fix possible bug in hddm_w

bdshoener:
I noticed what appears to be a bug in HDDM_W when two_sided_test is True. Because of how "_has_mean_changed" works, the check for a change when the mean _decreases_ always defaults to True. I noticed that the recent refactoring removed two lines that, in essence, checked if sample 1 or sample 2 had just been initialized (i.e., lines 149 and 150: 
```
if sample1.EWMA_estimator < 0 or sample2.EWMA_estimator < 0:
            return False
```
)

This pull request fixes that, if it indeed is a bug.

### DIFF
--- a/river/drift/hddm_w.py
+++ b/river/drift/hddm_w.py
@@ -145,6 +145,8 @@ class HDDM_W(DriftDetector):
     def _has_mean_changed(
         self, sample1: "SampleInfo", sample2: "SampleInfo", confidence: float
     ) -> bool:
+        if sample1.is_initialized or sample2.is_initialized:
+            return False
         ibc_sum = sample1.ibc + sample2.ibc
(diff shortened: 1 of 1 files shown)


smastelini:
Hi @bdshoener, thanks for reporting that. There are some problems with your suggested change, though:

1. I find the `_is_initialized = True`pattern misleading because the variable is set to `True` at the creation and then to `False`once it is updated. I would rather initialize it with `False` and check `if not (s1._is_init and s2._is_init)`. Moreover, `_is_init` could be a shortened but still self-explanatory variable name.
2. You define the `_is_initialized` variable but access it as `is_initialized` (no underscore). This is a typo :)

Could you perform these updates?

PR #1085, opened 2022-11-09
Add pyupgrade

kulbachcedric:
Hi @MaxHalford, @gbolmier, @smastelini ,

I added `pyupgrade` to the `.pre-commit-config.yaml` and ran it once. 
As I'm unsure to what extent it is useful to add `pyupgrade` and there are many changes, I have created a draft PR.
Most of the changes would affect the type-hints and the declaration of sets and tuples. 

To sum it up, feel free to comment or decline this PR ;-).

Best,
Cedric

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,9 @@ repos:
         language: python
         types: [python]
         entry: mypy
+
+      - id: pyupgrade
+        name: pyupgrade
+        language: python
+        types: [ python ]
+        entry: pyupgrade --py38-plus
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
         name: mypy
         language: python
         types: [python]
-        entry: mypy --no-strict-optional
+        entry: mypy --no-strict-optional --check-untyped-defs
 
       - id: pyupgrade
         name: pyupgrade
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
         name: mypy
         language: python
         types: [python]
-        entry: mypy
+        entry: mypy --no-strict-optional
 
       - id: pyupgrade
         name: pyupgrade
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,9 @@ repos:
         language: python
         types: [python]
         entry: mypy --implicit-optional
+
+      - id: pyupgrade
+        name: pyupgrade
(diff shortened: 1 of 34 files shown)


MaxHalford:
Actually the fact that it moves our codebase to the modern way of doing type annotations is really cool!

MaxHalford:
I see no downsides. What do you think @gbolmier?

kulbachcedric:
> Actually the fact that it moves our codebase to the modern way of doing type annotations is really cool!

All tests are failing 🤯
So I keep fixing the tests for the changes for `python 3.8 upwards`? :-)

gbolmier:
I don't see any downsides either except adding yet another dev tool. I'm not against it as long as it is well maintained (seems to be ✅) and not adding friction to the dev workflow.

Was mypy failing before the changes introduced by pyupgrade?

MaxHalford:
Yes @kulbachcedric you should fix the tests.

And no mypy was not failing, so I don't see why it's configuration should be changed.

kulbachcedric:
> Yes @kulbachcedric you should fix the tests.

sure! ☺️
> 
> And no mypy was not failing, so I don't see why it's configuration should be changed.

In the main branch I still get numerous error messages with mypy e.g.
`river/model_selection/greedy.py:51: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed in 0.990 its default to no_implicit_optional=True`

When setting `mypy --implicit-optional --check-untyped-defs` there are still the following error messages:

```
river/utils/rolling.py:31: error: Value of type "Rollable" is not indexable  [index]
river/stream/utils.py:19: error: Incompatible types in assignment (expression has type "TextIOWrapper", variable has type "IO[bytes]")  [assignment]
river/stream/utils.py:30: error: "object" not callable  [operator]
```

Do we stay on an older mypy version or do we migrate?

gbolmier:
On this PR, mypy is actually only failing for:

```sh
river/tree/hoeffding_tree.py:354: error: Return value expected  [return-value]
```

which is easy to fix. Other complaints are just suggestions to use the `--check-untyped-defs` option.

I'm getting more errors too on the main branch due to the new mypy. Either someone could open a PR to fix them or we can gradually fix them when the concerned files are modified.

> Do we stay on an older mypy version or do we migrate?

We should use the latest version as long as it is not broken.

kulbachcedric:
Do you know how to fix the Codeball issue?

Adding `pyupgrade` works now within the `pre-commit-config.yml`☺️

MaxHalford:
Forget Codeball, it's drunk

MaxHalford (approved):

PR #1089, opened 2022-11-10
update mypy

kulbachcedric:
Hi,

here is the PR for updating mypy.
There are 4 notes left when running mypy:
```
river/drift/kswin.py:109: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
river/tree/nodes/sgt_nodes.py:44: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
river/tree/nodes/hatc_nodes.py:139: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
river/tree/stochastic_gradient_tree.py:55: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```

However, when setting `mypy  --check-untyped-defs` as considered, we are facing a lot more errors...
So I left it like that for the beginning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,4 +33,4 @@ repos:
         name: mypy
         language: python
         types: [python]
-        entry: mypy
+        entry: mypy --implicit-optional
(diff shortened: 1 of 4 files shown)


MaxHalford:
Cheers!

PR #1112, opened 2022-12-07
Assert Graphviz import

kulbachcedric:
Hi,

the Hoeffding trees require [Graphviz](https://graphviz.org/download/) to be visualised. Therefore, I would suggest to assert whether it can be imported or not.

Best

### DIFF
--- a/river/tree/hoeffding_tree.py
+++ b/river/tree/hoeffding_tree.py
@@ -17,11 +17,6 @@ from .nodes.branch import (
 )
 from .nodes.leaf import HTLeaf
 
-try:
-    import graphviz
-except ImportError as e:
-    raise ValueError("You have to install graphviz to use the draw method,") from e
-
 
 class HoeffdingTree(ABC):
     """Base class for Hoeffding Decision Trees.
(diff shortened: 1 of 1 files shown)


MaxHalford:
Hey hey.

It's [more Pythonic](https://realpython.com/python-lbyl-vs-eafp/) to let the exception raise itself. Here at line 430, an exception will be raised if graphviz is not installed. Did that not happen to you?

MaxHalford:
Well we do want to keep the graphviz package optional and not bother our users. If you don't find the import error explicit enough, we can do this in the `draw` method:

```py
try:
    import graphviz
except ImportError:
    raise ValueError("You have to install graphviz to use the draw method") from e
```

river/tree/hoeffding_tree.py
```diff
@@ -405,6 +400,10 @@ def draw(self, max_depth: int = None):
         .. image:: ../../docs/img/dtree_draw.svg
             :align: center
         """
+        try:
+            import graphviz
+        except ImportError as e:
+            raise ValueError("You have to install graphviz to use the draw method,") from e
```
MaxHalford, line 406: Is the comma `,` necessary?

kulbachcedric: Should be a `.` 🙈

PR #1117, opened 2022-12-13
[datasets.synth] Remove unnecessary numpy usage and update np.random.RandomState to np.random.default_rng where necessary

smastelini:


MaxHalford:
Cool @smastelini! Why you're at it, can you also remove the `check_random_state` in `MonteCarloClassifierChain`? That way we can remove the `check_random_state` function altogether.

smastelini:
Yeap! That's the next step :) 

I also wanted to move the remaining functions elsewhere, or at least rename the file. But I lack a good idea.

Issue #1116, opened 2022-12-13
Error when using OutputCodeClassifier for Code-size greater than 20


Yasmen-Wahba:
Hello again,

I was testing the OCC classifier with more than 90 classes and the accuracy is very poor. I assume I need a huge code size, however I was testing different code-sizes (staring with a code-size of 10) and recording the accuracy when I came to a code-size of 40 and received the following error: 
`OverflowError: Python int too large to convert to C ssize_t`. Is there a way we can modify the occ classifier to allow for compact codes (as short as possible) while still providing enough discriminating power between the different classes.

MaxHalford:
Hey there @Yasmen-Wahba. Would you be able to share some code and a dataset so we can reproduce what you're doing? Also, does a OVR classifier provide good performance or not? If it doesn't then OCC usually won't do better.

Yasmen-Wahba:
The OVR crashes. Unfortunately, the dataset is confidential, but a similar dataset would be the consumer complaints benchmark published by the Consumer Financial Protection Bureau. <https://www.consumerfinance.gov/data-research/consumer-complaints/>
Since it's a hierarchical problem, I'm flattening the classes to be able to use River. For the Consumer Complaints, you could use the Product and Sub-product columns.

MaxHalford:
I will look into this sometime this week :)

MaxHalford:
So I downloaded the complaints dataset. However, I'm unsure what would be the target. Can you tell me?

```py
{'Date received': '2022-12-12',
  'Product': 'Credit reporting, credit repair services, or other personal consumer reports',
  'Sub-product': 'Credit reporting',
  'Issue': "Problem with a credit reporting company's investigation into an existing problem",
  'Sub-issue': 'Their investigation did not fix an error on your report',
  'Consumer complaint narrative': '',
  'Company public response': '',
  'Company': 'Experian Information Solutions Inc.',
  'State': 'NC',
  'ZIP code': '27356',
  'Tags': '',
  'Consumer consent provided?': '',
  'Submitted via': 'Web',
  'Date sent to company': '2022-12-12',
  'Company response to consumer': 'In progress',
  'Timely response?': 'Yes',
  'Consumer disputed?': 'N/A',
  'Complaint ID': '6311723'}
```

By the way, to answer your question:

> Is there a way we can modify the occ classifier to allow for compact codes (as short as possible) while still providing enough discriminating power between the different classes.

[...]

PR #1127, opened 2022-12-26
Add coding_method parameter to OCC

MaxHalford:
@Yasmen-Wahba I found the issue to #1116. The problem was that the list of all possible codes were generated in memory and then sampled from. This takes too memory when the code size is big, say 30. I added a way to choose what code sampling method to use. The `'random'` method is what scikit-learn does, while the `'exact'` method is what we did before.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -19,6 +19,10 @@
 
 - Renamed `use_dist` to `with_dist` in `linear_model.BayesianLinearRegression`'s `predict_one` method.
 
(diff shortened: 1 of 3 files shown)


smastelini:
LGTM!

MaxHalford:
Thanks for the swift review @smastelini. I'm just fixing the tests and then I'll merge :)

MaxHalford:
Close #1116.

MaxHalford:
Closes #1116.

PR #1131, opened 2022-12-30
Onelearn implementation: AMF & Mondrian Tree Classifiers

AlexandreChaussard:
## Description of the PR

Hi ! 👋 

This is a first version of [Onelearn](https://onelearn.readthedocs.io/en/latest/)'s library (classifiers only) implementation in River.
It contains:
- Mondrian Tree (Base and Classifier)
- Aggregated Mondrian Forest (Base and Classifier)

The original repository with proof of working implementation can be found [here](https://github.com/AlexandreChaussard/AMFLearning) (see [script.py](https://github.com/AlexandreChaussard/AMFLearning/blob/master/river/script.py)).

image

## Notes on the utils
Currently I placed two functions in the utils section:
- ~sample_discrete~
- log_sum_2_exp

It might seem overkill to place them as utils right now looking at where they're used in the code, but I'll need them for the regressors too when the times come. Maybe there's a better place for them keeping the regressors in mind though.


MaxHalford:
Hey hey :)

I don't have much time right now to check the code, but maybe @smastelini is available? I'll answer some of your questions though:

> Management of labels: labels must be positive integers at the moment, since it just makes my life easier. Please tell me how you would prefer this to be implemented in River's framework (I've seen labels presented as dictionary, but I'm not sure what these dictionaries contains: int? string?). This might be a simple trick for me to change how labels are managed I think, just need to know your preferences.

Trees are usually multi-class classifiers. We expact multi-class classifiers to support any label of type `bool`, `int` or `str` (see [`base/typing.py`](https://github.com/online-ml/river/blob/main/river/base/typing.py#L6))

> Examples of usage: MondrianTreeClassifier and AMFClassifier would need examples of implementations for users. I actually implemented one on my repo [here](https://github.com/AlexandreChaussard/AMFLearning/blob/master/river/example.py?rgh-link-date=2022-12-30T16%3A18%3A03Z) already, but I didn't manage to compile River so I couldn't get the scores 😢

What do you mean you didn't manage to compile? Yeah some doctests would be nice, at the minimum.

river/utils/random.py
```diff
@@ -27,3 +30,38 @@ def poisson(rate: float, rng=random) -> int:
         p *= rng.random()
 
     return k - 1
+
+
+def sample_discrete(distribution: list[float]) -> int:
```
MaxHalford, line 35: Does [`random.choices`](https://docs.python.org/3/library/random.html#random.choices) not suit what you need?

river/tree/mondrian_tree.py
```diff
@@ -0,0 +1,53 @@
+from abc import ABC
```
MaxHalford, line 1: In terms of file structure, I would created a sub-folder called `mondrian` in the `tree` folder

river/ensemble/aggregated_mondrian_forest.py
```diff
@@ -0,0 +1,444 @@
...
+    def loss(self):
+        """str: The loss used for the computation of the aggregation weights."""
+        return "log"
+
+    @loss.setter
+    def loss(self, val):
+        pass
+
+    def __repr__(self):
+        return f"AMFClassifier[n_classes={self.n_classes}; n_features={self.n_features}; n_models={self.n_estimators}]"
```
MaxHalford, line 444: You don't have to implement this. By default, a River model inherits from `base.Estimator`, which has a default `__repr__` method.

river/ensemble/aggregated_mondrian_forest.py
```diff
@@ -0,0 +1,444 @@
...
+                raise ValueError("`n_classes` must be >= 2")
+            else:
+                self._n_classes = val
+
+    @property
+    def dirichlet(self):
+        """float or None: Regularization level of the class frequencies."""
+        return self._dirichlet
+
+    @dirichlet.setter
```
MaxHalford, line 420: In terms of code-style, it's preferable not to use `setter` and instead validate the parameters in the `__init__` method. In fact, we usually try to avoid doing any validation, and instead properly document the class.

river/ensemble/aggregated_mondrian_forest.py
```diff
@@ -0,0 +1,444 @@
...
+            loss: str,
+            use_aggregation: bool,
+            split_pure: bool,
+            random_state: int,
+    ):
+        """Instantiates a `AMFLearner` instance.
+
+        Parameters
+        ----------
+        n_estimators : int
```
MaxHalford, line 31: No need to specify the type of each attribute. We determine the types from the type annotations in the `__init__` method.

MaxHalford (changes requested):

river/utils/random.py
AlexandreChaussard: Thanks, I removed the "sample_discrete" function in latest commit and used "random.choices" instead.

river/tree/mondrian_tree.py
AlexandreChaussard: I added a "mondrian" tree folder and moved the scripts related to the Mondrian Trees into that one

smastelini:
Hi @AlexandreChaussard and @MaxHalford 😃 

I can take care of this review. It's going to be a good opportunity to learn more about AMF. 

I'm currently enjoying the holidays with my family, but I will start reviewing next week!

If needed, besides leaving comments, I can create a PR to @AlexandreChaussard's fork with some suggestions.

river/ensemble/aggregated_mondrian_forest.py
AlexandreChaussard: Removed!

river/ensemble/aggregated_mondrian_forest.py
AlexandreChaussard: What do you mean by "validate the parameters" ? Like setting them within the `__init__`  ?
I removed all the @setter and @getter in latest commit, it makes the code easier to read as well.

MaxHalford:
Yes, let's take the time to review this properly, there's no rush. This has the potential to be one of our best algorithms, so it's worth spending time on it.

river/ensemble/aggregated_mondrian_forest.py
AlexandreChaussard: I think I fixed that one looking at the HoeffdingTree, please let me know if there's anything to change =)

AlexandreChaussard:
Hey ! 👋 

I've been fixing all the previously stated points, besides the example docstring that I keep for when the model will be completely reviewed and ready to go 😄

Happy new years team ❤️

river/tree/__init__.py
```diff
@@ -72,4 +73,5 @@
     "LabelCombinationHoeffdingTreeClassifier",
     "SGTClassifier",
     "SGTRegressor",
+    "MondrianTreeClassifier",
```
smastelini, line 76: Let's keep the imports and the placement of class names inside `__all__` in alphabetical order.

river/tree/mondrian/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+The `river.tree.mondrian` module includes learning and split node
+implementations for the Mondrian trees.
+"""
```
smastelini, line 4:
```suggestion
"""

```

river/tree/mondrian/mondrian_tree.py
```diff
@@ -0,0 +1,47 @@
...
+    their own design.
+
+    Parameters
+    ----------
+    n_features
+        Number of features
+    step
+        Step parameter of the tree
+    loss
+        Loss to minimize for each node of the tree
```
smastelini, line 20: Are there other options already available?

river/tree/mondrian/mondrian_tree.py
```diff
@@ -0,0 +1,47 @@
...
+        Pick between: "log", ...
+    use_aggregation
+        Should it use aggregation
```
smastelini, line 23:
```suggestion
        Whether or not the tree should it use aggregation.
```

river/tree/mondrian/mondrian_tree.py
```diff
@@ -0,0 +1,47 @@
...
+    split_pure
+        Should the tree split pure leafs when training
```
smastelini, line 25:
```suggestion
        Whether or not the tree should split pure leaves when training.
```

river/tree/mondrian/mondrian_tree.py
```diff
@@ -0,0 +1,47 @@
...
+    iteration
+        Number of iterations to run when training
```
smastelini, line 27:
```suggestion
        Number of iterations to run when training.
```

river/utils/math.py
```diff
@@ -353,3 +354,27 @@ def woodbury_matrix(A: np.ndarray, U: np.ndarray, V: np.ndarray):
...
     Au = A @ U
     A -= Au @ np.linalg.inv(eye + V @ Au) @ V @ A
+
+
+def log_sum_2_exp(a: float, b: float) -> float:
+    """Computation of log( (e^a + e^b) / 2) in an overflow-proof way
+
+    Parameters
+    ----------
+    a : float
```
smastelini, line 364: There is no need to specify the parameter type here :D

river/utils/math.py
```diff
@@ -353,3 +354,27 @@ def woodbury_matrix(A: np.ndarray, U: np.ndarray, V: np.ndarray):
...
+        First number
+
+    b : float
```
smastelini, line 367: same

river/utils/math.py
```diff
@@ -353,3 +354,27 @@ def woodbury_matrix(A: np.ndarray, U: np.ndarray, V: np.ndarray):
...
+        Second number
+
+    Returns
+    -------
+    output : float
+        Value of log( (e^a + e^b) / 2) for the given a and b
+    """
+    # TODO: if |a - b| > 50 skip
+    # TODO: try several log and exp implementations
```
smastelini, line 376: Is this still valid?

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
...
+            time: float
+    ):
+
+        super().__init__()
+
+        # Generic Node attributes
+        self.parent = parent
+        self.time = time
+        self.is_leaf = True
+        self.depth = 0
```
smastelini, line 40: Does this change?

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
...
+        self.feature = 0
+        self.weight = 0.0
+        self.log_weight_tree = 0
+        self.threshold = 0.0
+        self.n_samples = 0
+        self.n_features = n_features
+        self.memory_range_min = [0] * n_features
+        self.memory_range_max = [0] * n_features
+
+    def copy(self, node):
```
smastelini, line 52: Can you elaborate on the usage of this one? Maybe `self.clone` can do the trick 🤔

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
...
+        Set the right child node
+        Parameters
+        ----------
+        node
+            Child node to be set as right node
+
+        Returns
+        -------
+
+        """
```
smastelini, line 100: Python properties are the preferred strategy to use :D

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
+import typing
+
+from abc import ABC, abstractmethod
+
+from river.utils.math import log_sum_2_exp
+
+from math import log
```
smastelini, line 7: I suggest importing math and using `math.log`

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
...
+        ----------
+        sample_class
+            Class for which we want the score
+
+        dirichlet
+            Dirichlet parameter of the tree
+
+        Notes
+        -----
+        This uses Jeffreys prior with dirichlet parameter for smoothing
```
smastelini, line 249:
```suggestion
        This uses Jeffreys prior with Dirichlet parameter for smoothing.
```

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -0,0 +1,474 @@
...
+
+        Parameters
+        ----------
+        sample_class
+            A given class of the problem
+        dirichlet
+            Dirichlet parameter of the problem
+        """
+        sc = self.score(sample_class, dirichlet)
+        return -log(sc)
```
smastelini, line 283:
```suggestion
        return -math.log(sc)
```

smastelini (changes requested):
Hi @AlexandreChaussard, I started reviewing the code and am impressed by the effort you put into implementing AMF.

I barely scratched the surface yet and left some initial comments, mostly on style. One thing popped up though, I see you're keeping track of class numbers and expects a predefined number of classes. 

I suggest keeping defining a set of observed classes, as the Hoeffding Tree Classifier does. This way, new classes are allowed to appear and you can access the properties that are necessary to make predictions and perform other tasks.

If this is possible, the classes become any hashable object, as it is the case in the Hoeffding Trees. What do you think?

I'll keep reviewing your code

river/tree/mondrian/mondrian_tree.py
AlexandreChaussard: Not yet for the Classifier. I haven't checked yet for the Regressor, I'm just keeping it here in case the Regressor has different loss support (see MondrianTreeClassifier it doesn't have the loss parameter)

river/utils/math.py
AlexandreChaussard: This function is actually imported from Onelearn's implementation and would need review by a Computer Scientist maybe? I think it's meant to compute sums of exponentials without facing an Overflow Exception (see when the trees update) @smastelini 
If you think it's not relevant we can remove it =)

smastelini: My bad, I meant the # TODO in the code :D

AlexandreChaussard: Oh, that's what I meant too. These TODO are out of my reach tbh, so I kept them here in case somebody knows what it would mean to skip the |a - b| > 50 and different kinds of log & exp implementation to avoid the Overflow issue and still be able to compute the expected value.
I guess we can remove the TODO, but it might be good to remember the goal of this function and that it could apparently be improved?

smastelini: Oh, I see! Let's keep them, no problem :)

AlexandreChaussard:
> Hi @AlexandreChaussard, I started reviewing the code and am impressed by the effort you put into implementing AMF.

Thanks ❤️ 

> I suggest keeping defining a set of observed classes, as the Hoeffding Tree Classifier does. This way, new classes are allowed to appear and you can access the properties that are necessary to make predictions and perform other tasks.

Code-wise, I don't think this would be any problem to implement, there's already a dictionary collecting the observed classes that is called `_classes` in the classifiers. It is only used at the moment as a way to translate classes object into positive integers for easy computation later on.
However, it seems that the number of classes must be known before hand to compute the Jeffrey's prior (see description of AMF). If the number of classes could change over time, the current implementation would be incorrect (at first we would have 0 classes, so the dirichlet parameter would be discarded at the bottom and it'd be possible to have probabilities > 1 in the distribution using Jeffrey's prior).

If I'm making an error here, please let me know I'm no expert of AMF 😄 

> If this is possible, the classes become any hashable object, as it is the case in the Hoeffding Trees. What do you think?

The classes can be any string, int or boolean at the moment as @MaxHalford suggested (typing.ClfTarget). I'm not sure what you mean by hashable object instead (like something to put in an HashMap? It's kinda what does the dictionnary `_classes` right?)

I'm busy with exams at the moment, but I'll be working on all these on Thursday for sure!
Cheers 🎈

smastelini:
Yeap! You got it :) 

All these things (ints, etc.) can go in a set. Do worry about changing things now. Take your time with the exams. It's actually better this way, so that I can take some time to delve deep in the code until the end of the week. Cheers!

river/tree/mondrian/mondrian_tree_nodes.py
AlexandreChaussard: What are you targeting exactly ? I believe depth can change even after initialization, I'm not sure about the rest, but possibly it does (see when the "copy" function is being used, we push a node information into another).
I'd have to check it into the IDE though, I'm just throwing that here =)

river/tree/mondrian/mondrian_tree_nodes.py
AlexandreChaussard: Solved in latest commit.

river/tree/mondrian/mondrian_tree_nodes.py
AlexandreChaussard: I tried something different then using `property` and `variable.setter`. I'm not really an expert of OOP in Python, let me know if it's ok to you =)
It makes the code a lot more readable actually!

river/tree/mondrian/mondrian_tree_nodes.py
AlexandreChaussard: This function actually serves to copy characteristics of the node that are not related to samples into a child node when splitting. I could probably use `self.clone` and reset the characteristics that are related to samples the other way around (like instead of saying `child.copy(parent)` I would say `child = parent.clone(); child.n_samples = 0; child.memory_range_...`). Please tell me what would suit you best =)

For my part, I like the logic of having that "copy" function here because it's used for the Regressor as well, and it really is how I want nodes to be copied into another when passing the information from one node to the other, and it would avoid repeating the code for the regressor..

river/tree/mondrian/mondrian_tree_nodes.py
AlexandreChaussard: Hi ! Looking through the code right now, here is what I found out:
- parent: parent of the node can change
- time: splitting time of the node can change
- is_leaf: it does change for all node that is splitting
- depth: it is updated

river/utils/math.py
AlexandreChaussard: Solved in latest commit.

river/utils/math.py
AlexandreChaussard: Solved in latest commit.

AlexandreChaussard:
Hey 👋 

I've fixed/answered all the previous reviews, and I've added support for the random state !

smastelini (changes requested):
Here is one more wave of comments. Once they are fixed, I'll create a PR for your fork with code-style suggestions. Keep up the excellent work!

river/tree/mondrian/mondrian_tree_nodes.py
smastelini: Actually, the clone method creates a fresh copy of the object. If you set `include_parameters=True` then the property values will also be copied. Let's discuss this design decision once I got the gist of the code as a whole.

river/tree/mondrian/mondrian_tree_classifier.py
```diff
@@ -0,0 +1,462 @@
...
+
+from random import uniform
+from random import choices
+
+from river.tree.mondrian.mondrian_tree import MondrianTree
+
+from river.tree.mondrian.mondrian_tree_nodes import MondrianTreeBranchClassifier
+from river.tree.mondrian.mondrian_tree_nodes import MondrianTreeLeafClassifier
+
+from river.base import typing
```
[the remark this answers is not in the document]
MaxHalford: Yeah I would prefer from river import base and then base.typing.whatever 

(Sorry for poor formatting I'm on my phone)

AlexandreChaussard:
Hey guys 👋 

I've been reviewing all comments again @smastelini, feel free to let me know about the modifications 😄 

I've also been trying to fix the MyPy highlighted issues, but I'm struggling with a last one as you can see. If you guys could have a look that'd be awesome, coz I'm really out of ideas here... (maybe @MaxHalford you'd have some inputs too) I introduced a lot of unnecessary functions to solve the problem in my opinion. This test just goes against the AbstractClass concept, I have no idea how to satisfy the condition for it to be ok 😢 

Cheers 🎉

smastelini:
Don't worry @AlexandreChaussard. I saw in my local setup there were errors but I wanted to wait until you updated the code after checking the existing comments. 

My next step will be creating a PR to your fork with style suggestions and possibly fixing the mypy errors. You can take these suggestions and take from there. 

Does it sounds good?

AlexandreChaussard:
Actually I'm proposing a new version here that is accepted by MyPy and runs properly too.

I'm kinda disappointed though, the way I made it work is by detyping everything... I know it's not really "python-like" to put the classes to each and every object we manipulate through the code, but it pains me to see that it fixed it ngl 😢 

Apparently the code quality is still not matched, though I don't see any error in the logs when using `pre-commit` on my end ? Maybe I'm missing something here @smastelini 

Anyway, I think we can go for another round of comments on the code or forking on your side, as you which @smastelini , I think I went through everything?

smastelini:
Yeap. Let me check that until Monday.

smastelini:
Hey @AlexandreChaussard, the next step is checking the #TODO comments I have left in the code. In parallel, I'll check the AMF file :)

river/ensemble/aggregated_mondrian_forest.py
```diff
@@ -220,11 +191,7 @@ def learn_one(self, x, y):
         """
 
         # Updating the previously seen classes with the new sample
-        if y not in self._classes:
-            self._classes[y] = len(self._classes)
-
-        # Checking the features consistency
-        self._check_features_consistency(x)
+        self._classes.add(y)
```
smastelini, line 194: Perfect!

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -164,16 +168,16 @@ def range_extension(self, x_t, extensions):
         """
 
         extensions_sum = 0.0
-        for j in range(self.n_features):
-            x_tj = x_t[j]
-            feature_min_j, feature_max_j = self.range(j)
+        for feature in x_t.keys():
```
smastelini, line 171: same comment as before regarding `keys()`

river/tree/mondrian/mondrian_tree_nodes.py
```diff
@@ -348,18 +372,18 @@ def update_downwards(
         # Updating the range of the feature values known by the node
         # If it is the first sample, we copy the features vector into the min and max range
         if self.n_samples == 0:
-            for j in range(self.n_features):
-                x_tj = x_t[j]
-                self.memory_range_min[j] = x_tj
-                self.memory_range_max[j] = x_tj
+            for feature in x_t.keys():
```
smastelini, line 375: Same here

smastelini (changes requested):

river/tree/mondrian/mondrian_tree_classifier.py
```diff
@@ -0,0 +1,508 @@
...
+    Examples
+    --------
+    >>> from river import tree
+    >>> from river import evaluate
+    >>> from river import metrics
+    >>> from river.datasets import Bananas
+
+    >>> dataset = Bananas().take(500)
+
+    >>> model = tree.MondrianTreeClassifier(n_classes=2, step=0.1, use_aggregation=True, dirichlet=0.2, seed=1)
```
smastelini, line 52:
```suggestion
    >>> model = tree.mondrian.MondrianTreeClassifier(
    ...     n_classes=2,
    ...     step=0.1,
    ...     use_aggregation=True,
    ...     dirichlet=0.2,
    ...     seed=1
    ... )
```

AlexandreChaussard:
Hey (again) @MaxHalford !
@smastelini and I have worked hard on the classifiers, and we think it's ready to go!

Any chance you could have a look as well so maybe we can go for the merge :3 ?

Cheers 🥳

smastelini (approved):
Thank you so much, @AlexandreChaussard, for all the work on this PR!

You have addressed all the main problems identified in the code and paved the way for the upcoming AMF regressor. 

It remains to decide whether or not we can get rid of the `n_classes` parameter and track incoming classes automatically. We will wait for the answer from AMF's original authors and proceed to make changes if needed afterward.

PR #1133, opened 2022-12-31
Use parallelization to speed up pytest

boragokbakan:
As discussed with @MaxHalford on [Discord](https://discord.com/channels/1049726457515102238/1051186195775893564), we can use the `pytest-xdist` plugin to distribute unit tests over multiple CPU cores.

xdist


.github/workflows/branch-tests.yml
```diff
@@ -47,4 +48,4 @@ jobs:
         run: python -c "from river import datasets; datasets.CreditCard().download(); datasets.Elec2().download()"
 
       - name: pytest
-        run: pytest
+        run: pytest -n logical # Run pytest on all logical CPU cores
```
MaxHalford, line 51: I'm assuming `-n logical` provides the fastest results?

MaxHalford (approved):

.github/workflows/branch-tests.yml
boragokbakan: It should — yes. I just did not want to hard code this in case the runner specs changed in the future.

In the existing GitHub runners, `-n auto` and `-n logical` should give the same result, though.

PR #1134, opened 2022-12-31
venv caching for faster CI

boragokbakan:
As discussed with @MaxHalford on [Discord](https://discord.com/channels/1049726457515102238/1051186195775893564), caching the virtual environment allows us to bypass completely `pip install`s. The existing approach caches downloaded pip packages, although `pip install` takes considerable amounts of time even then.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -49,6 +49,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m venv ~/.venv
           source ~/.venv/bin/activate
+          pip install wheel
           pip install scikit-learn sqlalchemy
           pip install pytest-xdist[psutil]
 
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - feature/dep_cache
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - feature/dep_cache
```
MaxHalford, line 7: Don't forget to remove this once you're done.

MaxHalford (commented):
Really cool! Can you add the same to the `branch-docs`, `dev-docs`, and `branch-tests` jobs?

.github/workflows/unit-tests.yml
boragokbakan: > Really cool! Can you add the same to the `branch-docs`, `dev-docs`, and `branch-tests` jobs?
Now that I think about adding the same step, these jobs will use the same cache any way, and if a cache doesn't exist, they will each create their own environments and overwrite each other's.

So, how would you feel about simplifying these jobs, by moving the environment creation to a separate job, on which `unit-tests`, `branch-docs`, `dev-docs`, and `branch-tests`? This should be possible with GitHub actions ([blog post](https://github.blog/2021-11-18-7-advanced-workflow-automation-features-with-github-actions/#2-build-breakpoints-in-a-workflow-with-dependencies-or-create-dependencies-between-workflows)).

This could make the jobs easier to maintain (only one place to update) and downstream jobs very fast (although the end-to-end duration would remain the same).

In fact, we could extend this even to the river build step, so only one job would call `pip install -e ".[dev]"`.

MaxHalford: I agree that GitHub Actions is the way to go, I'm glad you mention it :)

MaxHalford:
FYI @boragokbakan I just used Actions [here](https://github.com/MaxHalford/myriade), hope that helps :)

boragokbakan:
Hi @MaxHalford I've now checked it out, thanks! :)

I played a little with artifacts, but it's slow for some reason ([issue](https://github.com/actions/upload-artifact/issues/199)) so I'm now trying again with caching the river directory.

I'll try to update the PR in a few days!

MaxHalford:
Sounds good! Take your time :)

PR #1142, opened 2023-01-06
Venv Caching and Dependent Jobs

boragokbakan:
As discussed with @MaxHalford, this PR contributes:

- venv caching instead of pip caching
- sequential jobs (e.g., both unit-tests and docs use the same river build)
- removed redundancy (e.g., branch-tests and unit-tests are very similar) through _reusable workflows_

This new CI workflow will:
- build river and install its dependencies (only for Python 3.9 on Ubuntu if it's a PR, see the screenshot below)
- build docs (and deploy if it's a push to the `main` branch)

The behaviour of the CI should remain unaffected. Hopefully, this PR should make it easier to create flexible workflows and it should lend itself well to non-linear setups.

image


boragokbakan:
Just to note, it should be possible to get rid of skipped checks by manipulating the matrix in the calling workflow rather than skipping jobs based on conditionals in the called workflow.

MaxHalford:
I don't know what say @boragokbakan, this is so good! I'm so grateful for you to have contributed this :). You went the extra mile and significantly improved our CI setup.

> Just to note, it should be possible to get rid of skipped checks by manipulating the matrix in the calling workflow rather than skipping jobs based on conditionals in the called workflow.

I see what you note. I'm not sure what is best though. It looks ok right now, in that showing what jobs were skipped is informative.

PR #1145, opened 2023-01-07
Docs fix

boragokbakan:
This PR fixes git issues during dev-docs flows, encountered after #1142. This PR removes the `.git` directory from the saved river build cache, to avoid conflicts.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,8 +4,6 @@ on:
   workflow_call:
 
 jobs:
-
-
   docs:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -39,4 +38,4 @@ jobs:
           git checkout gh-pages
           git pull
           git checkout main
-          mike deploy dev --push --remote https://github.com/online-ml/river.git
+          mike deploy dev --push --remote https://github.com/${{ github.repository }}.git
```
boragokbakan, line 41: The static repo URL always caused builds to fail in my own river fork (i.e., 403 / no authorization), which prevented me from realising when my main builds failed for another reason. So the error went unnoticed when I created the previous PR. :)

MaxHalford: Nice catch! This probably explains this: https://discord.com/channels/1049726457515102238/1051353581573374103/1061339713228767314

PR #1146, opened 2023-01-07
Adding xStream Model

isci1102:
We are currently working to implement PySAD into River, for the "polytechnique-project".
It is a second pull request to add the model "xstream" from PySAD.

Alexandra


river/anomaly/base.py
```diff
@@ -45,6 +48,63 @@ def score_one(self, x: dict) -> float:
 
         """
 
+    def fit(self, X, y=None):
```
MaxHalford, line 51: If I understand correctly, you're using the `_HSChains` internally in the `xStream` class. The `_HSChains` is a batch method, so you need to call `fit`. Even if that's the case, it's not a good idea to add a `fit` method to the `AnomalyDetector` base class.

river/anomaly/xstream.py
```diff
@@ -0,0 +1,251 @@
+import math
+import random
+from collections import Counter
+
+from river import anomaly
+from river.transform.projection.streamhash_projector import StreamhashProjector
+from river.utils.math import dict_zeros, get_minmax, merge
+
+random.seed(14)
```
MaxHalford, line 9: You'll want to seed the random generator in the class's `__init__` method.

river/anomaly/xstream.py
```diff
@@ -0,0 +1,251 @@
...
+        y
+            Ignored since the model is unsupervised (Default=None).
+
+        """
+        self.step += 1
+
+        X = self.streamhash.learn_one(x)
+        X = self.streamhash.transform_one(X)
+        self.cur_window.append(X)
+        self.hs_chains.fit(X)
```
MaxHalford, line 59: Is there no way to train `hs_chains` incrementally? Does it have `fit` on all of `X` each time a new sample is seen?

MaxHalford (commented):
I added a few comments to discuss how "batchy" xStream is.

Also, can you provide some experimental results? Did this method produce good results for you?

river/anomaly/base.py
isci1102: The results have been added in the file benchmarks/xStream_benchmark and random seed has been removed from the class.

MaxHalford:
Closing due to lack of activity.

PR #1148, opened 2023-01-08
Introduce base.DriftAndWarningDetector

MaxHalford:
Some drift detectors have a `warning_detected` property, some don't. I believe introducing a new base class clarifies this difference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,9 @@ venv.bak/
 .dmypy.json
 dmypy.json
 
+# ruff
+.ruff_cache
+
 # Pyre type checker
 .pyre/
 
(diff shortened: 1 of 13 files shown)


smastelini:
That might be helpful indeed. But keep in my mind that the input type is as important as the two level detection capabilities. I still didn't figure out a practical way to organize them in classes.

smastelini:
Some detectors support continual entries, other only binary inputs. Most of the binary ones have warning levels.

smastelini (approved):

MaxHalford:
Ah good catch @smastelini, I'll have a think about it then :)

MaxHalford:
Ok I'm going to merge this to help with #1149, but I'll keep your comment in mind. You're right, we should distinguish between continuous and discrete drift detectors.

PR #1149, opened 2023-01-09
add reset option to DriftRetrainingClassifier

danielnowakassis:


MaxHalford:
Hey hey. If I understand correctly, this is to support drift detectors which don't have a `warning_detected` property, correct? In that case, I would rather have a single class which supports both cases, rather than two classes. I have a feeling that will be less confusing for our users.

MaxHalford:
I just merged #1148, which should help here. For instance, it allows calling `isinstance`

danielnowakassis:
Actually, the main intention was to build a class that resets in drifts. I will put a parameter should_reset if the user wants to reset in case a drift is detected :)

river/drift/retrain.py
```diff
@@ -39,10 +41,22 @@ class DriftRetrainingClassifier(base.Wrapper, base.Classifier):
 
     """
 
-    def __init__(self, model: base.Classifier, drift_detector: base.DriftAndWarningDetector = None):
+    def __init__(
+        self,
+        model: base.Classifier,
+        drift_detector: base.DriftAndWarningDetector = None,
+        use_bkg: bool = True,
```
MaxHalford, line 48: We try to avoid acronyms, it can be cryptic. I would rather be obvious and call this parameter `train_in_background`

river/drift/retrain.py
```diff
@@ -39,10 +41,22 @@ class DriftRetrainingClassifier(base.Wrapper, base.Classifier):
...
         self.model = model
         self.bkg_model = model.clone()
+        self.use_bkg = use_bkg
         self.drift_detector = drift_detector if drift_detector is not None else drift.DDM()
+        if self.use_bkg:
+            self.bkg_model = model.clone()
+            if not isinstance(self.drift_detector, base.DriftAndWarningDetector):
+                raise Exception(
+                    "Drift detection method must have a warning tracker if parameter use_bkg is True"
+                )
```
MaxHalford, line 60: No need to check this. When the `warning_detected` property is checked in `learn_one`, an exception will be raised if necessary.

MaxHalford (changes requested):

river/drift/retrain.py
```diff
@@ -39,10 +41,18 @@ class DriftRetrainingClassifier(base.Wrapper, base.Classifier):
...
+        model: base.Classifier,
+        drift_detector: base.DriftAndWarningDetector = None,
+        train_in_background: bool = True,
+    ):
         self.model = model
         self.bkg_model = model.clone()
+        self.train_in_background = train_in_background
         self.drift_detector = drift_detector if drift_detector is not None else drift.DDM()
+        if self.train_in_background:
+            self.bkg_model = model.clone()
```
MaxHalford, line 55: No need for this right? You already do this right above (at line 51)

MaxHalford (changes requested):

river/drift/retrain.py
danielnowakassis: Actually a bkg_model must be created only if  self.train_in_background

MaxHalford: Right!

MaxHalford:
Good job.

PR #1159, opened 2023-01-17
Tidy up some tree stuff

MaxHalford:
Hey @smastelini. What do you think about moving the adaptive tree ensembles to the `tree` module? I think it makes more sense, but you tell me. I also made them pass the generic test suite :)


smastelini:
Hi @MaxHalford, I don't have a strong preference, but Forests are ensembles. Do you think users would be confused by this change? For example, in sklearn Random Forest belongs to the ensemble module. As a user coming from sklearn I would look directly into ensemble to find Random Forest models.

MaxHalford:
Indeed, sklearn does it that way. But I find it clearer to have all trees in one place. For me, the ensemble module should just contain generic wrappers which take as first input parameter a model.

MaxHalford:
Or we could have a `forest` module to be honest!

smastelini:
I see your point. I guess we can try it, and if we get too many questions, we decide if we go back to the previous state. I think there are too many ensembles in the module. Some of them are somewhat redundant. In the future, we can trim up that even further. For instance, everything `ADWIN*Classifier` or similar could be changed to accept any drift detector.

smastelini:
Oh, I like forests! I like it very much!!!

MaxHalford:
Sounds good, let me make the modification :)

PR #1162, opened 2023-01-20
Active learning beginning

MaxHalford:
- This PR introduces a new `active` module, for online active learning
- I implemented a first heuristic strategy based on entropy. It seems to work reasonably, although I'm sure there are better strategies. We can work on that later.
- The goal is really to provide an architecture and a base class to clarify how we envision active learning.
- I've also written a recipe to show how this all works.
- I've made it so that `progressive_val_score` works with active learning models. I've also slightly edited `progressive_val_score` to make it yield the result once the last sample has been processed.
- I cleaned up quite a few things too in the PR, sorry if I got carried away.


smastelini:
Good stuff!!

smastelini:
Good stuff!!

PR #1163, opened 2023-01-21
Show multioutput models some love

smastelini:


smastelini:
The failing test is related to the `active` model. A dataset needs to be downloaded during the test. We should cache it :D

We can fix that in another PR.

PR #1164, opened 2023-01-22
Handle probabilities in MultiClassEncoder

MaxHalford:
Hey @smastelini! I believe there is a little mistake in `MultiClassEncoder`. Basically, it seems to me that the predicted outputs are always 1 or 0 for each `False/True` of each output. That's because a probability is assigned to `True` or `False`, which is then converted to `1` or `0` with `utils.norm.normalize_values_in_dict`. This is difficult to notice because the metrics we use don't care about the probabilties. I think the implementation I suggest is correct, but I'll let you double check.

Anyway, good work! It's a nice model to have.

### DIFF
--- a/river/multioutput/encoder.py
+++ b/river/multioutput/encoder.py
@@ -51,7 +51,7 @@ class MultiClassEncoder(base.MultiLabelClassifier):
         super().__init__()
         self.model = model
 
-        self._label_map: typing.DefaultDict[tuple, int] = {}
+        self._label_map: dict[tuple, int] = {}
         self._r_label_map: dict[int, tuple] = {}
         self._labels: set[typing.Hashable] = set()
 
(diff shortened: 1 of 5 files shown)


river/multioutput/encoder.py
```diff
@@ -1,35 +1,37 @@
+from __future__ import annotations
```
MaxHalford, line 1: Note: when you add this to the top of the file, the `pre-commit` logic will automatically convert your type annotations (thanks @kulbachcedric!)

MaxHalford:
I'm also fixing the issue with `SMSSpam` not being available.

smastelini:
Good catch, @MaxHalford! It escaped me. The fix looks alright! 😃

smastelini (approved):

PR #1165, opened 2023-01-22
Update basic-concepts.md

Zenith1618:
Since I believe offline models are more susceptible to concept drift, the more accurate statement would be that


MaxHalford:
Cheers! Good catch.

Zenith1618:
Thanks

PR #1170, opened 2023-01-27
Dynamic classes for AMFClassifier

AlexandreChaussard:
Hi again @MaxHalford and @smastelini 👋 !

I've implemented J. Mourtada's suggestion to handle a dynamic number of classes during computation for the AMFClassifier.

Note a decrease from 84.77% to 82.16% on the Bananas dataset (2.61% drop). Maybe I didn't grasp the trick correctly, but I'm not sure what could be wrong really. Most likely imo, the drop can be seen as the effect of not computing the loss with the Jeffrey's prior with the right number of classes up until the dynamic number of classes finally meets the real one.

Do not hesitate to have a look so we can decide on whether to keep it, and possibly find what could be wrong!

Cheers 🎈

### DIFF
--- a/river/forest/aggregated_mondrian_forest.py
+++ b/river/forest/aggregated_mondrian_forest.py
@@ -94,7 +94,7 @@ class AMFClassifier(AMFLearner, base.Classifier):
     ----------
     n_classes
         Number of expected classes in the labels. This is required since we
-        don't know the number of classes in advance in a online setting.
+        don't know the number of classes in advance in an online setting.
     n_estimators
         The number of trees in the forest.
     step
--- a/river/tree/mondrian/mondrian_tree_classifier.py
+++ b/river/tree/mondrian/mondrian_tree_classifier.py
@@ -18,7 +18,7 @@ class MondrianTreeClassifier(MondrianTree, base.Classifier):
     Parameters
     ----------
     n_classes
-        Number of classes of the problem.
+        Number of expected classes of the problem.
     step
         Step of the tree.
     use_aggregation
@@ -124,7 +124,7 @@ class MondrianTreeClassifier(MondrianTree, base.Classifier):
             Node to evaluate the score.
         """
 
-        return node.score(self._y, self.dirichlet, self.n_classes)
+        return node.score(self._y, self.dirichlet, len(self._classes), self.n_classes)
 
     def _predict(self, node: MondrianNodeClassifier) -> dict[base.typing.ClfTarget, float]:
         """Compute the predictions scores of the node regarding all the classes scores.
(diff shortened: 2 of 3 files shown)


MaxHalford:
Hey @AlexandreChaussard, thanks for the PR. I pulled the code and played around with it a bit. 

I have to say, I wonder if the `n_classes` (the expected number of classes) is really needed. Can't we just use `len(self.classes)`? That seems to produce good results. You could even do `max(2, len(self.classes))` to make sure the quantity is at minimum 2.

I would remove the `n_classes` parameter. Users can control the Jeffrey prior via the `dirichlet` parameter. I know the `dirichlet` has to be set. From what I can tell from the code, the `dirichlet` is basically acting as the expected proportion of each class. So it's basically `dirichlet = 1 / n_classes`. I would implement that in the `score` method, which seems to me to be the only place you need `dirichlet`. I just did this in the `score` method:

```py
dirichlet = 1.0 / n_classes
return (count + dirichlet) / (self.n_samples + dirichlet * n_classes)
```

and got `Accuracy: 84.97%` in the doctest.

Let me know what you think!

AlexandreChaussard:
I guess we _can_ remove the `n_classes` looking at your experimental results, yet J. Mourtada seemed to be saying that it wasn't so easy to do on the theoretical side (hence his trick with the estimated number of classes). Also, one might highlight that the `max(2, len(n_classes))` has yet another behavior as the solutions proposed results so far: initial idea was using the real number of classes, Mourtada's trick introduces a rough estimation of this number for the **first sample** of a new class, while your proposition gives a constant value **for `N` first entries** that belongs to the same class before seeing another class if we are in a 2 classes situation (am I making sense?).

Dor the dirichlet, if it seems to provide good results by always putting it at `1/n_classes`, then why not. Personally I found it more interesting for the Bananas dataset around `0.2` with 2 classes, though maybe I'm wrong. Overall it feels nice to control the smoothing between regions with it imo.

All the suggested changes are just a matter of minutes to be changed really, though personally I'm struggling with the idea of exchanging the paper's ideas with experimental ones as it kinda feels fuzzy to me (maybe I don't master trees enough to grasp the whole thing too though, this may just be all quite equivalent that I wouldn't see it haha). 

I'll just follow your opinion @MaxHalford and @smastelini 's for that one if you don't mind 😅 it seems quite out of my league.

MaxHalford:
So I wouldn't overthink it. I agree that it's ideal to follow the paper with high fidelity. However, I think here we have some freedom. 

We simply have to think of the score as a [Bayesian average](https://www.wikiwand.com/en/Bayesian_average). We want to have a parameter which controls how much smoothing is applied.

Doing `dirichlet=0.01` when there's more than 2 classes is just a hack. We can do `dirichlet = 1 / len(classes)` and add a smoothing parameter. I can take care of this if you want.

AlexandreChaussard:
If you could take care of it that'd be great honestly, I'm not too comfortable with the approach so I'd rather let you handle it if you don't mind @MaxHalford 😅

Seems like very few changes to be made, though I'm not sure of the execution (thanks for the Bayesian average reference though, it makes things a little clearer)

MaxHalford:
Ok done in #1172. I'm going to release a new version of River now. We can improve on this in the future! Who knows, there might be some common things to put in common between Hoeffding and Mondrian trees.

Issue #1178, opened 2023-02-06
Inconsistent initialisation of quantile estimates


Tveten:
## Versions
**river version**: 0.13.0
**Python version**: 3.9.16
**Operating system**: Windows 10 Pro

## Describe the bug
During initialisation of river.Quantile (the first five observations input to .update()), the values returned by Quantile.get() behave inconsistently with respect to the set quantile q. As you can see in the example below, the low quantile stays high despite observing lower values, while the high quantile drops. Exected behaviour: q.get() returns values such that Quantile(0.01) <= Quantile(0.99) for all sample sizes. Something like the regular sample quantile for sample sizes <= 5.

## Steps/code to reproduce
```python
from river.stats import Quantile

# Low quantile
q = Quantile(0.01)
for x in [5, 0, 0, 0, 0, 0, 0, 0]:
    q.update(x)
    print(q.get())
# 5.0
# 5.0
# 5.0
# 5.0
# 5.0
# 0.0

# High quantile:
q = Quantile(0.99)
for x in [5, 0, 0, 0, 0, 0, 0, 0]:
    q.update(x)
    print(q.get())
# 5.0
# 0.0
# 0.0
# 0.0
# 0.0
# 0.0
# 0.27777777777777773

```

AdilZouitine:
First investigation, the inconsistent behavior was in pure python implementation too.
I'll try to fix that.

AdilZouitine:
I made a fix in the following PR: https://github.com/online-ml/river/pull/1179

Thanks @Tveten, for the report 😄

Tveten:
Fantastic, thanks for the quick fix!

MaxHalford:
Added unit tests [here](https://github.com/online-ml/river/blob/main/river/stats/test_quantile.py) :)

PR #1179, opened 2023-02-06
Bump watermill version

AdilZouitine:
Hi, I fix this issue: https://github.com/online-ml/river/issues/1178.

The bug was in the rust part; I sorted the `height` list after the sixth update, so the initialization wasn't consistent.

The fix here : https://github.com/online-ml/watermill.rs/commit/bb284c4c87dd77cd31a6d39db4893a0ea04cdce1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,6 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.16.5", features = ["extension-module"] }
-watermill = "0.1.0"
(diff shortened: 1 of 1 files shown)


MaxHalford:
Good job! Can you fix the tests? And also add an entry `unreleased.md`? There's no rush.

AdilZouitine:
The tests are good now. 
I have an issue with black because it doesn't support target version `3.11`.

MaxHalford:
Cheers, I'll take care of black.

PR #1189, opened 2023-02-22
Efficient Prequential AUC-ROC

davidlpgomes:
A C++ implementation of the Prequential AUC-ROC, which utilizes a sliding window with size S, calculating the AUC-ROC with the last S seen instances. It uses Cython to compile the C++ code.

Based on [Brzezinski and Stefanowski, 2017](https://link.springer.com/article/10.1007/s10115-017-1022-8).


river/metrics/preq_roc_auc.py
```diff
@@ -0,0 +1,67 @@
+from river import metrics, utils
+
+from .efficient_preqrocauc import EfficientPreqROCAUC
+
+__all__ = ["PreqROCAUC"]
+
+
+class PreqROCAUC(metrics.base.BinaryMetric):
```
MaxHalford, line 8: Could we call this `RollingROCAUC` instead? The rolling terminology is aligned with the rest of River.

river/metrics/__init__.py
```diff
@@ -88,6 +89,7 @@
     "MicroRecall",
     "MultiFBeta",
     "multioutput",
+    "efficient_preqrocauc",
```
MaxHalford, line 92: I would not expose this.

river/metrics/preq_roc_auc.py
```diff
@@ -0,0 +1,67 @@
...
+    """Prequential version of the Receiving Operating Characteristic Area Under the Curve.
+
+    The Prequential ROCAUC (predictive sequential) calculates the metric using the instances
+    in its window of size S. It keeps a queue of the instances, when a instance is added
+    and the queue length is equal or bigger than S, the last instance is removed. The metric
+    has a tree with ordered instances, in order to calculate the metric efficiently. It was
+    implemented based on the algorithm presented in Brzezinski and Stefanowski, 2017.
+
```
MaxHalford, line 16: Could you add a little paragraph to explain that this is an exact implementation, as opposed to the existing `ROCAUC` which is approximate? Users need to understand this trade-off.

MaxHalford (changes requested):
This is really good! Would it be possible to add some unit tests to `test_metrics.py`? For instance, we have a `test_rolling_metric` function to test the outputs of rolling metrics are exactly the same as what would have been obtained with scikit-learn.

davidlpgomes:
Thanks for the feedback. I will work on those issues 😃

river/metrics/__init__.py
davidlpgomes: Reference removed.

river/metrics/preq_roc_auc.py
davidlpgomes: Renamed.

river/metrics/preq_roc_auc.py
davidlpgomes: Added a paragraph explaining the difference to the standard ROCAUC.

river/metrics/rolling_roc_auc.py
```diff
@@ -0,0 +1,82 @@
...
+    beginning of the stream, while the RollingROCAUC calculates the exact value
+    considering only the last S instances. This approach may be beneficial if
+    it's necessary to evaluate the model's performance over time, since
+    calculating the metric using the entire stream may hide the current
+    performance of the classifier.
+
+    Parameters
+    ----------
+    window_size
+        The max length of the window
```
MaxHalford, line 29:
```suggestion
        The max length of the window.
```

river/metrics/rolling_roc_auc.py
```diff
@@ -0,0 +1,82 @@
...
+    pos_val
+        Value to treat as "positive"
```
MaxHalford, line 31:
```suggestion
        Value to treat as "positive".
```

river/metrics/test_metrics.py
```diff
@@ -13,6 +13,8 @@
 
 from river import metrics, utils
 
+from .test_wrappers import roc_auc_score as wrapper_roc_auc_score
```
MaxHalford, line 16: I would place the `wrapper_roc_auc_score` function in the `test_metrics.py` file, since it's only used there. Fine with you?

MaxHalford (changes requested):
Looking good!!

Could you add an entry to `unreleased.md`? That way we keep track of what changes were made from one release to the other.

river/metrics/test_metrics.py
davidlpgomes: Sure, no problem ;)

davidlpgomes:
@MaxHalford I added the RollingROCAUC entry to `unreleased.md`  and fixed the issues.

MaxHalford:
Ok, looks good to me ✅

Thank you very much for taking into account my feedback, and congratulations on your work! I love it.

davidlpgomes:
Sure @MaxHalford, thanks for the help! Congratulations on River, it's great! 😄

PR #1208, opened 2023-03-22
Add DriftRetrainRegressor

danielnowakassis:


river/drift/retrain.py
```diff
@@ -99,3 +101,77 @@ def _unit_test_params(cls):
...
+
+    This regressor is a wrapper for any regressor. It monitors the incoming data for concept
+    drifts in the model's predictions. In case a drift is detected, the model will be reset.
+
+    Parameters
+    ----------
+    model
+        The classifier and background classifier class.
+    drift_detector
+        Algorithm to track concept drifts.
```
MaxHalford, line 117: Can you document what algorithm is used by default?

river/drift/retrain.py
```diff
@@ -99,3 +101,77 @@ def _unit_test_params(cls):
...
+        self._var.update(drift_input)
+        if self._var.mean.n == 1:
+            self.drift_detector.update(0.5)
+        else:
+            sd = sqrt(self._var.get())
+            # We assume the error follows a normal distribution -> (empirical rule)
+            # 99.73% of the values lie  between [mean - 3*sd, mean + 3*sd]. We
+            # assume this range for the normalized data. Hence, we can apply the
+            # min-max norm to cope with  ADWIN's requirements
+            self.drift_detector.update((drift_input + 3 * sd) / (6 * sd) if sd > 0 else 0.5)
```
MaxHalford, line 172: Should these constants be passed as parameters to the class?

Also, do you have any paper/reference to back up this heuristic? It seems reasonable, but I can imagine other ways of doing this. For instance, using quantiles.

river/drift/retrain.py
```diff
@@ -99,3 +101,77 @@ def _unit_test_params(cls):
...
+        y_pred = self.model.predict_one(x)
+        if y_pred is None:
+            return
+
+        drift_input = y - y_pred
+        self._var.update(drift_input)
+        if self._var.mean.n == 1:
+            self.drift_detector.update(0.5)
+        else:
+            sd = sqrt(self._var.get())
```
MaxHalford, line 167:
```suggestion
            sd = math.sqrt(self._var.get())
```

MaxHalford (changes requested):

river/drift/retrain.py
```diff
@@ -99,3 +101,77 @@ def _unit_test_params(cls):
             "drift_detector": drift.binary.DDM(),
         }
         yield {"model": naive_bayes.GaussianNB(), "drift_detector": drift.binary.DDM()}
+
+
+class DriftRetrainingRegressor(base.Wrapper, base.Regressor):
+    """Drift retraining regressor.
+
+    This regressor is a wrapper for any regressor. It monitors the incoming data for concept
+    drifts in the model's predictions. In case a drift is detected, the model will be reset.
```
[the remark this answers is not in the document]
smastelini: Hmm, that is a good point. We can turn it into a utility class or function, but I'm not sure it will be used in many places.

A bit of history:

- The original ADWIN paper states that the algorithm works with data in the `[0, 1]` range. Anything different than that should be normalized to this range. That is why I introduced the empirical normalization rule when I coded part of ARF in skmultiflow.
- According to the authors (I discovered that much later), empirically speaking, ADWIN also works with data outside the `[0,1]` range. So, idk XD; better safe than sorry?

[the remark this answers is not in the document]
MaxHalford: Yes keep it, I'm sure it doesn't affect performance that much.

MaxHalford:
> FAILED river/test_estimators.py::test_check_estimator[DriftRetrainingRegressor(BayesianLinearRegression):check_wrapper_accepts_kwargs] - AssertionError

@danielnowakassis check the [unit test](https://github.com/online-ml/river/blob/50bc0b70323cf2b385b20d9e6aa8c9dbfff594ac/river/checks/common.py#L223). You have to add `**kwargs` support to your methods.

Issue #1211, opened 2023-03-28
DBSTREAM: recluster whenever needed


bmatschke:
Currently in river.cluster.DBSTREAM, `predict_one()` calls the offline phase `_recluster()` no matter whether it is necessary or not. At the same time, the attributes `clusters`, `n_clusters`, `centers` are "public" and do not call `_recluster()` even if needed. As if it currently is expected to call `predict_one()` before `clusters`, `n_clusters`, `centers` are updated.

It seems more natural to do the offline phase precisely whenever it is needed.

To achieve this I propose to 
- add an attribute `_clustering_is_up_to_date`, which is set to True at the end of each `_recluster()` call, set to False whenever `learn_one()` is called, and that `_recluster()` returns early in case `_clustering_is_up_to_date` is already True, 
- make `clusters`, `n_clusters`, `centers` into "private" attributes `_clusters`, `_n_clusters`, `_centers`,
- add properties `clusters`, `n_clusters`, `centers` to DBSTREAM that first call `_recluster()` before returning their corresponding private attributes,
- include an underscore in front of all other attributes in DBSTREAM to signify them as private.

I can provide a PR.

bmatschke:
The proposed code changes: https://github.com/bmatschke/river/tree/dbstream_recluster_whenever_needed

MaxHalford:
I believe this is now implemented.

PR #1213, opened 2023-03-28
in DBSTREAM: recluster when needed

bmatschke:
See issue https://github.com/online-ml/river/issues/1211.


river/cluster/dbstream.py
```diff
@@ -77,15 +77,15 @@ class DBSTREAM(base.Clusterer):
...
         micro clusters, or macro clusters that are generated by merging all $\alpha$-connected
         micro clusters. This set is generated through the offline phase of the algorithm.
-    centers
+    _centers
         Final clusters' centers.
-    micro_clusters
+    _micro_clusters
         Micro clusters generated by the algorithm. Instead of updating directly the new instance points
         into a nearest micro cluster, through each iteration, the weight and center will be modified
         so that the clusters are closer to the new points, using the Gaussian neighborhood function.
```
MaxHalford, line 91: You can still use the names without the `_` prefix.

river/cluster/dbstream.py
```diff
@@ -125,7 +125,7 @@ class DBSTREAM(base.Clusterer):
     >>> dbstream.predict_one({0: 5, 1: 2})
     1
 
-    >>> dbstream.n_clusters
+    >>> dbstream._n_clusters
```
MaxHalford, line 128: No, prefer without the `_`. We shouldn't expose private properties in docs.

river/cluster/dbstream.py
```diff
@@ -138,36 +138,38 @@ def __init__(
...
-        self.centers: typing.Dict = {}
-        self.micro_clusters: typing.Dict[int, "DBSTREAMMicroCluster"] = {}
+        self._n_clusters: int = 0
+        self._clusters: typing.Dict[int, "DBSTREAMMicroCluster"] = {}
+        self._centers: typing.Dict = {}
+        self._micro_clusters: typing.Dict[int, "DBSTREAMMicroCluster"] = {}
 
-        self.s: typing.Dict[int, typing.Dict[int, float]] = {}
-        self.s_t: typing.Dict[int, typing.Dict[int, float]] = {}
+        self._s: typing.Dict[int, typing.Dict[int, float]] = {}
```
MaxHalford, line 154: I don't think you need to make all these attributes "private". If they're not documented, people won't use them. Also, if we had to do this for every model, it would make the doc quite ugly. Can you revert this change for all the properties that don't require reclustering?

MaxHalford (changes requested):
Could you include a brief benchmark of before/after this PR? I expect there to be performance improvements.

bmatschke:
I just removed the previously added underscores for "private" variables. PEP 8 suggests them in general, however they suggest even stronger to keep a project in consistent style, which is the case here.

I can add a benchmark (only) by tomorrow. `predict_one()` indeed becomes more efficient if run several times without any `learn_one()` calls in between. However the more significant change is that the behaviour of `clusters`, `n_clusters`, and `centers` changed: reading these properties now checks first whether the offline step needs to be done, you have to decide whether you want this.

MaxHalford:
> I just removed the previously added underscores for "private" variables. PEP 8 suggests them in general, however they suggest even stronger to keep a project in consistent style, which is the case here.

Thanks for your understanding 🙏. Did you manage to see all my comments on the PR?

> I can add a benchmark (only) by tomorrow. predict_one() indeed becomes more efficient if run several times without any learn_one() calls in between. However the more significant change is that the behaviour of clusters, n_clusters, and centers changed: reading these properties now checks first whether the offline step needs to be done, you have to decide whether you want this.

Makes sense. Sorry if this is an annoying request. But I think it's interesting, in that we could re-use this lazy approach elsewhere in the `cluster` module.

river/cluster/dbstream.py
bmatschke: Thanks for catching this, this was by mistake. I'll resolve this shortly, in the next commit.

bmatschke:
About a simple benchmark: `_recluster` is at least quadratic in the number of clusters, whereas the remainder of `predict_one` is linear. E.g. for `n = 10000`, if cluster.DBSTREAM() with default parameters learns all `n` 1-dimensional features given by `data = [{'x': np.random.rand() * n} for i in range(n)]`, `_recluster` takes 0.94s, whereas the remainder of `predict_one` takes 0.10s. (Here one has around 4700 clusters.) So in case one calls `predict_one` many times, it makes sense to run the reclustering only once, instead of each time. So here, the running time improvement using this PR is by a factor of up to 9 (assuming a large number of prediction steps). This factor increases roughly linearly with `n` for the above `data`.

PR #1215, opened 2023-04-05
ClusterId out of scope

danielnowakassis:
ClusterId can be out of scope if "there is nothing to process"

### DIFF
--- a/river/cluster/textclust.py
+++ b/river/cluster/textclust.py
@@ -172,7 +172,7 @@ class TextClust(base.Clusterer):
         # realtime is only the current time non decoded to store for the plotter
         if self.realtime is not None:
             self.realtime = self.realtime
-
+        clusterId = None
         # if there is something to process
         if len(ngrams) > 0:
             # create artificial micro cluster with one observation


Dennis1989:
I think @MaxHalford and I were discussing this aspect, and as far as I remember, we agreed that it should be up to the user to catch this situation before learn_one is called. However, maybe my mind plays tricks on me...

MaxHalford:
Hey @danielnowakassis, can you provide more context to what this fixes? If I understand correctly, this handles the case where the `for` loop at line 156 is never entered. But why would that happen though? Because `x` is empty?

danielnowakassis:
I will give an example: 

```
from river import stream
from river import datasets
from river import compose
from river import feature_extraction
from river import metrics
from river import cluster

sms = datasets.SMSSpam()

metric = metrics.Rand()

model = compose.Pipeline(
     feature_extraction.BagOfWords(lowercase=True, ngram_range=(1, 2)),
     cluster.TextClust(real_time_fading=False, fading_factor=0.001, tgap=100, auto_r=True,
     radius=0.3)
   
)

for x, y in sms:
  text = x["body"].replace('\n', '')
  model.learn_one(text)
  pred = model.predict_one(text)
  metric = metric.update(y,pred)
  i+= 1
print(metric)
```

This error (`UnboundLocalError: local variable 'clusterId' referenced before assignment`) with ClusterId occurs both in learn_one being called before and after predict_one.

I'm not so sure why this happens.

MaxHalford:
Ok, thanks a lot for the code snippet. I'll merge this PR, because I assume it fixes your issue.

As I've already said, we (that includes me) need to spend more time in the cluster module.

PR #1218, opened 2023-04-11
Add of Online Hierarchical Clustering

kchardon:


MaxHalford:
Hey there! I hope it's ok for me to answer only by now.

Am I correct in assuming the algorithm stores all the data points it sees in memory (i.e. the `X` attribute)?

kchardon:
> Hey there! I hope it's ok for me to answer only by now.
> 
> Am I correct in assuming the algorithm stores all the data points it sees in memory (i.e. the `X` attribute)?

Hello, sorry for replying that late.
I use the `window_size` attribute and when there are more data than allowed, it deletes the oldest data point.
If `window_size` equals 0, then it stores all the data points

kchardon:
> > If window_size < 1, then it stores all the data points
> 
> Ok I see, fair. But I don't think we'll ever want that behavior. Could you remove it?

Yes I can. So I add an error if the value of `window_size` is not an integer > 0 ?

MaxHalford:
Nope, no need to check for an error. An exception will raise itself at some point. In general, we don't do input validation. Instead, we document well.

kchardon:
Okay ! I will delete it

kchardon:
If I delete the line, if the user put a window_size < 1, it will automatically store all the data points, because I do 

if len(self.X.keys()) >= self.window_size: delete oldest node and replace it by the new node
else : add the node as a new one

So since len(self.X.keys()) is always > 0, it will add all the nodes and use all the data points.

Do I replace the else with elif len(self.X.keys()) < self.window_size : add the node as a new one

With that it will raise an error if the window size is not > 0

hoanganhngo610:
Hi @kchardon! I am Hoang-Anh, the maintainer of the clustering module of River. For now, I think that it would be best if I can take over the review of this PR and proceed with it to get it merged to River as soon as possible, since from my first glance, the PR has been of really high quality.
If you wouldn't mind, first, I will start by refactoring the code and make changes to make it align with the rest from River. After all necessary changes are in place, I will get @MaxHalford to have a final look and give the green light.

hoanganhngo610:
@MaxHalford I think I'm quite happy with the current state of the code! Only one small concern is that the unit tests keep failing because there is an attribute error, saying that `HierarchicalClustering` does not have the attribute `distance_func`, although I believe atm there is no `distance_func` in my code anywhere anymore. Do you think of any potential reasons behind this?

hoanganhngo610:
@MaxHalford @kchardon If possible, I would really hope that both of you would be able to take a final look at the current state of the implementation and see if there are any changes you want to make. In case there is no opposition from both of you, when applicable, I would want to merge this into the main branch of `River`.

MaxHalford:
I will review next week :)

MaxHalford:
I (finally) took a look at this. There's a lot of nits to fix. But the main issue I see is that the code relies on numpy. It converts input dictionaries to numpy arrays. Would it be possible to use dictionaries only instead? I don't see any good reason to rely on numpy. Indeed, it's not in the spirit of River to rely on numpy when it can be avoided.

kchardon:
Hello, sorry to come back to this PR so late.
I've changed the code to not rely on numpy and fixed the issues during the checks.
Let me know if anything else can be improved!

MaxHalford:
Closing because the PR is messed up. But happy to pick back up the reviews if there's still some will!

PR #1241, opened 2023-05-24
Update year of release to 2023 for 0.15 and 0.16

khliland:
Only two integers in the documentation are touched.


MaxHalford:
Hehe good catch. So much for copy/pasting.

PR #1246, opened 2023-05-28
Type-check KNNR targets

Pablo-Davila:
`learn_one` only saves instances in KNN models. This means that without this type-check it is easy to face confusing exceptions when attempting to use the `predict_one` method.

I was very confused when I saw the following error when predicting after an initial training phase: `TypeError: can't convert type 'dict' to numerator/denominator`. It finally turned out that my target source array had shape `(n, 1)` instead of `(n,)` and that was messing with the aggregation function in prediction time.

### DIFF
--- a/river/neighbors/knn_regressor.py
+++ b/river/neighbors/knn_regressor.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 
 import functools
 import statistics
+from decimal import Decimal
+from fractions import Fraction
+from numbers import Integral
 
 from river import base, utils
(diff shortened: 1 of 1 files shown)


Pablo-Davila:
Hi Max. The point I was trying to make is that when training KNN models with badly-formatted data nothing happens because of the way these models work. Not until you try to predict afterwards, making it more difficult to spot the problem.

If the convention here is not to check inputs I will just close the PR. An assert was certainly a poor choice anyway.

PR #1258, opened 2023-06-05
Remove need for FunctionWrapper

MaxHalford:
What do you think @smastelini?


smastelini (approved):
LGTM! I will change some details in the tests of the multiouput module. Thank you so much for streamlining the code, @MaxHalford!

PR #1267, opened 2023-06-14
New parameters for Hoeffding Tree Classifiers

### LINKED ISSUE #1169: Splits happening on pure nodes.
## Versions

**river version**: 0.14.0
**Python version**: 3.10.8
**Operating system**: Windows 10 Pro

## Describe the bug

Hi,

I have started using River a couple of weeks ago, I have to say that I'm very satisfied with the tool and would like to encourage the community behind.

I have a misunderstanding of how Attribute Observers works, I have train a Hoeffding tree classifier (HTC) on a binary dataset. After processing a certain number of samples, I draw the current tree to see how it looks like. I am surprised to realize that some nodes are split although they contain a single class. 

## Steps/code to reproduce

```python
# Sample code to reproduce the problem
from river import datasets
from river import tree
from IPython.display import display

dataset = datasets.HTTP()

model = tree.HoeffdingTreeClassifier(grace_period=20)

display_after = 220000

for i, (x, y) in enumerate(dataset):
    model.learn_one(x, y)
    if i > 0 and not i % display_after:
        display(model.draw())
        break
```
Here is the tree you should obtain. Look at the split on the attribute `dst_bytes`, both left and right children are pure leaves with class label 0. It was also probably the case when the split at the root occurred. I would have expected the split to happen only when the node is not pure. Is this a normal behavior? Why?

image

smastelini:
Fixes #1169.

I added the `max_share_to_split` parameter, as discussed in the issue. Besides that, I also exposed another parameter that was inherited from MOA. This one prevents splitting when most of the data is concentrated in a single branch (regardless of the class distribution).

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -18,8 +18,8 @@
 
 ## neighbors
 
-- Add `neighbors.SWINN` to power-up approximate nearest neighbor search. SWINN uses graphs to speed up nearest neighbor search in large sliding windows of data.
-- Rename `neighbors.NearestNeighbors` to `neighbors.LazySearch`.
+- Added `neighbors.SWINN` to power-up approximate nearest neighbor search. SWINN uses graphs to speed up nearest neighbor search in large sliding windows of data.
+- Renamed `neighbors.NearestNeighbors` to `neighbors.LazySearch`.
 - Standardize and create base classes for generic nearest neighbor search utilities.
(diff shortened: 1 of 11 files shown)


smastelini:
In the near future, I might refactor the split criteria classes. They deserve some love :D

MaxHalford (approved):
Looks good to me!

I agree that the splitting "submodule" could be improved a little. It would be nice to have a bunch of unit tests for it.

smastelini:
Yeap, they are too Java-styled and can be streamlined for sure!

On a side note, the failing tests seem to be unrelated to the PR.

Issue #1269, opened 2023-06-22
Stateless inference for the last pipeline step


ColdTeapot273K:
## Versions

**river version**: current upstream
**Python version**: 3.10
**Operating system**: Fedora 38

## Describe the bug

Problem:

The current use of stateless inference context manager (https://github.com/online-ml/river/blob/main/river/compose/pipeline.py#L98) for a pipeline which consists only of unsupervised transformers (e.g. `StandardScaler`, `OneHotEncoder`) doesn't protect from calls to `learn_*` on the last transformer. 

Usecase:

Developing online models with other frameworks while leveraging handy online abstractions of `river` for pre-processing.

## Steps/code to reproduce

```python

import io
import logging

import river
from river import compose, datasets, linear_model, preprocessing, utils

model = compose.Select("ordinal_date", "gallup") | preprocessing.StandardScaler()


logger = logging.getLogger()
logger.setLevel(logging.DEBUG)


logs = io.StringIO()
sh = logging.StreamHandler(logs)
sh.setLevel(logging.DEBUG)
logger.addHandler(sh)


class_condition = lambda x: x.__class__.__name__ in ("StandardScaler", "LinearRegression")


with river.utils.log_method_calls(class_condition), river.compose.pure_inference_mode():
    for x, y in datasets.TrumpApproval().take(1):
        _ = model.transform_one(x)

print(logs.getvalue())
# desired output:
# StandardScaler.transform_one

# actual output:
# StandardScaler.learn_one
# StandardScaler.transform_one
```

---

P.S. To be frank, me and my team find it very questionable that `Pipeline` by default has the side effect of `learn_*` during inference (`predict_*`/`transform_*`). For we'd like to correctly estimate, how the model responds to unknown categorical features/values until in has been fit to fresh data. From my personal experience, systems-engineering-wise, instant model fit, mentioned in the docstrings, is more of a luxury than a common opportunity.

If anything, it's the side effect behaviour that should be optional, not vice-versa. Plus i don't think this side-effect is emphasized enough in the docs. 

Anyway, respecting the authors' vision i'asking for just a small adjustment.

MaxHalford:
I'm actually convinced. I agree that the current behavior is too advanced, and should be opt-in. I'll implement the necessary changes, and see what breaks (probably a lot of stuff) :)

PR #1270, opened 2023-06-22
Fix stateless inference for transformer-only pipelines.

ColdTeapot273K:
Addresses #1269

### DIFF
--- a/river/compose/pipeline.py
+++ b/river/compose/pipeline.py
@@ -821,8 +821,7 @@ class Pipeline(base.Estimator):
         """
         X, last_step = self._transform_many(X=X)
         if isinstance(last_step, base.MiniBatchTransformer):
-            if not self._STATELESS:
-                if not last_step._supervised:
+            if not self._STATELESS and not last_step._supervised:
                     last_step.learn_many(X)
             return last_step.transform_many(X)
         return X
--- a/river/compose/pipeline.py
+++ b/river/compose/pipeline.py
@@ -192,6 +192,7 @@ def pure_inference_mode():
         yield
     finally:
         Pipeline._STATELESS = False
+        # pass
 
 
 class Pipeline(base.Estimator):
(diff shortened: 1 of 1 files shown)


river/compose/pipeline.py
```diff
@@ -568,13 +567,17 @@ def _transform_one(self, x: dict):
             if utils.inspect.ischildobject(obj=t, class_name="AnomalyFilter"):
                 continue
 
+            print(f"{t=}")
```
MaxHalford, line 570: Can you remove `print` statements?

river/compose/pipeline.py
```diff
@@ -590,10 +593,14 @@ def transform_one(self, x: dict, **params):
...
 
         """
+        print(f"{self._STATELESS=}")
         x, last_step = self._transform_one(x)
+
         if isinstance(last_step, base.Transformer):
-            if not last_step._supervised:
-                last_step.learn_one(x)
+            if not self._STATELESS:
+                if not last_step._supervised:
```
MaxHalford, line 601:
```suggestion
            if not self._STATELESS and not last_step._supervised:
```

no?

river/compose/pipeline.py
```diff
@@ -815,8 +822,9 @@ def transform_many(self, X: pd.DataFrame):
         """
         X, last_step = self._transform_many(X=X)
         if isinstance(last_step, base.MiniBatchTransformer):
-            if not last_step._supervised:
-                last_step.learn_many(X)
+            if not self._STATELESS:
+                if not last_step._supervised:
```
MaxHalford, line 826:
```suggestion
            if not self._STATELESS and not last_step._supervised:
```

MaxHalford (changes requested):
Great catch! Could you add a unit test to test this logic?

river/compose/pipeline.py
ColdTeapot273K: Sure, just wanted to keep it in line with how things a written in other methods :)

MaxHalford:
If it's ok I'll close this, as I'm integrating the same modification in #1277

PR #1271, opened 2023-06-22
[WIP] Feature/ online ordinal encoder

ColdTeapot273K:
The idea:

Ordinal encoder for categoricals that robustly accomodates for:
- unfamilliar category values
- empty category values

Achieved by:
- using only positive integers for encoding (no custom mappings like in `sklearn` implementation)
- reserving two first values for unknown and empty category values

Note:
- positive-only indexing is useful when passing output to some tensor-based modeling frameworks which require indexing tensors and don't have flexible indexing things like `VectorDict` (i'm mixing `river` with jax-based PPL for online Bayesian modeling, proper `OrdinalEncoder` is very valuable)
- naturally i have included mini-batch implementation aswell

Example tests:

```python
transformer = OrdinalEncoder()
# %%
X = [
    {"country": "France", "place": "Taco Bell"},
    {"country": None, "place": None},
    {"country": "Sweden", "place": "Burger King"},
    {"country": "France", "place": "Burger King"},
    {"country": "Russia", "place": "Starbucks"},
    {"country": "Russia", "place": "Burger King"},
    {"country": "Russia", "place": "Taco Bell"},
    {"country": "Sweden", "place": "Taco Bell"},
    {"country": "Sweden", "place": "Burger King"},
    {"country": "France", "place": "Taco Bell"},
    {"country": "France", "place": "Burger King"},
    {"country": "Sweden", "place": "Starbucks"},
    {"country": "Sweden", "place": "Starbucks"},
    {"country": None, "place": None},
]

transformer = OrdinalEncoder()
for x in X:
    print(transformer.transform_one(x))
    transformer = transformer.learn_one(x)
# {'country': 1, 'place': 1}
# {'country': 1, 'place': 1}
# {'country': 1, 'place': 1}
# {'country': 2, 'place': 4}
# {'country': 1, 'place': 1}
# {'country': 5, 'place': 4}
# {'country': 5, 'place': 2}
# {'country': 4, 'place': 2}
# {'country': 4, 'place': 4}
# {'country': 2, 'place': 2}
# {'country': 2, 'place': 4}
# {'country': 4, 'place': 5}
# {'country': 4, 'place': 5}
# {'country': 0, 'place': 0}


transformer = OrdinalEncoder()
transformer.learn_many(pd.DataFrame(X))

assert transformer.categories_['country'] == {None: 0, 'France': 2, 'Sweden': 3, 'Russia': 4}
assert transformer.categories_['place'] == {None: 0, 'Taco Bell': 2, 'Burger King': 3, 'Starbucks': 4}

transformer.transform_many(pd.DataFrame(X))
#	country	place
#0	2	2
#1	0	0
#2	3	3
#3	2	3
#4	4	4
#5	4	3
#6	4	2
#7	3	2
#8	3	3
#9	2	2
#10	2	3
#11	3	4
#12	3	4
#13	0	0



```

### DIFF
--- a/river/preprocessing/ordinal.py
+++ b/river/preprocessing/ordinal.py
@@ -1,0 +1,91 @@
+import collections
+import math
+import numbers
+
+import numpy as np
+import pandas as pd
+
+from river import base
+
+
+class OrdinalEncoder(base.MiniBatchTransformer):
+    # INFO: makes sense to add None initially? so it'll be easier to spot
+    def __init__(self, handle_unknown="use_reserved_value", unknown_value=1, encoded_missing_value=0):
+        # self.categories = categories
+        # self.dtype = dtype
+        self.handle_unknown = handle_unknown
+        self.unknown_value = unknown_value
+        self.encoded_missing_value = encoded_missing_value
+        self.categories_ = collections.defaultdict(dict)
+        # self.categories_ = collections.defaultdict(collections.defaultdict)
+
+    def learn_one(self, x):
+        for i, xi in x.items():
+            if self.handle_unknown == "use_reserved_value":
+                # if self.categories_[i].get(xi, None) is None:
+                if xi not in self.categories_[i]:
+                    if xi is None:
+                        self.categories_[i][xi] = self.encoded_missing_value
+                    else:
+                        self.categories_[i][xi] = 2 + len(self.categories_[i])
+        # self.abs_max[i].update(xi)
+
+        return self
+
+    def transform_one(self, x):
+        # return {i: self.categories_[i][xi] for i, xi in x.items()}
+        return {i: self.categories_[i].get(xi, self.unknown_value) for i, xi in x.items()}
+
+    def learn_many(self, X):
+        for col in X.columns:
+            # self.values[col].update(X[col].unique())
+            known_uniqs = np.array(
+                list(self.categories_[col].keys())
+            )  # INFO: convert to numpy array for np.isin to work properly, otherwise unexpected behavior
+            current_uniqs = X[col].unique()
+
+            # if None in known_uniqs:
+            #     current_uniqs.dropna(inplace=True)
+            # elif
+            upd_encoding = dict()
+
+            # current_uniqs = pd.DataFrame(X)['country'].dropna().unique()
+
+            # new_unique_mask = np.isin(current_uniqs, known_uniqs, assume_unique=True, invert=True)
+            new_unique_mask = np.isin(current_uniqs, known_uniqs, assume_unique=True, invert=True)
+
+            new_uniqs = current_uniqs[new_unique_mask]
+            print(f"{known_uniqs=}")
+            print(f"{current_uniqs=}")
+            print(f"{new_uniqs=}")
+
+            # INFO: record and mask the first encounters of None or equivalents in the batch
+            if None in new_uniqs:
+                print("None detected")
+
+                upd_encoding.update({None: self.encoded_missing_value})
+                none_mask = np.isin(new_uniqs, [None], assume_unique=True, invert=True)
+                new_uniqs = new_uniqs[none_mask]
+
+            # INFO: process first encounters of other unique values except None or equivalents
+            upd_encoding.update(
+                {
+                    k: v + 2 + len(known_uniqs)
+                    # for k, v in zip(current_uniqs, range(0, current_uniqs.shape[0]))
+                    for v, k in enumerate(new_uniqs)
+                    # if (k not in known_uniqs) and (k is not None)
+                }
+            )
+
+            print(upd_encoding)
+
+            self.categories_[col].update(upd_encoding)
+
+        return self
+
+    def transform_many(self, X):
+        X_encoded = X.copy()
+        for col in X_encoded.columns:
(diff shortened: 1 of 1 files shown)


river/preprocessing/ordinal.py
```diff
@@ -0,0 +1,91 @@
...
+import numbers
+
+import numpy as np
+import pandas as pd
+
+from river import base
+
+
+class OrdinalEncoder(base.MiniBatchTransformer):
+    # INFO: makes sense to add None initially? so it'll be easier to spot
```
MaxHalford, line 12: Could you add a docstring? :)

river/preprocessing/ordinal.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, handle_unknown="use_reserved_value", unknown_value=1, encoded_missing_value=0):
```
MaxHalford, line 13: How many possible values are there for `handle_unknown`? If it's just a boolean, I rather have an `use_reserved_value_for_unknowns: bool` parameter.

MaxHalford: Also, wouldn't `-1` make more sense for the `unknown_value`? Isn't it surprising for the first value encountered, as well as unknown values to both be assigned `1`?

MaxHalford: Which is also making me think, because this method is unsupervised, why don't you do the update logic in `transform_one`? That way there won't be any unknown values, because you update before transforming. I believe we do that in some other places in River

river/preprocessing/ordinal.py
```diff
@@ -0,0 +1,91 @@
...
+            # elif
+            upd_encoding = dict()
+
+            # current_uniqs = pd.DataFrame(X)['country'].dropna().unique()
+
+            # new_unique_mask = np.isin(current_uniqs, known_uniqs, assume_unique=True, invert=True)
+            new_unique_mask = np.isin(current_uniqs, known_uniqs, assume_unique=True, invert=True)
+
+            new_uniqs = current_uniqs[new_unique_mask]
+            print(f"{known_uniqs=}")
```
MaxHalford, line 58: You can't leave `print` calls

river/preprocessing/ordinal.py
```diff
@@ -0,0 +1,91 @@
...
+        # self.dtype = dtype
+        self.handle_unknown = handle_unknown
+        self.unknown_value = unknown_value
+        self.encoded_missing_value = encoded_missing_value
+        self.categories_ = collections.defaultdict(dict)
+        # self.categories_ = collections.defaultdict(collections.defaultdict)
+
+    def learn_one(self, x):
+        for i, xi in x.items():
+            if self.handle_unknown == "use_reserved_value":
```
MaxHalford, line 24: So what happens in the other case? Nothing?

MaxHalford (commented):
I'm generally in favor of this pull request! I have a few questions though

river/preprocessing/ordinal.py
ColdTeapot273K: > Also, wouldn't `-1` make more sense for the `unknown_value`?

Please see the notes.
I've rejected this idea because there's no '-1' index in an array/tensor
This might seem not very conventional since we're used to '-1' sentinel value but positive-only indexing makes the component more reusable as i've mentioned.

> Isn't it surprising for the first value encountered, as well as unknown values to both be assigned `1`

The idea was to keep 2 options available:
1. same codes for unknown value & missed value
2. different codes for unknown value & missed value

But both codes should provisioned at the start since both should be expected. 
I have no principled idea wether it should be
- unknown = 1 
- missed = 0 
or vice-versa.

river/preprocessing/ordinal.py
ColdTeapot273K: Ofc! 
That was intended for debugging. I was going to remove them once we've settled on implementation, hence "WIP" tag (sorry i should have mentioned that)

river/preprocessing/ordinal.py
ColdTeapot273K: Open to suggestions! 

At the moment i just implemented what was relevant in my usecase.

Mimicking `sklearn` api here a bit. Tho raising an error for unkowns like `sklearn` does would be weird in online setting. Unless there's an option to supply a predefined mapping values -> encodings. Then unknown values start make sense in online setting. 
But i'm not sure we want to support that in `OrdinalEncoder` - user can just use `FunctionTransformer` if they want a mapping.

The only other option is to have the unknowns and the missing values map to the same code, as mentioned in the comments above. I can implement that.

river/preprocessing/ordinal.py
MaxHalford: Ok it's fine, I'm convinced. I'm not sure the +2 works if you specify different values for missing and unknown parameters.

Also, just another question: in your examples you explicitly set missing unknown values to None. But couldn't you instead preserve the sparse nature of the dict and implicitly assume missing keys are None? Not sure I'm being clear

MaxHalford:
If it's ok, I'll merge this PR and tweak it a tiny bit. I think it's valuable to have.

The logic is clearer to me now we've made `predict_one` stateless.

PR #1286, opened 2023-07-18
Added Jaccard Accuracy (also found in MOA)

killianLM:
This metric is interesting because it works fully with any multi-label dataset, and avoids possible result bias that can be seen with the Macro or Micro Average wrappers on metrics like the Accuracy when some labels are rarely used.

We thought that this metric would be interesting for anyone doing multi-label classification, and also for anyone coming from the MOA platform as it is one of the metrics found there that was missing in River.

### DIFF
--- a/river/metrics/multioutput/__init__.py
+++ b/river/metrics/multioutput/__init__.py
@@ -7,6 +7,7 @@ from .exact_match import ExactMatch
 from .macro import MacroAverage
 from .micro import MicroAverage
 from .per_output import PerOutput
+from .accuracy_moa_jaccard import Accuracy_MOA_Jaccard
 
 __all__ = [
     "base",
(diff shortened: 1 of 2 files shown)


MaxHalford:
Hey there. I understand the need for this metric. So yes let's add this to River. However, I'm not comfortable with the implementation for a bunch of reasons. For instance, you should use the already existing confusion matrix for multi-output targets instead of doing your own sums. The method names are also not correct.

You have two options: either I do a big review and you change a lot of stuff, either we merge and I take care of refactoring what needs refactoring. In both cases you will learn. The second option will be faster. What do you prefer?

killianLM:
Hi there, sorry for the delay in answering.

I think it would be easier if you were to handle the refactoring, as well as faster like you said.
Thank you very much.

MaxHalford:
Cool, no worries. Thanks for suggesting/contributing this valuable metric!

PR #1301, opened 2023-07-30
Add MultivariateGaussian

MarekWadinger:
Implemented MultivariateGaussian, as part of ContinuousDistributions, taking advantage of EmpiricalCovariance class, extending its example in doctest, showing basic usage of the class.

### DIFF
--- a/river/proba/gaussian.py
+++ b/river/proba/gaussian.py
@@ -249,8 +249,10 @@ class MultivariateGaussian(base.ContinuousDistribution):
 
     def __repr__(self):
         mu_str = ", ".join(f"{m:.3f}" for m in self.mu)
-        var_str = "\n ".join("[" + " ".join(f"{s:.3f}" for s in row) + "]" for row in self.var)
-        return f"𝒩(μ=({mu_str}),\nσ^2=({var_str}))"
+        var_str = "\n".join(
+            "        [" + ", ".join(f"{s:.3f}" for s in row) + "]" 
+            for row in self.var)
+        return f"𝒩(\n    μ=({mu_str}),\n    σ^2=(\n{var_str}\n    )\n)"
 
     def update(self, x, w=1.0):
         if w != 1.0:
--- a/river/proba/gaussian.py
+++ b/river/proba/gaussian.py
@@ -1,11 +1,16 @@
 from __future__ import annotations
 
 import math
+import warnings
 
+import numpy as np
+from scipy.stats import multivariate_normal
+
+from river import covariance
 from river import stats
 from river.proba import base
 
-__all__ = ["Gaussian"]
+__all__ = ["Gaussian", "MultivariateGaussian"]
 
 
 class Gaussian(base.ContinuousDistribution):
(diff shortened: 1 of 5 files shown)


river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+    ...     p_ = p_.update(x['blue'])
+    >>> p.sigma[0][0] == p_.sigma
+    True
+
+    Initiation of class from state is currently not implemented
+    >>> p = MultivariateGaussian()._from_state(
+    ...     0, 0, 0, 0)  # doctest: +IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+    ...
+    NotImplementedError: from_state_ not implemented yet.
```
MaxHalford, line 196: I also would not mention this if it's not implemented.

river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+
+    @property
+    def n_samples(self):
+        if not self._var.matrix:
+            return 0.0
+        else:
+            return list(self._var.matrix.values())[-1].mean.n
+
+    @property
+    def mu(self):
```
MaxHalford, line 217: This should return a dictionary, not a list

river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+
+    def cdf(self, x):
+        x = list(x.values())
+        return multivariate_normal(self.mu, self.var, allow_singular=True).cdf(x)
+
+    def sample(self):
+        return multivariate_normal(
+            self.mu,
+            self.var,
+        ).rvs()
```
MaxHalford, line 293:
```suggestion
        return float(multivariate_normal(
            self.mu,
            self.var,
        ).rvs())
```

It's more "River-like" to return a `float`, and not a `np.float64`

MaxHalford (changes requested):
Here's a first review! Overall, very good job. There's just some details to iron out :)

river/proba/gaussian.py
MarekWadinger: Thank you for your valuable suggestions. I made properties "mu" and "mode" return dictionary.

river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+        return list(
+            {
+                key1: values.mean.get()
+                for (key1, key2), values in self._var.matrix.items()
+                if key1 == key2
+            }.values()
+        )
+
+    @property
+    def var(self):
```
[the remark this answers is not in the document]
MarekWadinger: I made var and sigma return DataFrames. Seems to serve well for representations which are now aligned on dot.

MarekWadinger (commented):
Implementing suggestions from Max Halford.

river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+    >>> p = p.update(x.to_dict(), t=t + td(seconds=1), **{"w":2})
+
+    >>> from river.proba import Gaussian
+    >>> p = MultivariateGaussian()
+    >>> p_ = Gaussian()
+    >>> for t, x in X.iterrows():
+    ...     p = p.update(x.to_dict())
+    ...     p_ = p_.update(x['blue'])
+    >>> p.sigma[0][0] == p_.sigma
+    True
```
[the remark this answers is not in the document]
MarekWadinger: Definitely! Along with var, sigma now returns pd.DataFrame which allows feature name indexing.

river/proba/gaussian.py
```diff
@@ -90,3 +95,203 @@ def sample(self):
...
+
+    Weighted samples are currently not implemented. Updates with default w = 1.
+    >>> p = p.update(x, w=2)
+
+    Retrieving current state in nice format is simple
+    >>> p
+    𝒩(μ=(0.385, 0.376, 0.501),
+    σ^2=([0.069 0.019 -0.004]
+     [0.019 0.100 -0.044]
+     [-0.004 -0.044 0.078]))
```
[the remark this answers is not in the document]
MarekWadinger: Improves readability a lot. I modified the representation to match the idea.

river/proba/gaussian.py
```diff
@@ -242,29 +241,29 @@ def var(self):
...
     def __repr__(self):
-        mu_str = ", ".join(f"{m:.3f}" for m in self.mu)
-        var_str = "\n ".join("[" + " ".join(f"{s:.3f}" for s in row) + "]" for row in self.var)
-        return f"𝒩(μ=({mu_str}),\nσ^2=({var_str}))"
+        mu_str = ", ".join(f"{m:.3f}" for m in self.mu.values())
+        var_str = self.var.to_string(float_format="{:0.3f}".format, header=False, index=False)
+        var_str = "        [" + var_str.replace("\n", "]\n        [") + "]"
+        return f"𝒩(\n    μ=({mu_str}),\n    σ^2=(\n{var_str}\n    )\n)"
 
     def update(self, x, w=1.0):
```
MaxHalford, line 259:
```suggestion
    def update(self, x):
```

river/proba/gaussian.py
```diff
@@ -274,7 +273,7 @@ def __call__(self, x):
         var = self.var
         if var is not None:
             try:
-                return multivariate_normal(self.mu, var).pdf(x)
+                return multivariate_normal([*self.mu.values()], var).pdf(x)
```
MaxHalford, line 275: There might be a bug here. You have to align `self.mu` with `x`. For instance:

```py
return multivariate_normal([self.mu[i] for i in x], var).pdf(x)
```

See what I mean? Same goes `.cdf(x)`

river/proba/gaussian.py
```diff
@@ -286,13 +285,17 @@ def __call__(self, x):
 
     def cdf(self, x):
         x = list(x.values())
-        return multivariate_normal(self.mu, self.var, allow_singular=True).cdf(x)
+        return multivariate_normal([*self.mu.values()], self.var, allow_singular=True).cdf(x)
 
     def sample(self):
```
MaxHalford, line 289: This method should return a dictionary :)

MaxHalford (changes requested):
Looking good! Still a few comments :). But you're on the right track, keep it up

river/proba/gaussian.py
MarekWadinger: Thank you for pointing out this. I realized that neither mu and var were sorted the same way. Made both sort their keys alphabetically. Now I let x adapt the order of keys, which is consistent between mu and var.

river/proba/gaussian.py
MarekWadinger: Fixed. :)

river/proba/gaussian.py
MarekWadinger: Right, consistency is the key. :) Done.

MarekWadinger (commented):
Implemented suggestions from Max Halford.

MarekWadinger:
Introduced a new base class as well, I realized using proba.ContinuousDIstribution comes with type hinting which is not consistent with proba.MultivariateGaussian. @MaxHalford, you introduced this idea in discussion, please can you review, whether it is consistent? :)

MaxHalford:
I'm merging because it's very good! Congrats @MarekWadinger :)

MarekWadinger:
Thank you, I'm very happy to hear that! :) I see lot of potential in this project and I'd be happy to contribute further. 

Currently, I'm playing around with conversion to sklearn to employ my model in validation scheme using sklearn's hyperparameter tuning. I might have some questions/suggestions related to getting/setting parameters in River2SKLClassifier, which is troubling me a bit.

MaxHalford:
We have some basic hyperparameter optimization capabilities in River. Have you checked it out? I'm not sure it's well documented 😅

MarekWadinger:
I did, actually tried to use it with an ensemble of weak unsupervised learners to optimize their parameters, taking majority vote as ground truth. Turned out protection of learners was the key challenge and I had to postpone further work.

Nevertheless, here I need to fit to the validation framework of [this repo](https://github.com/CN-TU/py-outlier-detection-stream-data/tree/master), to validate my results against algorithms in recent comparison study of [streaming anomaly detectors](https://www.sciencedirect.com/science/article/pii/S0957417423014963).

Here they use sklearn's random grid search to avoid single point of failure which is wrong parameter selection.

PR #1309, opened 2023-08-06
MSTS ARFClassifier

danielnowakassis:
Dynamic Selection of Classifiers is a really interesting area in batch learning (https://deslib.readthedocs.io/en/latest/).

For each instance desired to classify, a different subset of classifiers from an ensemble can be selected for voting this instance.

Most of the methods (99%) get the 7NN from the instance (known as the region of competence) and evaluate the best classifiers to vote based on this region of competence.

In a streaming scenario, kNN + Ensemble cost can be significant. My paper proposes a method without the use of kNN.

All the additions to ARF have comments. Unfortunately, this can not be a Wrapper method, since it involves changing some ARF innerstructures. If it gets well, i can also implement for other ensemble methods.


smastelini:
Update: I will take care of making the minor modifications needed before merging this PR. I will create a PR to @danielnowakassis's fork with the changes as soon as possible.

PR #1318, opened 2023-08-26
Fix HDDM_*

smastelini:
Fixes #1315.

I've been able to track down the issue. I have used the MOA code as a reference, so I am unaware of possible minor differences between River and Frouros' code that might result in slight accuracy differences in the code example presented at #1315.

@ngabrek, can you please check if the fix addresses your problem?


ngabrek:
@smastelini, thank you for the fix! Now, everything works perfectly fine.

MaxHalford:
@smastelini can you add something to the release notes? :)

smastelini:
Yeap! I did that afterwards :)

PR #1326, opened 2023-10-03
added web traffic dataset and info

jacowp357:
Hi Max,

As we talked about here is the web traffic dataset update.

I'm not sure how to treat multiple target variables with `stream.iter_csv`, consequently, I added only one for the time being.


jacowp357:
removed the CSV file, added url and changed it to `base.RemoteDataset`

MaxHalford:
Cheers @jacowp357

PR #1327, opened 2023-10-03
Add _from_state methods to MultivariateGaussian and EmpiricalCovariance

MarekWadinger:
Hey there 👋

I was thinking about adding _from_state method to [MultivariateGaussian](https://github.com/online-ml/river/commit/e767f26692ac8bf4cae9d38bfacfa97080cf5f49) and [EmpiricalCovariance](https://github.com/online-ml/river/commit/be0b3257b579dc7401608d5dda1106e1bdf92cc2). 

I believe it might aid the versatility of the classes usage. Nevertheless, I must admit that it might be little tricky for some users to retrieve covariance as dictionary of correct form.

Here is a pull request with proposed changes! (Did not open discussion as I considered this quite minor, sorry for inconvenience)

Please let me know what you think.

Thank you 🙏

### DIFF
--- a/river/covariance/emp.py
+++ b/river/covariance/emp.py
@@ -178,35 +178,40 @@ class EmpiricalCovariance(SymmetricMatrix):
         mean_arr = X_arr.mean(axis=0)
         cov_arr = np.cov(X_arr.T, ddof=self.ddof)
 
+        n = len(X)
         mean = dict(zip(X.columns, mean_arr))
         cov = {
             (i, j): cov_arr[r, c]
             for (r, i), (c, j) in itertools.combinations_with_replacement(enumerate(X.columns), r=2)
         }
 
-        for i, j in itertools.combinations(sorted(X.columns), r=2):
+        self = self._from_state(n=n, mean=mean, cov=cov, ddof=self.ddof)
+
+        return self
+
+    @classmethod
+    def _from_state(cls, n, mean, cov, *, ddof=1):
+        new = cls(ddof=ddof)
+        for i, j in itertools.combinations(sorted(mean.keys()), r=2):
             try:
-                self[i, j]
+                new[i, j]
             except KeyError:
-                self._cov[i, j] = stats.Cov(self.ddof)
-            self._cov[i, j] += stats.Cov._from_state(
-                n=len(X),
+                new._cov[i, j] = stats.Cov(new.ddof)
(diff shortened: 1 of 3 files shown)


river/covariance/emp.py
```diff
@@ -178,35 +194,63 @@ def update_many(self, X: pd.DataFrame):
...
+        cls (type)
+            The class type.
+        n (int)
+            The number of data points.
+        mean (dict)
+            A dictionary of variable means.
+        cov (dict)
+            A dictionary of covariance or variance values.
+        ddof (int, optional)
+            Degrees of freedom for covariance calculation. Defaults to 1.
```
smastelini, line 223: There is no need to use type annotations in the doctrings. The docs will take it directly from the method signature :)

river/covariance/emp.py
```diff
@@ -178,35 +194,63 @@ def update_many(self, X: pd.DataFrame):
...
+                n=n,
                 mean_x=mean[i],
                 mean_y=mean[j],
                 cov=cov.get((i, j), cov.get((j, i))),
-                ddof=self.ddof,
+                ddof=new.ddof,
             )
 
-        for i in X.columns:
+        for i in sorted(mean.keys()):
```
smastelini, line 247: Is this sorting operation a must? I think we can save a couple of milliseconds and bring no impacts, functionality-wise. What do you think @MarekWadinger?

river/proba/gaussian.py
```diff
@@ -214,7 +214,11 @@ def __init__(self, seed=None):
         super().__init__(seed)
         self._var = covariance.EmpiricalCovariance(ddof=1)
 
-    # TODO: add method _from_state to initialize model (for warm starting)
+    @classmethod
+    def _from_state(cls, n, mean, cov, ddof):
```
smastelini, line 218: The `seed` parameter will always be `None` in this case. It should be optionally passed as a parameter to ensure reproducibility.

smastelini (changes requested):
Hi @MarekWadinger, thanks for this PR. It looks good. 

I've added a couple minor comments that needed to be addressed before merging.

river/covariance/emp.py
smastelini: The same applies to the previous for loop (the one taking combinations of the x's keys.

smastelini (approved):
LGTM! Thank you so much for the changes

Issue #1324, opened 2023-09-24
DBSTREAM: multiple issues


donny741:
Hi, I'm new to Python and data science, and I'm currently exploring the clustering algorithms. When comparing the DBSTREAM algorithm implementation with the original paper I noticed some inconsistencies:
- It seems the `-` signs before the `fading_factor` are missing [here](https://github.com/online-ml/river/blob/main/river/cluster/dbstream.py#L256) and [here](https://github.com/online-ml/river/blob/main/river/cluster/dbstream.py#L269) what prevents the clusters with low weight from being removed
- Length of the `micro_clusters` list is used as an index key when creating the new cluster ([here](https://github.com/online-ml/river/blob/main/river/cluster/dbstream.py#L185)) what can override the existing clusters because the length is decremented when clusters are removed from the list ([here](https://github.com/online-ml/river/blob/main/river/cluster/dbstream.py#L261))
- `clustering_is_up_to_date` variable introduced with [this PR](https://github.com/online-ml/river/pull/1213) is never being set to `True`

MaxHalford:
Hey @hoanganhngo610, do you have time to take a look at this issue? :)

hoanganhngo610:
@MaxHalford I will have a look at the issues by the end of this week. Also, thank you @donny741 for raising the issues!

hoanganhngo610:
@donny741 @MaxHalford I have just made a PR to fix the points that @donny741 proposed within this issue. Hope that the changes are satisfactory.

If you have some time, please do not hesitate to have a look through the PR and leave comments wherever necessary!

MaxHalford:
Good job! Just merged the PR.

@donny741 I'll let you close the PR if you're satisfied

donny741:
Looks good! Thank you!

PR #1336, opened 2023-10-09
DBSTREAM fix (related to issue #1324)

hoanganhngo610:
This PR fixes multiple issues proposed in issue #1324 by @donny741, with the following changes:
- Addition of the `-` sign before the `fading_factor` in accordance with the algorithm 2 proposed by Hashler and Bolanos to allow clusters with low weights to be removed.
- The new `micro_cluster` is added with the key derived from the maximum key of the existing micro clusters. If the set of micro clusters is still empty (`len = 0`), a new micro cluster is added with key `0`.
- `cluster_is_up_to_date` is set to `True` at the end of the `self._recluster()` function.


MaxHalford:
Good job @hoanganhngo610. Can you add the relevant entries to  UNRELEASED.md?

hoanganhngo610:
@MaxHalford UNRELEASED.md entries added!

PR #1338, opened 2023-10-14
DBSTREAM: fix density graph timestamps

donny741:
When the timestamp is set to 0 it may result in the immediate removal of the density graph entry because during cleanup 0 will be used as a current timestamp and the fading value will be incorrect ([here](https://github.com/online-ml/river/blob/main/river/cluster/dbstream.py#L276))


MaxHalford:
Nice! Can you add something to unreleased.md?

donny741:
Added ✅

PR #1339, opened 2023-10-14
DBSTREAM: fix microcluster labeling

donny741:
Neighbors of neighbors are not added to the `seed_set` so they receive different labels even though they have adjacent clusters.
Fixing by changing the `if` statement so neighbors without labels are appended to the `seed_set` and then assigned the label in following iterations.


MaxHalford:
This is excellent. Special kudos for writing unit tests!

PR #1340, opened 2023-10-15
dbstream: fix adjacency matrix building

donny741:
It is possible for the microcluster (`mc`) to be removed, but for the shared density graph entry (`Sij`) associated with that microcluster to remain because:
- The threshold for removing `Sij` is `α` (a.k.a. `intersection_factor`) times lower than the threshold for removing a `mc`;
- and no mechanism would ensure the removal of associated `Sij` when `mc` is removed.

This introduces a `KeyError` during the `weighted_adjacency_matrix` building step when looking up removed `mc`. Adding a `try` block to account for that

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -16,7 +16,7 @@ River's mini-batch methods now support pandas v2. In particular, River conforms 
   - `cluster_is_up_to_date` is set to `True` at the end of the `self._recluster()` function.
   - Shared density graph update timestamps are initialized with the current timestamp value
   - `neighbour_neighbours` are appended correctly to the `seed_set` when generating cluster labels
-  - When building weighted adjacency matrix the alhorithm accounts for possibly orphaned entries in shared density graph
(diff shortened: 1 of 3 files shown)


docs/releases/unreleased.md
```diff
@@ -16,6 +16,7 @@ River's mini-batch methods now support pandas v2. In particular, River conforms
   - `cluster_is_up_to_date` is set to `True` at the end of the `self._recluster()` function.
   - Shared density graph update timestamps are initialized with the current timestamp value
   - `neighbour_neighbours` are appended correctly to the `seed_set` when generating cluster labels
+  - When building weighted adjacency matrix the alhorithm accounts for possibly orphaned entries in shared density graph
```
MaxHalford, line 19:
```suggestion
  - When building weighted adjacency matrix the algorithm accounts for possibly orphaned entries in shared density graph
```

MaxHalford (approved):

PR #1357, opened 2023-10-31
Implementation of the StandardAbsoluteDeviation (SAD) anomaly detection algorithm

hoanganhngo610:
This PR contains the implementation of the Standard Absolute Deviation (SAD) anomaly detection algorithm.

This implementation is adapted from the implementation within [PySAD](https://github.com/selimfirat/pysad/blob/master/pysad/models/standard_absolute_deviation.py) (Python Streaming Anomaly Detection) package, which calculates the anomaly score by dividing the deviation from the mean/median to the standard deviation of all points seen within the data stream. This idea is adapted from the paper by [Hochenbaum et al. (2017)](https://doi.org/10.48550/ARXIV.1704.07706).

Despite being a fairly simple algorithm, the implementation of this algorithm allows the `anomaly` submodule of `River` to have a solid competitor to conduct any benchmarking work.

There is no unit test for this algorithm since the result has already been compared locally with the implementation in PySAD, which yielded exactly similar results. Creating unit tests upstream would require the installation of another package, which seem unnecessary.

### DIFF
--- a/river/anomaly/sad.py
+++ b/river/anomaly/sad.py
@@ -5,7 +5,7 @@ from river import anomaly, stats
 __all__ = ["StandardAbsoluteDeviation"]
 
 
-class StandardAbsoluteDeviation(anomaly.base.AnomalyDetector):
+class StandardAbsoluteDeviation(anomaly.base.SupervisedAnomalyDetector):
     r"""Standard Absolute Deviation (SAD).
 
     SAD is the model that calculates the anomaly score by using the deviation from the mean/median, divided by the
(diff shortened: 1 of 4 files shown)


hoanganhngo610:
@MaxHalford Would you mind having a look through this PR? There are two problems that are making the tests fail, including 
- `mypy`: I believe these errors do not come from the files that I modified.
- `test_estimators` tests: I believe I have added SAD to the list of estimators that are expected to be ignored upon testing, but somehow they are still tested. I personally do not encounter these errors running pytest locally.
Thank you so much in advance!

river/anomaly/sad.py
```diff
@@ -0,0 +1,89 @@
+from __future__ import annotations
+
+from river import anomaly, stats
+
+__all__ = ["StandardAbsoluteDeviation"]
+
+
+class StandardAbsoluteDeviation(anomaly.base.AnomalyDetector):
```
MaxHalford, line 8: The issue with this method is that it is univariate. You're trying to fit into the multivariate anomaly detection framework, which is not the right way.

What you should do is inspire yourself from `anomaly.GaussianScorer`. It's also a univariate anomaly detection method.

Let me know if this isn't clear

MaxHalford (changes requested):

river/anomaly/sad.py
hoanganhngo610: @MaxHalford I think I can understand the idea, and have also implemented it in my upcoming commits. It would be much appreciated if you can have a look through and see whether the new approach is appropriate!

MaxHalford (approved):
Cool! Can you add a note to `unreleased.md`?

hoanganhngo610:
@MaxHalford Thank you so much for approving the changes! The entry has also been added to the UNRELEASED.md file, so please proceed to merge at your convenience!

PR #1453, opened 2023-11-15
Update scale.py

icaroperetti:
Fix division by zero issue by adding check for zero divisor in division operation.

### DIFF
--- a/river/preprocessing/scale.py
+++ b/river/preprocessing/scale.py
@@ -28,7 +28,7 @@ def safe_div(a, b):
     the denominator can be nil if a feature has no variance.
 
     """
-    return a / b if b else 0.0
+    return a / b if b != 0 else 0.0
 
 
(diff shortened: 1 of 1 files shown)


icaroperetti:
I'm using python 3.8 in my enviroment. Maybe the RuntimeWarning do not occur for other versions.

icaroperetti:
Using Python 3.8 in my enviroment, maybe the error do not occur in other versions.

MaxHalford (commented):
I'm sorry but I don't understand what this edit changes. The `if` already handles the case where the value is equal to 0.

Why are you attempting to fix this? Did you get a bug in a script or something?

icaroperetti:
I'm getting a RuntimeWarning

MaxHalford:
I have a hard time believing you get no warning with this change :). Can you share an example?

PR #1474, opened 2023-12-12
Drift detectors

gabrieljaguiar:
In recent surveys new drift detectors have been used, while in River we do not have acess to those implementations. FHDDM and FHDDMS have been cited more than 250 times and are present in recent literature. Therefore, the presence of those methods are needed in a package like river.


MaxHalford:
I'll let @smastelini handle this one 🙏

My only input is that I don't like the idea of adding more classes. Especially if these are just variants of a HDDM. I would rather have a single HDDM class with parameters to choose between variants. But I don't enough about drift detection to tell if this is possible or not.

smastelini:
Hey @gabrieljaguiar, you told me these methods are not directly related to HDDM. Care to expand on that?

gabrieljaguiar:
Hi @MaxHalford and @smastelini . 

So, the only common thing between FHDDM and HDDM is the Hoeffing Inequality. While HDDM maintain mean standard deviation of the distribution, FHDDM maintain 1 (or 2) window (s) of predictions and use the inequality to determine whether a drift occured or not. Actually, FHDDM would be more similar to ADWIN than HDDM.

river/drift/binary/fhddm.py
```diff
@@ -0,0 +1,102 @@
...
+    represents correct/normal values.
+
+    For example, if a classifier's prediction $y'$ is right or wrong w.r.t. the
+    true target label $y$:
+
+    - 0: Correct, $y=y'$
+
+    - 1: Error, $y \\neq y'$
+
+    *Implementation based on MOA.*
```
smastelini, line 24:
```suggestion

    *Implementation based on MOA.*
```

river/drift/binary/fhddm.py
```diff
@@ -0,0 +1,102 @@
...
+    >>> data_stream = rng.choices([0, 1], k=1000)
+    >>> # Increase the probability of 1's appearing in the next 1000 instances
+    >>> data_stream = data_stream + rng.choices([0, 1], k=1000, weights=[0.7, 0.3])
+    >>> # Update drift detector and verify if change is detected
+    >>> for i, x in enumerate(data_stream):
+    ...     fhddm.update(x)
+    ...     if fhddm.drift_detected:
+    ...         print(f"Change detected at index {i}")
+    ...         exit(0)
+    Change detected at index 1057
```
smastelini, line 52: @gabrieljaguiar, do you think it is doable to reduce the number of instances used in the docstring test?

The idea is to have a good reference example that runs fast so that the complete pytest suite is as quick as possible.

river/drift/binary/fhddm.py
```diff
@@ -0,0 +1,102 @@
...
+        self.n_one = 0
+
+    def _reset(self):
+        self.sliding_window = [0] * self.sliding_window_size
+        self.pointer = 0
+        self.epsilon = math.sqrt(
+            (math.log(1 / self.confidence_level)) / (2 * self.sliding_window_size)
+        )
+        self.u_max = 0
+        self.n_one = 0
```
smastelini, line 81: I would rather avoid repeating code from init. Redundancy might come to haunt us in the future. Please check other detectors in the module for reference of how to deal with the `_reset` method.

river/drift/binary/fhddm.py
```diff
@@ -0,0 +1,102 @@
...
+
+    """
+    def __init__(
+        self, sliding_window_size: int = 100, confidence_level: float = 0.000001
+    ):
+
+        super().__init__()
+        self.sliding_window_size = sliding_window_size
+        self.confidence_level = confidence_level
+        self.sliding_window = [0] * self.sliding_window_size
```
smastelini, line 66: I advise using `collections.deque` instead.

river/drift/binary/fhddm_s.py
```diff
@@ -0,0 +1,122 @@
...
+    >>> data_stream = rng.choices([0, 1], k=1000)
+    >>> # Increase the probability of 1's appearing in the next 1000 instances
+    >>> data_stream = data_stream + rng.choices([0, 1], k=1000, weights=[0.7, 0.3])
+    >>> # Update drift detector and verify if change is detected
+    >>> for i, x in enumerate(data_stream):
+    ...     fhddms.update(x)
+    ...     if fhddms.drift_detected:
+    ...         print(f"Change detected at index {i}")
+    ...         exit(0)
+    Change detected at index 1057
```
smastelini, line 52: Same here (regarding the sample size)

river/drift/binary/fhddm_s.py
```diff
@@ -0,0 +1,122 @@
...
+
+    - 1: Error, $y \\neq y'$
+
+    *Implementation based on MOA.*
+    Parameters
+    ----------
+    sliding_window_size
+        The minimum required number of analyzed samples so change can be detected.
+    confidence_level
+        Threshold to decide if a drift was detected. The default value gives a 99\\% of confidence
```
smastelini, line 30: It is actually more than that, if we want to go into detail, right?

river/drift/binary/fhddm_s.py
```diff
@@ -0,0 +1,122 @@
...
+        short_window_size: int = 25,
+        confidence_level: float = 0.000001,
+    ):
+        super().__init__()
+        self.stack_size = stack_size
+        self.short_window_size = short_window_size
+        self.long_window_size = short_window_size * stack_size
+        self.confidence_level = confidence_level
+
+        self.stack = [0] * (self.short_window_size * self.stack_size)
```
smastelini, line 72: I also suggest using ``collections.deque` here and in other similar instances.

river/drift/binary/fhddm_s.py
```diff
@@ -0,0 +1,122 @@
...
+        self.epsilon_s = math.sqrt(
+            (math.log(1 / self.confidence_level)) / (2 * self.short_window_size)
+        )
+        self.epsilon_l = math.sqrt(
+            (math.log(1 / self.confidence_level)) / (2 * self.long_window_size)
+        )
+        self.pointer = 0
+        self.u_s_max = 0
+        self.u_l_max = 0
```
smastelini, line 93: Same comment about repeating stuff in the `_reset` method

river/drift/binary/fhddm.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def update(self, x):
+        if self.drift_detected:
+            self._drift_detected = False
+            self._reset()
+
+        self._sliding_window.append(x)
+
+        if len(self._sliding_window) == self.sliding_window_size:
+            n_one = self._sliding_window.count(1)
```
smastelini, line 95: For performance reasons, keeping the sum of inputs (which are either 0 or 1) is a better idea. Otherwise, each insertion will have an O(w), where w is the length of the window.

smastelini (changes requested):
Hi @gabrieljaguiar, you streamlined the code a lot! I left one additional request regarding functionality.

Apart from that, would be nice to document how to use both versions of the detector. For example, you can expand the example section. Besides that, please add an entry on `docs/unreleased.md` with the new additions.

river/drift/binary/fhddm.py
gabrieljaguiar: I do not know if I understood correctly, but if I keep the sum I would never know which one to remove from the sliding window when it is full.

smastelini: You will remove the value of `self._sliding_window.popleft()` :D

smastelini (approved):

PR #1498, opened 2024-02-01
Update entropy.py

dansok:
classmethod -> staticmethod


MaxHalford:
Why only make a change at one place? The `_unit_test_params` method is in many places. Anyway, I don't see the value of doing this.

dansok:
it's just better form, that's all

hoanganhngo610:
Hi @dansok. As Max mentioned, since `_unit_test_params` are present in many places,  along with the fact that classmethod and staticmethod are of minimal difference in terms of performance, I suggest keeping anything as is.

Moreover, since this PR has already been inactive for some time, I will proceed to close this.

PR #1504, opened 2024-02-14
Add stream.iter_polars

### LINKED ISSUE #1503: feature req: stream.iter_polars
[polars](https://github.com/pola-rs/polars) is an interesting dataframe library that is getting more and more adopted.
I think we should add support for it on river

niccolopetti:
Fixes #1503


AdilZouitine (commented):
Hey, 
Thanks for your contribution, LGTM :smile:

PR #1509, opened 2024-03-06
Add decomposition methods OnlineSVD, OnlinePCA, OnlineDMD/wC + Hankelizer

### LINKED ISSUE #1366: PCA (SVD)
- Incremental PCA is already [implemented in scikit-learn](https://scikit-learn.org/stable/auto_examples/decomposition/plot_incremental_pca.html) but seems like it is straightforward to implement. See [this paper](http://cs-www.cs.yale.edu/homes/el327/papers/opca.pdf) too. [This paper](https://arxiv.org/abs/1307.0032) looks quite interesting too.
- [Streaming Principal Component Analysis From Incomplete Data](https://jmlr.org/papers/volume20/16-627/16-627.pdf)
- Maybe this is easier once [SVD](https://www.notion.so/SVD-62d1847b87334fb38590e195fe38889b) is implemented
- [Incremental Singular Value Decomposition of Uncertain Data with Missing Values](https://link.springer.com/content/pdf/10.1007%2F3-540-47969-4_47.pdf)

MarekWadinger:
Hello @MaxHalford, @hoanganhngo610, and everyone  👋,

In https://github.com/online-ml/river/issues/1366, @MaxHalford showed interest in implementation of OnlinePCA and OnlineSVD methods in river. 

Given my current project involvement with online decomposition methods, I believe the community could benefit from having access to these methods and their maintenance over time. Additionally, I am particularly interested in DMD, which combines the advantages of PCA and FFT. Hence, I propose the introduction of three new methods as part of the new decomposition module:

`decomposition.OnlineSVD` implemented based on [Brand, M. (2006)](https://doi.org/10.1016/j.laa.2005.07.021)  (proposed by @MaxHalford in [issue](https://github.com/online-ml/river/issues/1366)) with some considerations on re-orthogonalization. Since it is required quite often, compromising computation speed, it could be interesting to align with [Zhang, Y. (2022)](https://doi.org/10.48550/arXiv.2204.05398) (I made some effort to implement but I'm yet to expore validity and possibility to implement `revert` in similar vein).

`decomposition.OnlinePCA` implemented based on [Eftekhari, A. (2019)](http://jmlr.org/papers/v20/16-627.html) (proposed by @MaxHalford in [issue](https://github.com/online-ml/river/issues/1366)), as it is currently state-of-the-art with all the proofs and guarantees. Would be happy to validate together if all considerations are handled in proposed OnlinePCA.

`decomposition.OnlineDMD` implemented based on [Zhang, H. 2019](https://doi.org/10.1137/18m1192329). It can operate as MiniBatchTransformer, MiniBatchRegressor (sort of), and works with Rolling so I would need some help figuring out how we'd like to classify it (maybe new base class Decomposer.

Additionally, I propose `preprocessing.Hankelizer`, which could be beneficial for various regressors and particularly useful for enhancing feature space by introducing time-delayed embedding.

I've tried to include all necessary tests. However, I need to investigate why re-orthogonalization in OnlineSVD yields significantly different values when tested on various operating systems (locally, all tests pass).

Looking forward for your comments and revisions. 😌


MarekWadinger:
@MaxHalford, thank you for the recognition! Your review would be priceless.

I will gladly provide benchmarks as I am aware that the reliance on numpy is much higher than usual methods in river. Nevertheless, I will also try to decrease it over time as I resolve some higher priority issues.

hoanganhngo610:
First of all, thank you so much @MarekWadinger for taking the initiative to create this PR, and I would say this is both a huge and incredible work.

From my side, I will also try my best to dedicate some time to review the PR, at least making sure that the code quality and layout aligns with what is currently available within `River`.

MarekWadinger:
Hello @MaxHalford and @hoanganhngo610, 👋

I believe the methods are ready for benchmarking. The results are published in [this](https://github.com/online-ml/river/blob/f5352c6c8d6def9b9c6874f25ce8dd64e0d1ebd5/benchmarks/decomposition_methods.ipynb) notebook.

In the plot I combine two checks, performance w.r.t. number of features and delay imposed by conversion from pd.DataFrame (dict) to np.array used in the core.
perf-pd_np-n_features

Mean absolute number of processed samples per second is provided here (for n features in range(3,20) as it remains pretty stable):
- np.array
3102 OnlineDMD
19553  OnlinePCA
631  OnlineSVD (Probably will be completely replaced by Zhang implementation bellow)
3503 OnlineSVDZhang
- pd.DataFrame
1267 OnlineDMD
18012  OnlinePCA
683  OnlineSVD
1718 OnlineSVDZhang

The results in the notebook indicate that using pd.DataFrame slows down OnlinePCA, which is the fastest decomposition implementation, by up to 14%. However, I believe your concerns are likely related to the fact that the core of the decomposition methods works with np.arrays, correct?

What are your thoughts on the performance and adequacy of the evaluation?

Thanks for your time 🙏

MarekWadinger:
Hey @s-bessing.

I used to really like UMAP in my fault detection projects. I believe OnlineDMD could be a match but there are some bottlenecks. It works much better on reasonably noisy data as high autocorrelation may break the underlying SVD computation in case of piece-wise constant behavior (this happens to me, for instance, in OnlineDMDc where I have information about control signal noiseless and does not change for a while). But if there are certain periodic components and dominant patterns in your data, I think you should give it a hit. :)

MaxHalford:
Hello @MarekWadinger. I'm open to merging your contributions, but not like this. What I'd like to do is open one PR for each contribution, and take a thorough look at each method. These methods could be useful, so we need to benchmark them, assess their relevance, and provide usage documentation. Merging this whole PR as is would not provide benefits for users, I believe.

PR #1528, opened 2024-04-19
docs: recipe pipelines

FBruzzesi:
# Description

As discussed in [Misunderstanding in Pipelines Recipe](https://github.com/online-ml/river/discussions/1527), the current documentation seems to be a bit out of sync with the current implementation of the class.

This is an attempt to fix that. I know that documentation and phrasing can be quite opinionated

### DIFF
--- a/docs/recipes/pipelines.ipynb
+++ b/docs/recipes/pipelines.ipynb
@@ -20,7 +20,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 80,
+   "execution_count": 1,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:52.864965Z",
@@ -53,7 +53,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 81,
+   "execution_count": 2,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:53.146332Z",
@@ -81,7 +81,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 82,
+   "execution_count": 3,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:53.156607Z",
@@ -107,7 +107,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 83,
+   "execution_count": 4,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:53.167122Z",
@@ -276,7 +276,7 @@
        ")"
       ]
      },
-     "execution_count": 83,
+     "execution_count": 4,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -296,7 +296,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 84,
+   "execution_count": 5,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:53.181786Z",
@@ -319,7 +319,7 @@
        " 43.75505)"
       ]
      },
-     "execution_count": 84,
+     "execution_count": 5,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -334,7 +334,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 85,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [
     {
@@ -343,7 +343,7 @@
        "28.28960916"
       ]
      },
-     "execution_count": 85,
+     "execution_count": 6,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -371,7 +371,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 86,
+   "execution_count": 7,
    "metadata": {
     "execution": {
      "iopub.execute_input": "2023-12-04T17:49:53.332637Z",
@@ -414,7 +414,7 @@
        " 'you_gov*you_gov': 1.0}"
       ]
      },
-     "execution_count": 86,
+     "execution_count": 7,
      "metadata": {},
      "output_type": "execute_result"
     }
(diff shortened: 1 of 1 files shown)


docs/recipes/pipelines.ipynb
```diff
@@ -13,7 +13,9 @@
    "source": [
     "Pipelines are an integral part of River. We encourage their usage and apply them in many of their examples.\n",
     "\n",
-    "The `compose.Pipeline` contains all the logic for building and applying pipelines. A pipeline is essentially a list of estimators that are applied in sequence. The only requirement is that the first `n - 1` steps be transformers. The last step can be a regressor, a classifier, a clusterer, a transformer, etc. Here is an example:"
+    "The `compose.Pipeline` contains all the logic for building and applying pipelines. A pipeline is essentially a list of estimators that are applied in sequence. The only requirement is that the first `n - 1` steps be transformers. The last step can be a regressor, a classifier, a clusterer, etc.\n",
```
FBruzzesi, line 16: QQ: Does it make sense to have a transformer-like as last step? From my understanding it doesn't as:
- it would not be called in `transform_one`
- it would break any `predict_*` method

docs/recipes/pipelines.ipynb
```diff
@@ -325,35 +327,20 @@
    "source": [
     "from river import datasets\n",
     "\n",
-    "dataset = datasets.TrumpApproval()\n",
-    "x, y = next(iter(dataset))\n",
+    "dataset = iter(datasets.TrumpApproval())\n",
+    "x, y = next(dataset)\n",
```
FBruzzesi, line 331: Otherwise `x, y` would always be the same datapoint

docs/recipes/pipelines.ipynb
```diff
@@ -709,7 +665,109 @@
...
+    "Finally, having your model in a single object means that you can move it around more easily.\n",
+    "\n",
+    "Note that you can include user-defined functions in a pipeline by using a `compose.FuncTransformer`."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Learning during predict\n",
```
FBruzzesi, line 679: Moved the _learning during predict_ in a dedicated section of the guide

docs/recipes/pipelines.ipynb
```diff
@@ -709,7 +665,109 @@
...
+    "\n",
+    "In online macine learning, we can update the unsupervised parts of our model when a sample arrives. We don't _really_ have to wait for the ground truth to arrive in order to update unsupervised estimators that don't depend on it.\n",
```
gbolmier, line 681:
```suggestion
    "In online machine learning, we can update the unsupervised parts of our model when a sample arrives. We don't _really_ have to wait for the ground truth to arrive in order to update unsupervised estimators that don't depend on it.\n",
```

docs/recipes/pipelines.ipynb
```diff
@@ -709,7 +665,109 @@
...
+    "\n",
+    "In other words, in a pipeline, `learn_one` updates the supervised parts, whilst `predict_one` (or `predict_proba_one` for that matter) **can** update the unsupervised parts. \n",
```
gbolmier, line 683:
```suggestion
    "In other words, in a pipeline, `learn_one` updates the supervised parts, whilst `predict_one` (or `predict_proba_one` for that matter) **can** update the unsupervised parts, which often yields better results. \n",
```

If anyone has the motivation, it could actually be nice to also write a new notebook in the examples section of the docs illustrating the potential performance difference from using `learn_during_predict` on a suitable dataset.

Could be a good first issue.

gbolmier (commented):
Thank you for promptly opening this PR and putting your time in it @FBruzzesi 🙌

I answered your comments and left some of my own. Hope it'll help

docs/recipes/pipelines.ipynb
FBruzzesi: I will try to integrate a sort of _comparison_ section

docs/recipes/pipelines.ipynb
```diff
@@ -709,7 +699,282 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 17,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
```
FBruzzesi, line 884: Best way to see this is to spin the docs locally or wherever.
Here a snapshot of how it renders in my local:

image

FBruzzesi: I got to admit, I was not expecting this much of a difference! That's awesome

docs/recipes/pipelines.ipynb
gbolmier: Don't bother too much with that here if it gets complex. The goal is also to have something easily digestible to new users

FBruzzesi: Implemented code to obtain [this result](https://github.com/online-ml/river/pull/1528#discussion_r1573231480).
Not sure if you had something in particular in in mind, but I was curious and tried such experiment.

docs/recipes/pipelines.ipynb
gbolmier: > I got to admit, I was not expecting this much of a difference! That's awesome.

Yeah and you're actually learning from observations directly after predicting them (before seeing the next sample). In real world settings there is usually a lag before the ground truth gets available, which makes you predict for way more samples before you can learn from complete ones. A more realistic benchmark would probably be even more impressive.

> One question I have however is: does the unsupervised part learn from the same datapoint twice? i.e. both during `predict_*` and `learn_*`?

Nope, the context manager sets the [`_LEARN_UNSUPERVISED_DURING_PREDICT` var to `True`](https://github.com/online-ml/river/blob/c7fbedfaba9d68622588b79b23cca20a115e5d2d/river/compose/pipeline.py#L114) so that the unsupervised transformers do not get updated within `learn_*` calls (see [condition](https://github.com/online-ml/river/blob/c7fbedfaba9d68622588b79b23cca20a115e5d2d/river/compose/pipeline.py#L448-L450))

gbolmier (approved):
LGTM! Thank you for the contribution @FBruzzesi 🙌

PR #1533, opened 2024-04-24
Update pyproject.toml

PabloRuizCuevas:
Changed python requirements,

The current setup is correct in a very strict sense, but it enforces the users of the library that uses poetry to have also this type of strict requirements and is unpractical.

For instance I had  python = "^3.11"  in my requirements which is also perfectly correct for my code as python 3.13 is still not in public release and even when is released is still most likely compatible.  

When you do "poetry add river", If you put  python = "^3.11"  then you get:

```
Because no versions of river match >0.21.1,<0.22.0
 and river (0.21.1) requires Python <3.13,>=3.9, river is forbidden.
So, because pytrade depends on river (^0.21.1), version solving failed.
```

Notice that this is the only package in a long list of dependencies that made me change my own toml python version.

I suggest to relax the requirements, as I don't think that preventing future python versions makes much sense, also if I want tomorrow to test the beta test python 3.13 I will get this issue again.

Other topic would be if a current version, already released in python is not working, then I would let the stricter requirements.


MaxHalford:
Hey there. Did you try running `poetry lock` with this change? For me it doesn't work because of SciPy.

Screenshot 2024-04-25 at 00 54 09

pyproject.toml
```diff
@@ -28,7 +28,7 @@ generate-setup-file = true
 script = "build.py"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.13"
+python = "^3.9"
 numpy = "^1.23.0"
 scipy = "^1.8.1"
```
MaxHalford, line 33:
```suggestion
scipy = "^1.12.1"
```

It works for me if you bump SciPy too, and then remove it from `[tool.poetry.group.benchmark.dependencies]`

PabloRuizCuevas: You are right, I made this changes and run poetry lock

MaxHalford:
Perfect, thanks a lot!

PR #1559, opened 2024-06-21
Space Saving, HyperLogLog and Hierarchical Heavy Hitters algorithms

laraabastoss:
Added coded and respective documentation for the Space Saving, HyperLogLog and Hierarchical Heavy Hitters algorithms within the sketch section.

### DIFF
--- a/river/sketch/hierarchical_heavy_hitters.py
+++ b/river/sketch/hierarchical_heavy_hitters.py
@@ -1,0 +1,290 @@
+from __future__ import annotations
+import typing
+
+import math
+
+from river import base
+
+class HierarchicalHeavyHitters(base.Base):
+
+    """Full Ancestry Algorithm implementation for the Hierarchical Heavy Hitters problem.[^1]
+
+    The Hierarchical Heavy Hitters problem involves identifying the most frequent items in a data stream while organizing them hierarchically. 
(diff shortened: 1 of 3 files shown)


smastelini:
Hi @laraabastoss, thanks for your contribution! Recently some errors were fixed in the automated tests, so I am re-running them for this PR. Let's see how that goes and if you need to change something in your code. Perhaps you will need to pull the latest changes from the `main` branch.

Aside from that, I wanted to discuss a scope question. River already has a Heavy Hitters algorithm that is bound to provide the same functionality as Space Saving. I noticed that the current implementation in River supports a fading factor. I do not know the pros and cons of Space Saving vs Lossy Count with Forgetting Factor (the core of River's version), but I think we could do some renaming to keep both versions. 

The idea is to follow the convention we followed so far for the stuff in `river.sketch`:

- We use names that reflect functionality, rather than the actual algorithm name. For example, `Counter`, `Set`, and so on. The algorithm name and related info go in the documentation. So, in your case:
    - Space Saving -> Heavy Hitters (in this case, we will need to find a new name for the current implementation in `sketch.HeavyHitters`, like `FadingHeavyHitters` or something else -- suggestions are welcome) 
    - Hierarchical Heavy Hitters is already conforming to the implicit convention
    - HyperLogLog -> Cardinality? Suggestions are welcome!
- We try as much as possible to inspire in Python's `collections` module for API usage. This brings familiarity to the users and brings name choices tested by time :D. You can check the existing methods in the `sketch` module for inspiration.

PR #1564, opened 2024-07-04
Fix a Rust panic in the tests

e10e3:
Following #1563, PyO3 is upgraded to resolve a crash in the test suite that appeared after Rust 1.78 was released.


gbolmier (approved):
Thank you so much @e10e3! Super helpful 🙏

PR #1566, opened 2024-07-05
Make kNN more resilient to nominal data

### LINKED ISSUE #1565: kNN causes an error when the input has nominal data
## Versions

**River version**: 0.21.1
**Python version**: 3.12.4
**Operating system**: macOS 14.5

## Describe the bug

When a kNN model does a prediction, it computes the distance of its input with previous data points.\
If some of the feature are nominal (*i.e.* not numbers), kNN will cause an error because it tries to perform a subtraction on unsupported data types.

One would expect kNN to be resilient to nominal data, since there exist ways to derive a distance between non-numeric features. A simple one is to give a distance of 0 when they are equal and 1 if they are different.

## Code to reproduce

```python
from river import neighbors

m = neighbors.KNNClassifier()

m.learn_one({"a": "river"}, 0)
m.predict_one({"a": "river"})
```

### Output

```
Traceback (most recent call last):
  File "/path/to/river/knn-strings.py", line 6, in <module>
    m.predict_one({"a": "river"})
  File "/path/to/river/river/base/classifier.py", line 66, in predict_one
    y_pred = self.predict_proba_one(x, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/neighbors/knn_classifier.py", line 149, in predict_proba_one
    nearest = self._nn.search((x, None), n_neighbors=self.n_neighbors, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/neighbors/ann/swinn.py", line 431, in search
    return self._linear_scan(item, n_neighbors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/neighbors/ann/swinn.py", line 339, in _linear_scan
    points = [(p.item, self.dist_func(item, p.item)) for p in self]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/neighbors/base.py", line 33, in __call__
    return self.distance_function(a[0], b[0])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/utils/math.py", line 165, in minkowski_distance
    return sum((abs(a.get(k, 0.0) - b.get(k, 0.0))) ** p for k in {*a.keys(), *b.keys()}) ** (1 / p)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/river/river/utils/math.py", line 165, in <genexpr>
    return sum((abs(a.get(k, 0.0) - b.get(k, 0.0))) ** p for k in {*a.keys(), *b.keys()}) ** (1 / p)
                    ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'str' and 'str'
```

e10e3:
Fixes #1565 by using a different formula for the difference when the data is not numeric.

This allows the distance not to error out when non-numbers are given, while keeping a reasonable behaviour.

### Caveat

Because of the added code paths that are difficult to optimise away, kNN is roughly twice as slow to make a prediction.

### DIFF
--- a/river/utils/math.py
+++ b/river/utils/math.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import functools
 import itertools
 import math
+import numbers
 import operator
 
 import numpy as np
@@ -162,7 +163,23 @@ def minkowski_distance(a: dict, b: dict, p: int):
         Manhattan distance. When `p=2`, this is equivalent to using the Euclidean distance.
 
     """
-    return sum((abs(a.get(k, 0.0) - b.get(k, 0.0))) ** p for k in {*a.keys(), *b.keys()}) ** (1 / p)
+
+    def abs_diff(a, b) -> float:
+        """Naïve absolute difference of two objects.
+
+        If a and b are both numbers, this is the regular absolute difference.
+
+        Otherwise, the difference is 0 is the objects are the same and 1 if they are different.
+        """
+        if isinstance(a, numbers.Real) and isinstance(b, numbers.Real):
+            return float(abs(a - b))
+        elif a == b:
+            return 0.0
+        return 1.0
+
+    return sum((abs_diff(a.get(k, 0.0), b.get(k, 0.0))) ** p for k in {*a.keys(), *b.keys()}) ** (
+        1 / p
+    )
 
 
 def softmax(y_pred: dict):


smastelini:
Please, see my comments on #1565.

PR #1583, opened 2024-07-25
Use units based on powers of 2

### LINKED ISSUE #1582: The wrong units are used for binary size
## Versions

**River version**: 0.21.2

## Describe the bug

There is a confusion in the units used to report the binary size of objects (the number of bytes).

In the library's documentation, decimal units are mentioned (kB, MB, GB, etc.), but the code handling them implicitly expects binary units and converts the quantities to bytes using powers of 2.

Instead, binary units should be used (KiB, MiB, GiB, etc.). This is mostly a documentation change.
There is nothing wrong in using base-2 units, given the proper unit names are used.

### Why does this matter?

River is commonly used for machine learning research, and being precise in one's intent is important. This helps to clear an ambiguity in reporting.

## Steps/code to reproduce

See `humanize_bytes`: the decimal units are used, but the quantity is divided by a power of 1024.

https://github.com/online-ml/river/blob/f18594c253cf10a8110e0060813d879a00620aac/river/utils/pretty.py#L63-L79

e10e3:
Where unit multiples of the byte are used, this PR uses the unambiguous name of the unit.

Binary units have a different symbol (KiB, MiB) to distinguish them from the general-use decimal units (kB, MB). River only uses the binary units.

The only functional change is in `tree.utils.calculate_object_size`, where the recognised valued for the second (optional) argument were changed. This optional argument was not used in River.

Closes #1582

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -1,1 +1,3 @@
 # Unreleased
+
+- The units used in River have been corrected to be based on powers of 2 (KiB, MiB). This only changes the display, the behaviour is unchanged.
--- a/river/forest/adaptive_random_forest.py
+++ b/river/forest/adaptive_random_forest.py
@@ -534,7 +534,7 @@ class ARFClassifier(BaseForest, base.Classifier):
         in the majority class is smaller than this parameter value. This parameter avoids
         performing splits when most of the data belongs to a single class.
     max_size
-        [*Tree parameter*] Maximum memory (MB) consumed by the tree.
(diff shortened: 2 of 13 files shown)


MaxHalford:
Thanks, this is a cool kind of cleanup. Thanks a lot for adding an entry to the release docs!

smastelini:
Oh, I didn't see this PR. I was working in parallel on another PR that actually used MB, GB, and so on. For completeness, as we opted for the binary units, the guideline on hoeffding trees should also be updated! Thanks for the PR @e10e3!

smastelini:
I will take care of the tree guideline!

Issue #1591, opened 2024-08-19
The class CentralMoments from module stats cannot be instantiated


e10e3:
## Versions

**River version**: 0.21.2
**Python version**: 3.12.5
**Operating system**: macOS 14.6.1

## Describe the bug

The class non-abstract [`CentralMoments`](https://github.com/online-ml/river/blob/main/river/stats/moments.py) inherits from `stats.base.Univariate`, but it does not implement any of the abstract methods. As a consequence, creating an instance throws a TypeError.
In addition, the class has no public method apart from the constructor.

There seem to be no reference to this class in River's code.

The last reference to this class in River appears to have been removed during the migration to Rust statistics (commit 5f88f7c80837afb80b37bcb5f57c1149126acfea).

## Steps/code to reproduce

```python
from river.stats.moments import CentralMoments
c = CentralMoments()
```

Output:
```
TypeError: Can't instantiate abstract class CentralMoments without an implementation for abstract methods 'get', 'update'
```

gbolmier:
Good catch @e10e3! Feel free to delete the `stats.moments` module

PR #1593, opened 2024-08-22
Remove the class CentralMoments

e10e3:
This was a utility class internal to River, and the last references to it were removed in a previous version of the library, with the move to Rust statistics.

It inherits from `stats.base.Univariate`, but does not implement any of the abstract methods. As a consequence, creating an instance throws a TypeError.

Closes: #1591


e10e3:
Is it worth adding a changelog entry? It has no public method (not counting the constructor) and was not exported by the `stats`module.
But we are still removing a non-private class.

MaxHalford:
> Is it worth adding a changelog entry? It has no public method (not counting the constructor) and was not exported by the statsmodule.

It doesn't cost much to mention it, so I would do it 👍

MaxHalford (approved):

PR #1594, opened 2024-08-22
Add type stubs for Cython files

e10e3:
Adding type stubs for C extension modules enables type checker to verify the calls made to these modules. They are ignored by the Python interpreter when running the program.
The type stubs here are not all type-complete.

Their addition also allows re-enabling missing imports checks for the `river` package. Without the stubs, checkers like MyPy would complain about not being able to find the C extension modules.

I also fixed the definition of `__pow__` from `vectordict.pyx` to bring it in line with [its signature in Python](https://docs.python.org/3/library/operator.html#operator.__pow__). The `modulo` parameter was unused.


MaxHalford (approved):
This is insane, thanks a lot!

I'm curious to know your background and how you to came to find River :)

PR #1597, opened 2024-08-24
Feature/expected categories

ColdTeapot273K:
Add support for processing only explicitly expected categories for `preprocessing.OneHotEncoder`, `preprocessing.OrdinalEncoder`, akin to `sklearn` api for respective encoders.

All doctests pass (i've added some).

Rationale:
`sklearn` has this neat feature where you can explicitly pass category values you want to see in the encoder state, other values are filtered out. See `categories` parameter: [OneHotEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html), [OrdinalEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html)

This is convenient when you work with high cardinality category spaces where some values are rare and you want to regularize your model. E.g. I've had a practical problem where constraining only to pre-selected top 20% frequent categories in 1 000 000 cardinality space can give you a 10%+ latency boost with no significant loss in metrics, and also make a model lighter on RAM.

This implementation is hackable so if user wants to modify lists of expected categories between training steps, they can do so by direct attribute access. E.g. can glue with modules like [TargetAgg](https://riverml.xyz/latest/api/feature-extraction/TargetAgg/) for some cool dynamic reevaluation of expected category lists. 

P.S. Pls bump Ruff, my LSP config compains coz api changes. Also MyPy complained a lot about about `str | dict | defaultdict` type hints for `category` parameter, I just had to give up on them, maybe someone has better ideas how to  handle them.

### DIFF
--- a/river/preprocessing/one_hot.py
+++ b/river/preprocessing/one_hot.py
@@ -89,15 +89,6 @@ class OneHotEncoder(base.MiniBatchTransformer):
     This is handy when you want to constrain category encoding space
     to e.g. top 20% most popular category values you've picked in advance.
 
-    X = [
-        {
-            'c1': random.choice(alphabet),
-            'c2': random.choice(alphabet),
-        }
-        for _ in range(4)
-    ]
-    pprint(X)
-
     >>> categories = {'c1': {'a', 'h'}, 'c2': {'x', 'e'}}
     >>> oh = preprocessing.OneHotEncoder(categories=categories)
     >>> # oh = preprocessing.OneHotEncoder()
@@ -116,13 +107,6 @@ class OneHotEncoder(base.MiniBatchTransformer):
     ['a', 'h']
     c2
     ['e', 'x']
-
-
-    oh.values.items()
-    [{'c1': {'a', 'h'}, 'c2': {'e', 'x'}}]
-    [{'c1': {'a', 'h'}, 'c2': {'e', 'x'}}]
-
-    {'c1': {'a', 'h', 'i', 'u'}, 'c2': {'d', 'e', 'h', 'x'}}
 
     A subset of the features can be one-hot encoded by piping a `compose.Select` into the
     `OneHotEncoder`.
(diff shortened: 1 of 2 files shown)


MaxHalford:
> P.S. Pls bump Ruff, my LSP config compains coz api changes. Also MyPy complained a lot about about str | dict | defaultdict type hints for category parameter, I just had to give up on them, maybe someone has better ideas how to handle them.

Duly noted, I'll take a look.

river/preprocessing/one_hot.py
```diff
@@ -85,6 +85,29 @@ class OneHotEncoder(base.MiniBatchTransformer):
     {'c2_h': 1}
     {'c2_e': 1}
 
+    Like in `scikit-learn`, you can also specify the expected categories manually.
+    This is handy when you want to constrain category encoding space
+    to e.g. top 20% most popular category values you've picked in advance.
+
+    >>> categories = {'c1': {'a', 'h'}, 'c2': {'x', 'e'}}
+    >>> oh = preprocessing.OneHotEncoder(categories=categories)
+    >>> # oh = preprocessing.OneHotEncoder()
```
MaxHalford, line 94:
```suggestion
```

river/preprocessing/one_hot.py
```diff
@@ -192,23 +215,43 @@ class OneHotEncoder(base.MiniBatchTransformer):
...
+
+    # oh = preprocessing.OneHotEncoder()
+    >>> oh.learn_many(X)
+    >>> df = oh.transform_many(X)
+    >>> df.sort_index(axis="columns")
+       c1_a  c1_h  c2_e  c2_x
+    0     0     0     0     0
+    1     1     0     0     1
+    2     0     0     0     0
     """
```
MaxHalford, line 230:
```suggestion
    2     0     0     0     0
    
    """
```

MaxHalford (commented):
Thanks for this PR, it's useful.

My preference here would be to not adhere to sklearn, and set `categories` to `None` rather than `auto`.

ColdTeapot273K:
> Thanks for this PR, it's useful.
> 
> My preference here would be to not adhere to sklearn, and set `categories` to `None` rather than `auto`.

Deal. I shall modify.

MaxHalford:
@ColdTeapot273K sorry for not replying in a while! I like the changes, we can merge. Before that though, could you add an entry to [`unreleased.md`](https://github.com/online-ml/river/blob/main/docs/releases/unreleased.md)?

ColdTeapot273K:
@MaxHalford no problem, i understand.

Done, please check.

Issue #1417, opened 2023-11-01
No more return self


MaxHalford:
[Suggestion: no more return self · Discussion #935 · online-ml/river](https://github.com/online-ml/river/discussions/935)

MaxHalford:
Yeah we're good. Great job 👏

PR #1601, opened 2024-08-28
Type hints for 'return self'

e10e3:
This is an attempt at #1417.
Some commits have already been done about this, but a few uses were left in the code.

I added type hints for the `learn_one`, `learn_many`, `update`, and  `revert` methods in the base classes, meaning MyPy can now be used to detect child classes with the wrong return type.

I believe the only `update` method with a non-None return is the one in LazySearch, I think the boolean is voluntary and has a use case.

Since the original discussion was about update-like methods in general, I also updated a few methods like `_fit` where it didn't break the logic.


e10e3:
I saw that `SortedWindow.append` has a `return self` as well, but I ultimately decided against changing it in fear of breaking something, even though it could fit the mould.
Please tell me if it should be changed as well.

MaxHalford:
> I saw that SortedWindow.append has a return self as well, but I ultimately decided against changing it in fear of breaking something, even though it could fit the mould.

You can change it. In theory it shouldn't return `self`. We can fix downstream issues if they arise.

MaxHalford (approved):

MaxHalford (approved):

PR #1603, opened 2024-08-29
Add Hack for stats.IQR get method to avoid panicking

### LINKED ISSUE #1602: iqr.get() without values raises an pyo3_runtime.PanicException
## Versions

**river version**: 0.21.2
**Python version**: 3.11
**Operating system**: Windows 10

## Describe the bug
Calling .get() for  stats.IQR raises this error: pyo3_runtime.PanicException: index out of bounds: the len is 0 but the index is 0.
I expect it to behave like stats.Quantile and return None if there is no value available.

It seems like it is fixed for all stats related with quantiles, but was forgotten for IQR.

## Steps/code to reproduce
The following snippet shows the error.
`q.get()` for the Quantile Case does not raise an error, but `iqr.get()` raises an pyo3_runtime.PanicException, which can not be catched with try/except clauses

```python
from river.stats import IQR, Quantile

q = Quantile(0.5)
q.get()

iqr = IQR()
iqr.get()
```

Daveismus:
This fixes #1602 with the same method, as in the other quantlie related stats modules.

### DIFF
--- a/river/stats/iqr.py
+++ b/river/stats/iqr.py
@@ -61,6 +61,10 @@ class IQR(stats.base.Univariate):
             self._is_updated = True
 
     def get(self):
+        # HACK: Avoid crash if get is called before update
+        # panicked at 'index out of bounds: the len is 0 but the index is 0'
+        if not self._is_updated:
+            return None
         return self._iqr.get()
 
     def __repr__(self):


AdilZouitine:
LGTM, Good catch thank you :smile:

AdilZouitine (approved):

PR #1610, opened 2024-09-05
Adding Local Adaptive Streaming Tree

danielnowakassis:
Hi.

I've implemented my decision tree in River and run all the tests for adding a new estimator.

The published paper is here: https://dl.acm.org/doi/10.1145/3605098.3635899

We proposed an adaptive splitting mechanism, where a change detector monitors the error rate or data distribution purity (parameter) of the leaf node to determine the split point (and if merit > 0). It had some good results against SOTA decision trees, and I have also tested it in real datasets in River.

Accuracy :

accuracy

I'm open to any questions. LAST would be error rate monitoring, LAST_D would be data distribution purity monitoring.


river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -130,7 +130,6 @@ class HoeffdingAdaptiveTreeClassifier(HoeffdingTreeClassifier):
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
     Accuracy: 91.49%
```
smastelini, line 132:
```suggestion
    Accuracy: 91.49%

```

river/tree/nodes/last_nodes.py
```diff
@@ -0,0 +1,272 @@
...
+    depth
+        The depth of the node.
+    splitter
+        The numeric attribute observer algorithm used to monitor target statistics
+        and perform split attempts.
+    kwargs
+        Other parameters passed to the learning node.
+    """
+
+    def __init__(self, stats, depth, splitter,change_detector, split_criterion = None, **kwargs):
```
smastelini, line 27:
```suggestion
    def __init__(self, stats, depth, splitter, change_detector, split_criterion = None, **kwargs):
```

river/tree/last_classifier.py
```diff
@@ -0,0 +1,404 @@
...
+        If True, disable poor attributes to reduce memory usage.
+    merit_preprune
+        If True, enable merit-based tree pre-pruning.
+
+    Notes
+    -----
+    Local Adaptive Streaming Tree [^1] (LAST) is an incremental decision tree with
+    adaptive splitting mechanisms. At each leaf, LAST maintains a change detector, 
+    that in case of a change detection in error or the data distribution of the leaf,
+    it performs a split.
```
smastelini, line 74:
```suggestion
    adaptive splitting mechanisms. LAST maintains a change detector at each leaf and splits
    this node if a change is detected in the error or the leaf`s data distribution.
```

river/tree/last_classifier.py
```diff
@@ -0,0 +1,404 @@
...
+    ----------
+
+    [^1]: Daniel Nowak Assis, Jean Paul Barddal, and Fabrício Enembreck.
+    Just Change on Change: Adaptive Splitting Time for Decision Trees in
+    Data Stream Classification . In Proceedings of ACM SAC Conference (SAC’24).
+
+    Examples
+    --------
+
+    >>> from river.datasets import synth
```
smastelini, line 89:
```suggestion

```

river/tree/last_classifier.py
```diff
@@ -0,0 +1,404 @@
...
+        leaf
+            The leaf to evaluate.
+        parent
+            The leaf's parent.
+        parent_branch
+            Parent leaf's branch index.
+        kwargs
+            Other parameters passed to the new branch.
+        """
+        if not leaf.observed_class_distribution_is_pure():  # type: ignore
```
smastelini, line 258: I was wondering. Rather than using the "purity" method you introduced for each split criterion, could something like this be used instead?

smastelini (changes requested):
Hi @danielnowakassis, good stuff! I finished a first pass in the code and left some suggestions/comments :D

I really like the idea of using change detectors to decide when to split, rather than making checks at predefined intervals.

My main suggestion is leveraging existing functionalities via class inheritance. I think that if you make your class inherit from `tree.HoeffdingTreeClassifier` you could remove a lot of repeated code and only focus on the parts that really change from one type of tree to another.

river/tree/split_criterion/variance_ratio_split_criterion.py
```diff
@@ -34,6 +34,9 @@ def merit_of_split(self, pre_split_dist, post_split_dist):
                 vr -= (n_i / n) * (self.compute_var(post_split_dist[i]) / var)
         return vr
 
+    def purity(self, dist):
```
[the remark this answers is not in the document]
smastelini: Before looking into that, please consider my comment about the `leaf.observed_class_distribution_is_pure` method

river/tree/last_classifier.py
danielnowakassis: If we have a general method for getting the merit for a single distribution of a split criterion, we save verifications of which split criterion was chosen. In my opinion, a better name of the method would be get_merit_single_dist(). What do you think?

smastelini (changes requested):
Hi @danielnowakassis, thanks for the changes already in place!

I left some additional comments in your code and noticed some problems with the tests:

- you need to run the pre-commit actions locally. It seems ruff is modifying some formating of the code in the CI, which makes the code quality tests fail
- You will also need to implement `current_merit` for the remaining split criteria, as these classes cannot be currently instantiated with an abstract method (please check the tests logs for more details).

Also, please do not forget to add an entry in the release notes (`docs/releases/unreleased.md`) with your contributions (including the changes in arff files handling)

river/tree/nodes/last_nodes.py
```diff
@@ -0,0 +1,187 @@
...
+    kwargs
+        Other parameters passed to the learning node.
+    """
+
+    def __init__(self, stats, depth, splitter, change_detector, split_criterion=None, **kwargs):
+        super().__init__(stats, depth, splitter, **kwargs)
+        self.change_detector = change_detector
+        self.split_criterion = (
+            split_criterion  # if None, the change detector will have binary inputs
+        )
```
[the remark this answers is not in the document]
smastelini: *Update:* trees actually don't have a split criterion object. `tree.split_criterion` is currently a string with the alias of a valid split criterion. In the future, we might refactor a bit and make this property become the criteria directly. This however would imply a significant deal of refactoring and documentation as the users would need to pass an object rather than a string when instantiating the tree model.

For now, let's keep things as they are to avoid too many changes outside the scope of this PR.

river/tree/last_classifier.py
```diff
@@ -0,0 +1,376 @@
...
+                )
+            elif self._leaf_prediction == self._NAIVE_BAYES:
+                return LeafNaiveBayesWithDetector(
+                    initial_stats, depth, self.splitter, self.change_detector.clone()
+                )
+            else:  # Naives Bayes Adaptive (default)
+                return LeafNaiveBayesAdaptiveWithDetector(
+                    initial_stats, depth, self.splitter, self.change_detector.clone()
+                )
+        else:
```
smastelini, line 176: what about:

```py
split_criterion = self._new_split_criterion() if not self.track_error else None
```

?

docs/releases/unreleased.md
```diff
@@ -17,3 +17,9 @@
 ## tree
 
 - Instead of letting trees grow indefinitely, setting the `max_depth` parameter to `None` will stop the trees from growing when they reach the system recursion limit.
+
+-Added `LASTClassifier` (Local Adaptive Streaming Tree Classifier)
```
smastelini, line 21:
```suggestion
-Added `tree.LASTClassifier` (Local Adaptive Streaming Tree Classifier).
```

river/tree/hoeffding_adaptive_tree_classifier.py
```diff
@@ -129,6 +129,7 @@ class HoeffdingAdaptiveTreeClassifier(HoeffdingTreeClassifier):
     >>> metric = metrics.Accuracy()
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
+
     Accuracy: 91.49%
```
smastelini, line 133:
```suggestion
    >>> evaluate.progressive_val_score(dataset, model, metric)
    Accuracy: 91.49%
```

river/tree/last_classifier.py
```diff
@@ -0,0 +1,361 @@
...
+    ...                        seed=1, position=1500, width=50)
+    >>> dataset = iter(gen.take(3000))
+
+    >>> model = tree.LASTClassifier()
+
+    >>> metric = metrics.Accuracy()
+
+    >>> evaluate.progressive_val_score(dataset, model, metric)
+
+    Accuracy: 91.10%
```
smastelini, line 103:
```suggestion
    >>> evaluate.progressive_val_score(dataset, model, metric)
    Accuracy: 91.10%
```

smastelini (approved):
Thank you so much for the contribution, @danielnowakassis. I really like the idea behind LAST and I am eager to see the next steps of this theoretical framework.

As a note for posterity: @danielnowakassis and I have identified a possible way to improve the tree module via some refactoring, much akin to what was done to the tree splitters a long time ago. The impacts of these changes are yet to be measured. An issue will be open to track this effort.

river/tree/last_classifier.py
```diff
@@ -0,0 +1,363 @@
...
+
+class LASTClassifier(HoeffdingTreeClassifier, base.Classifier):
+    """Local Adaptive Streaming Tree Classifier.
+
+    Local Adaptive Streaming Tree [^1] (LAST) is an incremental decision tree with
+    adaptive splitting mechanisms. LAST maintains a change detector at each leaf and splits
+    this node if a change is detected in the error or the leaf`s data distribution.
+
+    LAST is still not suitable as classifiers of an ensemble due to the change detectors.
+    The authors in [^1] are working in a version of LAST that overcomes this.
```
smastelini, line 25:
```suggestion
    LAST is still not suitable for use as a base classifier in ensembles due to the change detectors.
    The authors in [^1] are working on a version of LAST that overcomes this limitation.
```

danielnowakassis:
> Thank you so much for the contribution, @danielnowakassis. I really like the idea behind LAST and I am eager to see the next steps of this theoretical framework.
> 
> 
> 
> As a note for posterity: @danielnowakassis and I have identified a possible way to improve the tree module via some refactoring, much akin to what was done to the tree splitters a long time ago. The impacts of these changes are yet to be measured. An issue will be open to track this effort.

Thank you, @smastelini. We did a great job :)

PR #1613, opened 2024-09-11
Bugfix for detecting change points based on MOA

denisesato:

### DIFF
--- a/river/drift/adwin_c.pyx
+++ b/river/drift/adwin_c.pyx
@@ -241,7 +241,7 @@ cdef class AdaptiveWindowing:
                         break
                     bucket = self.bucket_deque[idx]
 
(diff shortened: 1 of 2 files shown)


MaxHalford (approved):
I trust you! Can you share a bit more about how you noticed this bug?

denisesato:
I reported the bug on issue Change points reported different from MOA - Method _detect_change #1614

denisesato:
I have updated the failed tests.

smastelini (approved):
Thank you, @denisesato for the latest changes! 

It is interesting to see that changing ADWIN  impacts both positively and negatively some algorithms. In any case, the doc tests use small samples for learning and I don't think they are a comprehensive benchmark.

PR #1628, opened 2024-11-13
Fix tests warnings

e10e3:
River's test suite raises an important quantity of warnings, and their number may hide defects in new code. Most warnings are created by a few pieces of code run many times. This PR aims to greatly reduce the number of warnings in the tests.

Before: 209139 warnings
After : 17 warnings

Most warnings left are caused by libraries like NumPy and Scikit-learn.

Four of the remaining warnings are a deprecation : `pickle`, `copy`, and `deepcopy` support will be removed from `itertools` in Python 3.14. *This may warrant an action.*

This PR also fixes a semi-bug: 
In [`naives_bayes.BernoulliNB.joint_log_likelihood_many()`](https://github.com/online-ml/river/blob/main/river/naive_bayes/bernoulli.py#L266), the `if hasattr(X, "sparse")` path was never used (at least in the tests) because sparse dataframes were completed with non-sparse values, becoming non-sparse.

Thankfully, both types of dataframes have the same behaviour so nothing bad happened and only a warning was issued by Pandas.
In removing the warning, I restored the intended behaviour (or the one I believe was intended).

### DIFF
--- a/river/bandit/bayes_ucb.py
+++ b/river/bandit/bayes_ucb.py
@@ -79,7 +79,7 @@ class BayesUCB(bandit.base.Policy):
         """the p-th quantile of the beta distribution for the arm"""
         p = 1 - 1 / (self._n + 1)
         posterior = self._posteriors[arm_id]
-        return scipy.special.btdtri(posterior.alpha, posterior.beta, p)
+        return scipy.special.betaincinv(posterior.alpha, posterior.beta, p)
 
     def update(self, arm_id, *reward_args, **reward_kwargs):
         """Rewrite update function"""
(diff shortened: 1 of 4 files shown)


MaxHalford (approved):

e10e3:
> Maybe it's worth reactivating the pytest warnings?

Sorry, I don't understand what you mean by this. The warnings are active and displayed at the end of every Pytest run, this is how I realised there were so many.
Is your question related to the CI?

MaxHalford:
No my bad. They're deactivated for me by default, which is why I didn't notice them 👍

PR #1634, opened 2024-11-15
Fix for Issues with CluStream clustering

PhilJahn:
Two fixes regarding issues with the clustering procedure of CluStream:
1. fixes mismatch between centers used in distance calculation and cluster assignment that lead to incorrect assignments
2. ensures kMeans clustering at timegap, rather than allowing for skipping when a data point is added to an existing micro-cluster
A more in-depth description can be found here: [https://github.com/online-ml/river/discussions/1633](https://github.com/online-ml/river/discussions/1633)


e10e3 (approved):
This looks good to me.

Issue #1649, opened 2024-11-24
Failing `release-docs` workflow for release `0.22.0`


gbolmier:
I've been trying to make the new `0.22.0` release today but I've been facing some issues. All the workflows are now running properly except the `release-docs` one.

The docs environment [installs spaCy](https://github.com/online-ml/river/blob/b9861986a6afe3aaea5fcd5fa3a20b9d87e65853/pyproject.toml#L84) for the [sentence-classification nb example](https://riverml.xyz/dev/examples/sentence-classification/), unfortunately spaCy does not support python 3.13 yet (see explosion/spaCy#13658), apparently because of its dependency to srsly (see explosion/srsly#112).

I tried temporarily downgrading the python version in the docs workflows (`dev-docs` and `release-docs`) but the `release-docs` one still fails (see [run](https://github.com/online-ml/river/actions/runs/11992693465/job/33432838755)).

I may not have the time to investigate more tomorrow so if someone want to give it a try by then, please do.
Cc @AdilZouitine, @MaxHalford, @smastelini, @agriyakhetarpal

MaxHalford:
Hey @gbolmier! Thanks for taking a look. Honestly I would say it's ok to drop that last part of the notebook that uses SpaCy. It's not a big deal. We can always revisit and add it back later.

gbolmier:
In the end, didn't have to drop the spaCy part, incrementing poetry cache key in install-env action fixed it. The release should be out in 3-4 hours (time for the [PyPI job](https://github.com/online-ml/river/actions/runs/11997868815) to finish)

MaxHalford:
Great job @gbolmier, thank you very much 🙏

agriyakhetarpal:
Thank you very much indeed, @gbolmier! Great to see [`v0.22.0`](https://github.com/online-ml/river/releases/tag/0.22.0) tagged :D

gbolmier:
Thanks guys! Got some further issues with the pypi workflow but hopefully the [current run](https://github.com/online-ml/river/actions/runs/12014345139) is the right one 🤞

agriyakhetarpal:
I'd recommend splitting off the QEMU-based builds into a separate job or add a matrix, that might be faster for future iterations ;) Also, setting `CIBW_BUILD_VERBOSITY` to `2` or changing the build frontend to `pypa/build`–which is going to become the canonical pathway in future–would display the build steps a bit better.

agriyakhetarpal:
[...]

PR #1653, opened 2024-11-28
Add WASM wheels build

gbolmier:
Let's first test the build of the WASM wheels in an independant workflow so that it is easier to iterate and make sure it works before adding the logic to the pypi workflow.

See #1649 thread for more context.

Current blocker is a conflict between River `pandas<3.0.0,>=2.2.3` requirement and Pyodide pandas version (`2.2.0`) (see [run error](https://github.com/online-ml/river/actions/runs/12062931204/job/33637361981?pr=1653)).

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -97,10 +97,59 @@ jobs:
           name: cibw-sdist
           path: dist/*.tar.gz
 
+  build_pyodide_wheels:
+    name: Build Python ${{ matrix.python-version }} ~ pyodide
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # CPython 3.13 -> Pyodide 0.29.x (pyemscripten_2025_0). Pyodide pins a
+          # Rust nightly there, and the stock Emscripten target std is built
+          # without the emscripten-wasm-eh ABI, so std is rebuilt from source
+          # with -Zbuild-std (needs the rust-src component).
+          - python-version: "313"
+            rust_toolchain: "nightly-2025-02-01"
+            rust_component: "--component rust-src"
+            cargo_build_std: "panic_abort,std"
+            cibw_enable: ""
+          # CPython 3.14 -> Pyodide 314.x (pyemscripten_2026_0). Builds on stable
+          # Rust, where emscripten-wasm-eh is the default and the stock std is
+          # usable as-is, so no -Zbuild-std is needed.
+          - python-version: "314"
+            rust_toolchain: "1.93.0"
+            rust_component: ""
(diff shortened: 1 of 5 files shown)


MaxHalford:
Great stuff @gbolmier!

For the pandas version, I guess it's fine to loosen it on our end.

gbolmier:
> For the pandas version, I guess it's fine to loosen it on our end.

I would prefer too, if possible. @agriyakhetarpal do you remember why you bumped pandas from `^2.1` to `^2.2.3` in #1632?

I just realized that Pyodide is not supporting NumPy `>=2.0` yet (see [supported list](https://pyodide.org/en/stable/usage/packages-in-pyodide.html)). That's why their supported pandas version is `2.2.0` (which is not supporting NumPy `>=2.0` (see [reqs](https://github.com/pandas-dev/pandas/blob/f538741432edf55c6b9fb5d0d496d2dd1d7c2457/pyproject.toml#L33-L35)). 

@agriyakhetarpal does it mean that we need to wait for Pyodide to support newer pandas and NumPy versions in order to test River's WASM build?

agriyakhetarpal:
Hi @gbolmier and @MaxHalford, Pyodide 0.27 has been released on 01/01/2025: https://github.com/pyodide/pyodide/releases/tag/0.27.0

We should be able to go ahead with this PR now 🚀 This does require https://github.com/pypa/cibuildwheel/pull/2117, though – but we may switch to the non-`cibuildwheel` implementation and that would work just fine.

Here's an example workflow if you'd like: https://github.com/scikit-image/scikit-image/blob/107c1a9a2c49f86668c41e8e5a1c101d71ff122f/.github/workflows/emscripten.yml, which just needs the `PYODIDE_VERSION` to be set to 0.27.0 instead of the alpha release.

MaxHalford:
@gbolmier do you have some time to look back into this? Fair if you don't :)

gbolmier:
I'll try to give it a shot before next week

MaxHalford:
There's no such rush, take your time!

agriyakhetarpal:
Hi, y'all! I would suggest waiting a little longer for this, actually. [PEP 783](https://peps.python.org/pep-0783/) was recently accepted, which means that we'll be able to soon publish these wheels to PyPI directly (https://github.com/pypi/warehouse/pull/19804). There is a [new ABI for 2026](https://pyodide.org/en/latest/development/abi.html) (based on Python 3.14), for which we are adding cross-build environments in cibuildwheel. If you could test out https://github.com/pypa/cibuildwheel/pull/2812 here in the meantime, that would be really appreciated!

MaxHalford:
Ok @gbolmier @agriyakhetarpal we're looking good!

PR #1656, opened 2024-11-30
Hoeffding races implementation

slach31:


e10e3 (commented):
Thank you for this PR!
Sorry for the delay before the review.

Your proposition looks good, I have a few questions to better understand the code.

Where does this implementation come from? Is it based on a paper, or maybe on an existing piece of code?
If this is an original work, can you explain its benefits?
If this is based on an existing work, please cite your sources.

PR #1666, opened 2025-01-16
Replace pre-commit in the CI

### LINKED ISSUE #1665: Pre-commit does not report MyPy errors
## Versions

**River version**: 0.22.0
**Python version**: 3.12.8
**Operating system**: macOS 15.2

## Describe the bug
Pre-commit only runs its checks on changed files.
If a configuration was changed in a way that should impact the CI's output but no code file was changed, Pre-commit will not report any error.
(Surprisingly, the option `--all-files` does not help here).

The CI for the most recent commit reports no typing error, but running MyPy manually does.

In this case, here is what I suspect happened:
In commit e069b67, MyPy's version was bumped to 1.13.0 (see [the diff here](https://github.com/online-ml/river/commit/e069b674187331ea9a104686f18f37da93d4af28#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89L2414-R2510)).
This new version of MyPy enabled new checks.
But because some Python files were not touched since, the CI never reported the new errors.

## Steps/code to reproduce

1.  Checkout the main branch (currently at commit 5531ae5)
2. Run `pre-commit run -a`
3. See there are no errors
4. Run `mypy river`
5. 5 errors are reported

### Expected behaviour

Pre-commit should report the MyPy errors, and not pass the MyPy hook.
Alternatively: The CI should report all errors, whether they are in modified files or not.

e10e3:
Because pre-commit only checks edited files, side effects can cause errors to appear in other files and go unnoticed. On the contrary, the CI's role is to guarantee the soundness of the entire codebase; only checking what changed is insufficient.

This PR replaces pre-commit with an explicit call to the linters (in CI only, pre-commit is still available as a hook).

This process discovered previously silent typing errors.
If anything, we lose pre-commit's interface to select file types: the files must now be given as command arguments.

Closes #1665

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -20,5 +20,11 @@ jobs:
           python-version: "3.13"
           build-root: false
 
-      - name: Run pre-commit on all files
-        run: poetry run pre-commit run --all-files
+      - name: MyPy type check
+        run: poetry run mypy
+
+      - name: Ruff code linting
+        run: poetry run ruff check --output-format=github river/
+
+      - name: Ruff code formatting
+        run: poetry run ruff format --check river/
--- a/river/anomaly/pad.py
+++ b/river/anomaly/pad.py
@@ -130,7 +130,7 @@ class PredictiveAnomalyDetection(anomaly.base.SupervisedAnomalyDetector):
             else:
                 self.predictive_model.learn_one(y=y, x=x)
         else:
-            self.predictive_model.learn_one(x=x, y=y)
+            self.predictive_model.learn_one(x=x, y=y)  # type: ignore[union-attr]
 
     def score_one(self, x: dict, y: base.typing.Target):
         # Return the predicted value of x from the predictive model, first by checking whether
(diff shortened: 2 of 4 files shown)


MaxHalford (approved):

e10e3:
@MaxHalford Do you remember why you added kwargs handling in `simulate_qa`?

https://github.com/online-ml/river/blob/5531ae5cfa7028acacc433adb25bd6099f8be4d5/river/stream/qa.py#L145-L146

I'm a bit puzzled by it, no dataset from River yields any additional argument.

MaxHalford:
Hey 👋

I believe it was for progressive validation... But I'd need to check deeper. Do tests pass if you remove it?

e10e3:
I may have found why: some datasets like MovieLens100K have an optional parameter to unpack attributes.

https://github.com/online-ml/river/blob/5531ae5cfa7028acacc433adb25bd6099f8be4d5/river/datasets/movielens100k.py#L52-L56

As you wrote in [an example](https://riverml.xyz/latest/examples/matrix-factorization-for-recommender-systems/part-1/), this mechanism is used by recommender models.

All good then.

e10e3:
It looks like the tests timed out while fetching a remote dataset. I didn't touch that!

MaxHalford:
Indeed, it was for the user_id and item_id fields in the recsys models. Good catch!

e10e3:
There is a maintenance downside to this PR: the CI configuration and the pre-commit file could go out of sync because they are not the same file any more.
It may still be worth it, but it's not mine to decide.

Also, the tests are now green, thank you @smastelini for re-running them!

PR #1668, opened 2025-01-26
Full type hints for the module base

e10e3:
Continuing #1592, this PR adds type hints for the `base`module.

I am not entirely familiar with the code, so we may have things to adjust. Hence, I start with a draft.

Because the PR is quite big, I tried to have clean commits, you can use them to help the review.

Notable changes:
- I changed the regression target to float, because Python's type pyramid recommends not to use the `numbers` module for type annotations. And MyPy doesn't support it anyway.
   Also, `numers.Number` also includes complex numbers, but I am confident saying all the code expects floats. Integers are converted automatically to floats if needed.
- Where transformers are used, the annotation is now `BaseTransformer` to allow all transformers, not just the supervised ones.

~~A few things I would like to discuss before leaving the "draft" status:~~ (These items are addressed now.)
- [x] ~~The Estimator Problem: many functions expect a generic model that has the `predict_*` and `learn_*` methods, but there is no single type representing this capability.~~
   ~~I tagged them as requiring an `Estimator`, but this is not satisfactory.~~
   ~~Should we:~~
    - ~~Enumerate all possible types every time (`Classifier`, `Regressor`, `Clusterer`, `Pipeline`)~~
    - ~~Create a Protocol (a Python interface) for models that can learn, predict, and do both~~
    - ~~Create a new abstract class that introduces `learn_*` and `predict_*`~~
- [X] ~~The `Ensemble` class inherits from `UserList`. Because it is an implementation of the list type, it is mutable, leading it to be type-invariant. Since `Ensemble` is now typed with `Estimator`, all models are bare estimators. An ensemble of Hoeffding trees would not be able to access the HT-specific methods (even things like `learn_one`!) without violating the typing -- runtime behaviour is not impacted, Python is dynamic.~~

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -1,1 +1,7 @@
 # Unreleased
+
+## base
+
+- The `tags` and `more_tags` properties of `base.Estimator` are now both a set of strings.
+- The `base` module is now fully type-annotated. Some type hints have changed, but this does not impact the behaviour of the code. For instance, the regression target is now indicated as a float instead of a Number.
+- `base.Ensemble`, `base.Wrapper`, and `base.WrapperEnsemble` became generic with regard to the type they encapsulate.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,7 +175,6 @@ ignore_missing_imports = true
 [[tool.mypy.overrides]]
 # Disable strict mode for all non fully-typed modules
 module = [
-    "river.base.*",
     "river.metrics.*",
     "river.utils.*",
     "river.stats.*",
--- a/river/anomaly/pad.py
+++ b/river/anomaly/pad.py
@@ -130,7 +130,7 @@ class PredictiveAnomalyDetection(anomaly.base.SupervisedAnomalyDetector):
             else:
                 self.predictive_model.learn_one(y=y, x=x)
         else:
-            self.predictive_model.learn_one(x=x, y=y)  # type: ignore[union-attr]
+            self.predictive_model.learn_one(x=x, y=y)  # type:ignore[attr-defined]
 
     def score_one(self, x: dict, y: base.typing.Target):
         # Return the predicted value of x from the predictive model, first by checking whether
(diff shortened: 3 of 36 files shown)


MaxHalford:
Thanks for the info about `numbers.Number`, makes sense 👍

> The Estimator Problem: many functions expect a generic model that has the predict_* and learn_* methods, but there is no single type representing this capability.

I like the `typing.Protocol` solution, it's modern and probably the right solution here.

> The Ensemble class inherits from UserList (but only partially implements its methods). Because it is an implementation of the list type, it is mutable, leading it to be type-invariant. Since Ensemble is now typed with Estimator, all models are bare estimators. An ensemble of Hoeffding trees would not be able to access the HT-specific methods (even things like learn_one!) without violating the typing -- runtime behaviour is not impacted, Python is dynamic.

We could definitely simplify here: we don't have to inherit from `UserList`. We could instead implement the necessary methods manually. We could also define an option to make the list appendable or not.

e10e3:
> We could definitely simplify here: we don't have to inherit from UserList. We could instead implement the necessary methods manually. We could also define an option to make the list appendable or not.

For the Ensemble thing: after more thought, I think a simple solution would be to make the class generic. This way, we adapt exactly to what type it contains. It is after all an implementation of the `list` type, and `list` is generic.
This way we don't have to rewrite all the methods :)

e10e3:
New version!

I made `base.Ensemble`, `base.Wrapper`, and `base.WrapperEnsemble` generic. This removes a few of the errors I needed to ignore.
They work the same as before, their parameters are still subclasses of Estimator, but now the type checker won't lose track of the actual type of the models they wrap around.

I also created two protocols as discussed above, but I did not find a straightforward use for them yet. One of the problems is the substitution principle; the different kind of models in River all have different parameters (and some have more than others), and I wasn't able to find a type annotation to cover all the cases.
The protocols may still find a use, but that would be for another module. Since they're not used as is, I can remove them.

I also rebased the branch on top of main to fix merge conflicts.

MaxHalford (approved):
Great stuff!

I'll approve and take a deeper look soon (promise!)

PR #1676, opened 2025-04-09
Sync pre-commit's repo versions with Poetry

e10e3:
The tool versions listed for pre-commit were different from the ones in `poetry.lock`, leading to different behaviour between a local invocation and from pre-commit.

In the case of MyPy, pre-commit's version was severely outdated, causing false positives and false negatives.

When River's dependencies are updated, pre-commit's configuration should be updated to have the same versions.

I realise now that this is one of the causes of #1665; but this PR does not solve the main problem of dependencies between files anyway.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.5.7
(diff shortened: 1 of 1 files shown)


gbolmier (approved):

e10e3:
As a note, one of the unit tests failed, but I cannot reproduce the failure on my side. I believe this error is just some flakiness.

Issue #1682, opened 2025-05-13
compose.Select fix type annotation


tauferp:
## Versions

**river version**: 0.21.0
**Python version**: 3.12.4
**Operating system**: Ubuntu 24

## Describe the bug
`compose.Select` init method has weird/likely wrong type annotations:
```
    def __init__(self, *keys: tuple[base.typing.FeatureName]):
```
This makes it seem like every key should be a tuple, when the real use case looks more like:

```
    def __init__(self, *keys: base.typing.FeatureName):
```

## Steps/code to reproduce
Using example from https://riverml.xyz/dev/api/compose/Select/#examples
```python
from river import compose

x = {'a': 42, 'b': 12, 'c': 13}
compose.Select('c').transform_one(x)
```
Running current mypy `1.15.0` yeilds:

```
$ mypy sample.py 
sample.py:4: error: Call to untyped function "transform_one" in typed context  [no-untyped-call]
sample.py:4: error: Argument 1 to "Select" has incompatible type "str"; expected "tuple[Hashable]"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```

Where using the proposed annotations yields:

```
$ mypy sample.py 
sample.py:4: error: Call to untyped function "transform_one" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```

PR #1683, opened 2025-05-13
Improve compose.Select typing (#1682)

tauferp:
A type hint improvement for compose.Select related to #1682


AdilZouitine (commented):
LGTM 😄

PR #1698, opened 2025-07-10
Type hints for the module misc

e10e3:
Continuing on the objective of #1592, I added type hints to the `misc` module.

This module is small and relatively isolated from the remainder of the library, making it easy to type-check.

The most notable thing to note in this PR is the numerous ignores on `deque.maxlen`. SFDT initialises its window with a maximum length, meaning that –provided the types are respected– accessing `maxlen` should never return `None`. The errors become safe to ignore.


smastelini (approved):
LGTM!

Issue #1680, opened 2025-05-03
Possible enhancement at feature selection


caiogimenes:
SelectKBest transformer currently ranks features based on raw Pearson correlation values [-1,1]. So features that are negatively correlated with the target might be excluded, even if they are highly predictive.
e.g:
0.8 → ranked high
−0.9 → ranked low or excluded, despite being stronger

Possible fix is to add a new param `use_abs` to allow user to control using absolute value and adding it to `learn_one` method.

    def learn_one(self, x, y):
        for i, xi in x.items():
            self.similarities[i].update(xi, y)
            if self.use_abs:
                similarity_value = abs(self.similarities[i].get())
            else:
                similarity_value = self.similarities[i].get()
            self.leaderboard[i] = similarity_value

MaxHalford:
Good idea!

PR #1700, opened 2025-07-19
Add use_abs parameter to SelectKBest

caiogimenes:
This PR is related to Issue #1680

### Overview:
- Add `use_abs` parameter to `SelectKBest` for absolute similarity 
- Add docstring test for `use_abs` parameter

### Motivation:
`SelectKBest` transformer currently ranks features based on raw Pearson correlation values [-1,1]. So features that are negatively correlated with the target might be excluded, even if they are highly predictive.
e.g:
0.8 → ranked high
−0.9 → ranked low or excluded, despite being stronger

### Changes:
- A new `use_abs` parameter has been added to the `SelectKBest` constructor. When set to `True`, the selector ranks features based on the absolute value of their similarity scores. (Default is `False` so it won't break client code)
- The leaderboard logic inside `learn_one` method has been modified to respect the `use_abs` value while ranking.
- The docstring for `SelectKBest` has been updated to show the how use_abs change the tranformer's behavior.

### DIFF
--- a/river/feature_selection/k_best.py
+++ b/river/feature_selection/k_best.py
@@ -69,46 +69,44 @@ class SelectKBest(base.SupervisedTransformer):
 
     >>> # Example demonstrating the `use_abs` parameter
 
-    >>> import numpy as np
+    >>> import random
 
-    >>> np.random.seed(42)
-    >>> X_abs = np.random.rand(100, 3)
-    >>> y_abs = (
-    ...     0.6 * X_abs[:, 0]
-    ...     - 0.9 * X_abs[:, 1]
-    ...     + 0.1 * X_abs[:, 2]
-    ...     + np.random.normal(0, 0.1, 100)
-    ... )
+    >>> random.seed(42)
+    >>> X_abs = [[random.random() for _ in range(3)] for _ in range(100)]
+    >>> y_abs = [
+    ...     0.6 * x[0]
+    ...     - 0.9 * x[1]
+    ...     + 0.1 * x[2]
+    ...     + random.gauss(0, 0.1)
+    ...     for x in X_abs
+    ... ]
 
-    >>> # With use_abs=False (the default), feature 0 is selected.
     >>> selector_no_abs = feature_selection.SelectKBest(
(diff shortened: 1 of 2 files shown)


river/feature_selection/k_best.py
```diff
@@ -63,15 +67,55 @@ class SelectKBest(base.SupervisedTransformer):
     >>> selector.transform_one(xi)
     {7: -1.2795, 9: -1.8408}
 
+    >>> # Example demonstrating the `use_abs` parameter
+
+    >>> import numpy as np
+
+    >>> np.random.seed(42)
```
smastelini, line 74: Please, avoid using numpy unless strictly necessary. I suggest using Python's built-in random module for example data generation.

river/feature_selection/k_best.py
```diff
@@ -63,15 +67,55 @@ class SelectKBest(base.SupervisedTransformer):
...
+
+    >>> for xi, yi in stream.iter_array(X_abs, y_abs):
+    ...     selector_with_abs.learn_one(xi, yi)
+
+    >>> selector_with_abs.transform_one({i: v for i, v in enumerate(X_abs[-1])})
+    {1: np.float64(0.6228904758190003)}
     """
 
-    def __init__(self, similarity: stats.base.Bivariate, k=10):
+    def __init__(self, similarity: stats.base.Bivariate, use_abs: bool = False, k=10):
```
smastelini, line 111: Please, change the new parameter position (after k -- the one that will be usually changed by users), as to match the docstring.

smastelini (changes requested):
Thank you so much for the proposal @caiogimenes. I left two minor comments to address. Aside from that, everything looks fine.

Also, please add an entry in the release notes to document the change.

caiogimenes:
Thank you for the feedback @smastelini. Just commited the updated code with the suggested changes.

river/feature_selection/k_best.py
caiogimenes: Done

smastelini (approved):

PR #1702, opened 2025-08-06
Add the update_many method to PearsonCorr.

jdfekete:
For some reason, the `update_many` method was not available on PearsonCorr when it was trivial to do.


MaxHalford:
Thanks! Apologies for the late merge.

PR #1703, opened 2025-08-14
Binary Drift Detector: McDiarmid Drift Detection Method

aditya0by0:
PR for Discussion : https://github.com/online-ml/river/discussions/1672

### **Summary**

Implements three new binary drift detectors based on McDiarmid's inequality:

* **`MDDM_A`** – Arithmetic weighting (linear).
* **`MDDM_E`** – Exponential weighting.
* **`MDDM_G`** – Geometric weighting.


kulbachcedric:
Hi @aditya0by0,
thanks for the transparency regarding the performance and excuse the very late reply ... 
Thats a good question, actually I do not know any case where the McDiarmid Drift detection method was applied in practice. However, I still would like to have it in river as there are already [many papers out there where it has been applied](https://scholar.google.com/scholar?hl=de&as_sdt=0%2C5&q=McDiarmid+Drift+detection+&btnG=).

@MaxHalford what do you think?

MaxHalford:
Yes I think this is worth of being added to River :)

But there are a lot of implementation details that need reworking for this to be acceptable.

Issue #1717, opened 2025-11-05
Division by zero in HoeffdingAdaptivetree with small memory estimation period


jsvobo:
**river version**:
river                     0.22.0 (installed with pip)
**Python version**:
python 3.10
**Operating system**:
Debian

## Describe the bug
When the memory estimation periodis small (can be forced by seting memory_estimate_period=1), and having large ADWIN Boosting classifier number of models (here I had 50), We can randomly run into this error:

```
File ~/<my env>/lib/python3.10/site-packages/river/ensemble/boosting.py:176, in ADWINBoostingClassifier.learn_one(self, x, y, **kwargs)
    174 for i, model in enumerate(self):
    175     for _ in range(utils.random.poisson(1, self._rng)):
--> [176]  model.learn_one(x, y, **kwargs)
    178     if model.predict_one(x) == y:
    179         self.correct_weight[i] += lambda_poisson

File ~/<my env>/lib/python3.10/site-packages/river/tree/hoeffding_adaptive_tree_classifier.py:234, in HoeffdingAdaptiveTreeClassifier.learn_one(self, x, y, w)
    231 self._root.learn_one(x, y, w=w, tree=self)
    233 if self._train_weight_seen_by_model % self.memory_estimate_period == 0:
--> [234]   self._estimate_model_size()

File ~/<my env>/lib/python3.10/site-packages/river/tree/hoeffding_tree.py:300, in HoeffdingTree._estimate_model_size(self)
    298         total_inactive_size += calculate_object_size(leaf)
    299 if total_active_size > 0:
--> [300]  self._active_leaf_size_estimate = total_active_size / self._n_active_leaves
    301 if total_inactive_size > 0:
    302     self._inactive_leaf_size_estimate = total_inactive_size / self._n_inactive_leaves
```

## Code snippet:

tree.HoeffdingAdaptiveTreeClassifier(grace_period=100,max_size=10,max_depth=15,memory_estimate_period=10000)
model = ensemble.ADWINBoostingClassifier(model = tree_Hoeff,n_models=50, seed=seed)

After this init, i trained learn_one() on my data on dataset that i know works with other models.

## Comment:
We should brace this code : "self._active_leaf_size_estimate = total_active_size / self._n_active_leaves" on line 300 in "river/tree/hoeffding_tree.py" by checking is there are any active leaves first.

PR #1721, opened 2025-11-10
Fix division error in inactive leaf size estimation

kulbachcedric:
Hi all,
just added the check for devision errors as mentioned in #1717 

Best
Cedric

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -3,5 +3,8 @@
 ## stats
 
 - Added `update_many` method to `stats.PearsonCorr`.
-- Added handlying for division by zero in `tree.hoeffding_tree` for leaf size estimation.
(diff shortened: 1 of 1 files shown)


kulbachcedric:
Hi @e10e3 
you're right, but are we not eventually facing the same issue in line 301?

e10e3:
> are we not eventually facing the same issue in line 301?

Yes, we are, you're correct.
The code is similar, and the same problem could happen with the next division, so both line 299 and line 301 can benefit from the same fix you propose.

Your first commit was not wrong _per se_, but because you mentioned that you just applied the suggestions from the issue reporter, I wanted to point out that the issue was about a different line, so that you didn't end up patching a different problem while keeping the original one unsolved 😁.

In the end, both lines are affected by the same bug and a complete fix would apply to both.

PR #1723, opened 2025-11-13
Fix the build in CI

e10e3:
This reverts commit 6984c5c0e7663e919d60a5c17e35e4351d00df16.

Commit 6984c5c changed the build system to Setuptools instead of Poetry's own `poetry-core`.
Unfortunately, this change prevents River from being built and breaks CI 🙁.

River, with its C++ and Rust extensions, needs an external compilation step and is currently configured for poetry-core.
Using Setuptools could certainly be done, but it requires more configuration change than swapping the build system 😄.

I propose to restore the ability to build River with Poetry for the time being, and if the transition to Setuptools is wanted, that a follow-up PR can bring a more complete transition.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,10 +6,7 @@ requires = [
     "setuptools>=70.1.0",
     "setuptools-rust",
 ]
-build-backend = "setuptools.build_meta"
-
-[tool.setuptools]
-packages = ["river"]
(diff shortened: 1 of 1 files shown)


MaxHalford:
Yes apologies for having pushed that to main.

I made a separate branch where I'm attempting to use uv. It should make life easier!

PR #1764, opened 2026-03-25
Add ForecastingTrack and change evalulate method

kulbachcedric:
This PR switches 

from river.time_series.evaluate import _iter_with_horizon

to 

from river.evaluate.forecasting import _iter_with_horizon


river/evaluate/forecasting.py
```diff
@@ -4,14 +4,17 @@
...
-        typing.Union[dict, None],  # noqa: UP007
+    tuple[
+        dict | None,
         numbers.Number,
-        typing.Union[typing.List[dict], None],  # noqa: UP006, UP007
-        typing.List[numbers.Number],  # noqa: UP006
+        list[dict] | None,
+        list[numbers.Number],
     ]
 ]
```
e10e3:
```suggestion
TimeSeries = typing.Iterator[
    tuple[
        dict | None,
        float,
        list[dict] | None,
        list[float],
    ]
]
```

The interfaces from the module `number` are not compatible with type-checking, they should be replaced with `float` in this type alias.

river/evaluate/tracks.py
```diff
@@ -110,3 +114,56 @@ def __init__(self):
...
+            dataset=dataset,
+            model=model,
+            metric=self.metric.clone(),
+            horizon=horizon,
+            agg_func=statistics.mean,
+            grace_period=grace_period,
+        )
+
+        for step, (*_, horizon_metric) in enumerate(steps, start=1):
+            if step % checkpoint:
```
e10e3:
```suggestion
            if step % checkpoint != 0:
```
I recommend explicitly writing the check, automatic int-to-bool casts can be misleading.

e10e3 (approved):
Good!
My comments are mostly nits.

I like how you handled the deprecation, with a test for the warning and all.

PR #1799, opened 2026-04-10
Fix progressive_val_score for clustering metrics

### LINKED ISSUE #1616: standard scaler in a pipeline with CluStream returns zero values
**river version**: 0.21.1
**Python version**: 3.12.3
**Operating system**: windows

I dont' understand why I get all values to 0 if i run debug one in pipeline composed by a standard scaler and a CluStream model.

```python
from river import stream
from river import compose
from river import metrics
from river import evaluate

```

MaxHalford:
Alright this is the last PR before we make a new release :)

## Summary

- Fix `progressive_val_score` and `iter_progressive_val_score` to work with `ClusteringMetric` (e.g. `Silhouette`). Previously clustering metrics received `(y_true, y_pred)` instead of `(x, y_pred, centers)`, producing `inf` values.
- Add clustering recipe notebook.

Closes #1616


smastelini:
Do you have an idea why CI/CD is failling?

MaxHalford:
It seems the UCI dataset website is down. I'll wait a bit to give them to fix it, and if not we can move the SMS dataset elsewhere.

MaxHalford:
Looks like it's back.

PR #1804, opened 2026-04-14
Fix Mondrian Tree bounding box loss during branch splits

### LINKED ISSUE #1801: Mondrian Tree branch nodes lose bounding box memory during split (missing copy_all=True)
## Versions
**river version**: 0.23.0
**Python version**: Python 3.13.5
**Operating system**: Windows 11

## Describe the bug
```
    def _split(
        self,
        node: MondrianLeafClassifier | MondrianBranchClassifier,
        split_time: float,
        threshold: float,
        feature: base.typing.FeatureName,
        is_right_extension: bool,
    ) -> MondrianBranchClassifier:
        Split the given node and set the split time, threshold, etc., to the node.

        Parameters
        ----------
        node
            Target node.
        split_time
            Split time of the node in the Mondrian process.
        threshold
            Threshold of acceptance of the node.
        feature
            Feature of the node.
        is_right_extension
            Should we extend the tree in the right or left direction.

        

        new_depth = node.depth + 1

        # To calm down mypy
        left: MondrianLeafClassifier | MondrianBranchClassifier
        right: MondrianLeafClassifier | MondrianBranchClassifier

        # The node is already a branch: we create a new branch above it and move the existing
        # node one level down the tree
        if isinstance(node, MondrianBranchClassifier):
            old_left, old_right = node.children
            if is_right_extension:
                left = MondrianBranchClassifier(
                    node, split_time, new_depth, node.feature, node.threshold
                )
                right = MondrianLeafClassifier(node, split_time, new_depth)
                left.replant(node) # point 1

                old_left.parent = left
                old_right.parent = left

                left.children = (old_left, old_right)
            else:
                right = MondrianBranchClassifier(
                    node, split_time, new_depth, node.feature, node.threshold
                )
                left = MondrianLeafClassifier(node, split_time, new_depth)
                right.replant(node) # point 2

                old_left.parent = right
                old_right.parent = right

                right.children = (old_left, old_right)

            # Update the level of the modified nodes
            new_depth += 1
            old_left.update_depth(new_depth)
            old_right.update_depth(new_depth)

            # Update split info
            node.feature = feature
            node.threshold = threshold
            node.children = (left, right)

            return node

        # We promote the leaf to a branch
        branch = MondrianBranchClassifier(node.parent, node.time, node.depth, feature, threshold)
        left = MondrianLeafClassifier(branch, split_time, new_depth)
        right = MondrianLeafClassifier(branch, split_time, new_depth)
        branch.children = (left, right)

        # Copy properties from the previous leaf
        branch.replant(node, True)

        if is_right_extension:
            left.replant(node)
        else:
            right.replant(node)

        # To avoid leaving garbage behind
        del node

        return branch
```
Because replant() is called without setting copy_all=True (see Point 1 and Point 2 in the code above), the new branch node doesn't copy the range limits from the old node. As a result, this new branch node is left with a completely empty range.

Additionally, when initializing the node, memory_range_min and memory_range_max are generated using collections.defaultdict(int):

```
        self.memory_range_min = collections.defaultdict(int)
        self.memory_range_max = collections.defaultdict(int)
```
Because of how defaultdict(int) works in Python, if a feature key is missing, it automatically evaluates to 0. This causes a hidden mathematical error in range_extension(). When a split occurs, the new node has empty memory dictionaries, meaning it calculates the extension distance between 0 and the new sample, rather than using the actual spatial boundaries of the tree.

I don't know if this behavior is intentional, or if it is just a missing True parameter when calling the replant function.

# Sample code to reproduce the problem
```
import collections

memory_range_min = collections.defaultdict(int)

memory_range_min['feature_1'] = 15.5

print("--- Accessing Keys ---")
print(f"Existing 'feature_1': {memory_range_min['feature_1']}")

print(f"Missing 'feature_2' : {memory_range_min['feature_2']}")
print(f"Missing 'feature_3' : {memory_range_min['feature_3']}")

print("\n--- Final Dictionary State ---")
print(dict(memory_range_min))
--- Accessing Keys ---
Existing 'feature_1': 15.5
Missing 'feature_2' : 0
Missing 'feature_3' : 0
```

MaxHalford:
## Summary
- **Fix missing `copy_all=True`** in `replant()` calls when splitting branch nodes in both `MondrianTreeClassifier` and `MondrianTreeRegressor`. Without this, new child branches lost their `memory_range_min`/`memory_range_max`, causing `range_extension()` to compute against `defaultdict(int)` zeros instead of actual data bounds.
- **Change default `step` from 0.1 to 1.0** for `MondrianTreeClassifier` and `MondrianTreeRegressor`, matching the [onelearn reference implementation](https://github.com/onelearn/onelearn) and the existing `AMFClassifier`/`AMFRegressor` defaults.

Fixes #1801

## Verification

Confirmed the bug by tracing `replant()` calls — branch-to-branch splits were not copying ranges. Cross-referenced with [onelearn's `copy_node`](https://github.com/onelearn/onelearn), which explicitly copies `memory_range_min`/`memory_range_max` in the equivalent code path.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -85,6 +85,8 @@ The `dummy` module is now fully type-annotated.
 
 ## tree
 
+- Fixed Mondrian Tree branch nodes losing bounding box ranges during splits. When a branch was split, the new child branch did not inherit `memory_range_min`/`memory_range_max`, causing incorrect range extension calculations. This affected both `MondrianTreeClassifier` and `MondrianTreeRegressor`, as well as their forest variants (`AMFClassifier`, `AMFRegressor`). Fixes [#1801](https://github.com/online-ml/river/issues/1801).
+- Changed the default `step` parameter for `MondrianTreeClassifier` and `MondrianTreeRegressor` from `0.1` to `1.0`, matching the onelearn reference implementation and the existing default in `AMFClassifier`/`AMFRegressor`.
 - Added `cond_log_proba` to `GaussianSplitter` and optimized `do_naive_bayes_prediction` to use direct log-probabilities, avoiding the `exp`/`log` round-trip.
 - Added handling for division by zero in `tree.hoeffding_tree` for leaf size estimation.
 - Optimized Mondrian trees and AMF (Aggregated Mondrian Forest): replaced dict-based node storage with list-indexed storage, added feature/class-to-index mappings, inlined all hot-path methods, and moved the core traversal loops (`_go_downwards`, `update_downwards`, `predict_proba` upward walk) to Cython. AMFClassifier is **3.2x faster**, AMFRegressor is **2.9x faster**.
--- a/river/forest/aggregated_mondrian_forest.py
+++ b/river/forest/aggregated_mondrian_forest.py
@@ -149,7 +149,7 @@ class AMFClassifier(AMFLearner, base.Classifier):
     >>> metric = metrics.Accuracy()
 
     >>> evaluate.progressive_val_score(dataset, model, metric)
-    Accuracy: 85.37%
+    Accuracy: 84.37%
 
     References
     ----------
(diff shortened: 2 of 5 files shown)


MaxHalford:
Nice bug solved, just in time for the release.

Issue #1803, opened 2026-04-14
Remove rich, ipython, ipykernel dependencies


MaxHalford:
We don't need these really, and they come with too many transitive dependencies that need updating.

MaxHalford:
Done in https://github.com/online-ml/river/pull/1808

PR #1808, opened 2026-04-14
Remove rich, ipython, and ipykernel from dev dependencies Issue #1803

NataliSemi:
Hey 

Removed rich, ipython, and ipykernel from the development dependency groups in pyproject.toml. Regenerated uv.lock. 
Ran the full test suite locally: 3757 passed 44 skipped 54 deselected


MaxHalford:
That was quick! Thank you. Could you remove this code block too? I think it's the only place where rich is used.

https://github.com/online-ml/river/blob/047c72c91a7d281ae220b80a617325c91f69a9b6/.github/workflows/release-docs.yml#L31-L34

NataliSemi:
Good catch, thanks! I've removed that block.

MaxHalford:
Looking good! Thanks a lot.

PR #1814, opened 2026-04-15
Remove matplotlib dependency

### LINKED ISSUE #1802: Remove matplotlib dependency
matplotlib is bloated and has too many transitive dependencies that regularly need updating. Let's use something like Altair instead.

KOTHA-SRIVIBHU:
## Changes
- Removed matplotlib from project dependencies

## Reason
Matplotlib is not used in the core library and introduces unnecessary transitive dependencies.

## Testing
- Project builds successfully
- No new test failures introduced

Closes #1802


MaxHalford:
You can't just remove the dependency. You also need to update the notebooks to use a different viz library.

PR #1819, opened 2026-04-15
Remove matplotlib dependency and replace with Altair in notebooks

### LINKED ISSUE #1802: Remove matplotlib dependency

KOTHA-SRIVIBHU:
## Changes

- Removed matplotlib usage from notebooks
- Replaced all plots with Altair-based visualizations
- Removed matplotlib from dev dependencies
- Added Altair as a project dependency

## Testing

- Verified all modified notebooks run successfully
- Ensured plots render correctly using Altair
- Ran test suite with no new failures

Closes #1802


pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "scipy>=1.16,<2",
     "pandas>=2.2,<3",
     "numpy>=2.3.4,<3",
+    "altair>=5.0.0",
```
MaxHalford, line 13: altair should be a dev dependency

MaxHalford (commented):
Could you provide screenshots of each plot so we can see what they look like?

KOTHA-SRIVIBHU:
> Could you provide screenshots of each plot so we can see what they look like?

Thanks for the feedback, i have added altair as a dev dependency and i have added screenshots of Altair versions here: 
https://drive.google.com/drive/folders/1rDo_-V8Y76d1i8rDP5h4VIXwdF0BzLp-?usp=sharing 
the Altair plots are interactive, which may help with exploration.

One issue I noticed is that in the content-personalization notebook, the CTR of some graphs is showing 0% in the Altair version, whereas the original matplotlib version produces a non-zero CTR. This seems to be related to the simulation or reward logic rather than the plotting itself. 
 
I’m happy to look into this further and apply any suggested fixes.

MaxHalford:
> One issue I noticed is that in the content-personalization notebook, the CTR of some graphs is showing 0% in the Altair version, whereas the original matplotlib version produces a non-zero CTR. This seems to be related to the simulation or reward logic rather than the plotting itself.

Correct, that part is brittle. I will take a look later.

Thanks for the PR, the screenshots look great!

KOTHA-SRIVIBHU:
Thanks for the clarification

Glad the screenshorts look good. if you like me to adjust anything else on my side please let me know.

MaxHalford:
No it's all good, thanks for offering. Some of the charts look weird but it's for other reasons, I'll have to look into more deeply.

Issue #1809, opened 2026-04-14
Remove SpaCy


MaxHalford:
SpaCy is less relevant nowadays, and I feel our SpaCy [example](https://riverml.xyz/latest/examples/sentence-classification/) in the docs is not very useful. I'm keen to remove the example or rework it, and remove the SpaCy dependency.

NataliSemi:
Hi! I'd like to work on this issue if that's okay. It seems similar to one I worked on a few days ago.

MaxHalford:
Sure thing, and thanks!

MaxHalford:
Done in https://github.com/online-ml/river/pull/1829

PR #1829, opened 2026-04-16
Remove SpaCy dependency and related docs usage #1809

NataliSemi:
This PR removes the spaCy dependency and related usage from the project. Issue 1809

Removed spaCy from pyproject.toml
Removed spaCy references from documentation (docs/examples/sentence-classification.ipynb)
Removed installation of en_core_web_sm from the docs release workflow (workflows/release-docs.yml)
Updated lock file to reflect dependency changes


.github/workflows/release-docs.yml
```diff
@@ -24,9 +24,10 @@ jobs:
       - name: Install the project
         run: uv sync --locked --all-extras --dev
 
-      - name: Install dependencies
+      - name: Use Rich in notebooks
```
MaxHalford, line 27: Is this related?

NataliSemi: I think yes. The “Install dependencies” step installs the en_core_web_sm model, and it is used only together with spaCy. I checked the repository with this:

grep -Rin --exclude-dir=.git --exclude-dir=.venv --exclude=uv.lock "en_core_web_sm" .

I couldn’t find any other place where en_core_web_sm is used, so there is probably no use for it. I am happy to add it back if it doesn’t seem like a good idea.

MaxHalford: Ah sorry I got confused, because I opened [another issue](https://github.com/online-ml/river/issues/1803) to remove rich and some other dependencies.

NataliSemi: I see it is similar

MaxHalford:
Very good, thanks!

PR #1830, opened 2026-04-18
stats: add ChiSquared statistic

### LINKED ISSUE #1395: Chi-squared test

KOTHA-SRIVIBHU:
This PR Adds a streaming implementation of the Chi-squared statistic to river.stats.

The class maintains a contingency table between feature values and target classes and computes the Chi-squared value incrementally. This can be useful for measuring the dependency between categorical variables in a streaming setting.

Includes basic and edge case tests to verify correctness.

issue closes: #1395

### DIFF
--- a/river/stats/chi_squared.py
+++ b/river/stats/chi_squared.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+import collections
+import math
+
+from river import stats
+
+
+class ChiSquared(stats.base.Bivariate):
+    """Streaming Chi-squared statistic.
+
+    Maintains a contingency table between feature values (x) and classes (y)
+    and computes the Chi-squared statistic incrementally.
+
+    This can be used to measure the dependency between a categorical feature
+    and a target variable in a streaming setting.
+
+    Examples
+    --------
+    >>> from river import stats
+
+    >>> chi = stats.ChiSquared()
+
+    >>> data = [
+    ...     ("A", 0),
+    ...     ("A", 0),
+    ...     ("B", 1),
+    ...     ("B", 1),
+    ... ]
+
+    >>> for x, y in data:
+    ...     chi.update(x, y)
+
+    >>> round(chi.get(), 3)
+    4.0
+    """
+
+    def __init__(self):
+        # counts[value][class] = frequency
+        self.counts = collections.defaultdict(collections.Counter)
+
+        # total counts per class
+        self.class_totals = collections.Counter()
+
+        # total counts per feature value
+        self.value_totals = collections.Counter()
+
+        # total observations
+        self.n = 0
+
+    @property
+    def name(self):
+        return "chi_squared"
+
+    def update(self, x, y):
+        """Update the statistic with a new observation.
+
+        Parameters
+        ----------
+        x
+            Feature value (categorical).
+        y
+            Target class.
+        """
+        self.counts[x][y] += 1
+        self.class_totals[y] += 1
+        self.value_totals[x] += 1
+        self.n += 1
+
+    def get(self):
+        """Return the current Chi-squared statistic."""
+        if self.n == 0:
+            return 0.0
+
+        chi2 = 0.0
+
+        # Iterate over contingency table
+        for x in self.counts:
+            for y in self.class_totals:
+
+                observed = self.counts[x].get(y, 0)
+
+                expected = (
+                    self.value_totals[x] * self.class_totals[y]
+                ) / self.n
+
+                if expected > 0:
+                    chi2 += (observed - expected) ** 2 / expected
+
+        return chi2
+
+    
--- a/river/stats/chi_squared.py
+++ b/river/stats/chi_squared.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import collections
+import typing
 
 from river import stats, utils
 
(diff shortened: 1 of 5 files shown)


river/stats/chi_squared.py
```diff
@@ -0,0 +1,87 @@
...
+    ...     ("B", 1),
+    ...     ("B", 1),
+    ... ]
+
+    >>> for x, y in data:
+    ...     chi.update(x, y)
+
+    >>> round(chi.get(), 3)
+    4.0
+    """
```
MaxHalford, line 42:
```suggestion
    4.0
    
    """
```

MaxHalford (commented):
This is a good start, but I think you should go a bit deeper:

- It would be nice to have a helper property to do a chi-squared test, and determine if the test is significant. 
- It would maybe be useful to have a rolling version.
- I recommend adding unit tests to guarantee the implementation matches the batch version in SciPy.

Generally speaking I would try to make this a bit more user-friendly. The implementation works, but my guess is that users (or agents!) wouldn't know how to properly use it as such.

KOTHA-SRIVIBHU:
Thanks for the feedback!

I’ve made a few changes based on your suggestions:

- Added p_value, degrees_of_freedom, and is_significant(alpha) to ChiSquared
- Implemented a RollingChiSquared class using revert for efficient updates
- Introduced RollingBivariate in base.py to match the existing rolling pattern
- Updated the tests to properly validate rolling behavior using a windowed (tail-based) approach and handle the initial warm-up steps better
- Added tests comparing against scipy.stats.chi2_contingency

I have made a couple of small fixes in test_stats.py:
- Changed the tail helper to return a list instead of a deque to avoid compatibility issues with NumPy/SciPy
- Added checks for None and NaN before assertions so tests don’t fail during early steps or edge cases

Could you take a look and let me know if anything should be changed?

e10e3:
Since you are adding new classes, could it be possible to add the type annotations for these two classes (at least the function signatures)? It would help a lot ❤️

KOTHA-SRIVIBHU:
> Since you are adding new classes, could it be possible to add the type annotations for these two classes (at least the function signatures)? It would help a lot ❤️

Thanks for the suggestion! I've added type annotations to the function signatures, properties, and internal state for both ChiSquared and RollingChiSquared.

e10e3:
Thank you for the type hints!

In the meantime I sent a change that makes the `utils.Rolling` and co. classes actual generics.
If you rebase, it should allow you to remove the type casts in `RollingChiSquared`.

river/stats/test_stats.py
```diff
@@ -53,7 +53,7 @@ def test_pickling(stat):
 @pytest.mark.parametrize("stat", load_stats(), ids=lambda stat: stat.__class__.__name__)
 def test_pickling_value(stat):
     for i in range(10):
-        if isinstance(stat, stats.base.Bivariate):
+        if isinstance(stat, (stats.base.Bivariate, stats.base.RollingBivariate)):
```
MaxHalford, line 56: I don't think you need this

river/stats/test_stats.py
```diff
@@ -261,9 +286,10 @@ def tail(iterable, n):
     for i, (x, y) in enumerate(zip(X, Y)):
         stat.update(x, y)
         if i >= 1:
-            x_tail = tail(X[: i + 1], n)
-            y_tail = tail(Y[: i + 1], n)
-            assert math.isclose(stat.get(), func(x_tail, y_tail), abs_tol=1e-10)
+            val = stat.get()
+            expected = func(tail(X[: i + 1], n), tail(Y[: i + 1], n))
+            if val is not None and not math.isnan(expected):
+                assert math.isclose(val, expected, abs_tol=1e-10)
```
MaxHalford, line 292: I'm not a fan of changing this. I would instead return a non-nan value in `_chi2_stat`

river/stats/test_chi_squared.py
```diff
@@ -0,0 +1,199 @@
...
+    contingency = []
+    for cat in categories:
+        row = []
+        for cl in classes:
+            row.append(np.sum((x == cat) & (y == cl)))
+        contingency.append(row)
+
+    scipy_chi2, scipy_p, scipy_df, _ = scipy.stats.chi2_contingency(contingency, correction=False)
+
+    assert np.isclose(chi.get(), scipy_chi2)
```
MaxHalford, line 197: It's preferable to use `math.isclose`, and avoid `numpy`

river/stats/chi_squared.py
```diff
@@ -0,0 +1,205 @@
...
+    def revert(self, x: typing.Any, y: typing.Any) -> None:
+        """Revert the statistic with a new observation.
+
+        Parameters
+        ----------
+        x
+            Feature value (categorical).
+        y
+            Target class.
+        """
```
MaxHalford, line 88: `y` is not necessarily the target. Just like for `Cov`, `y` can just be a different variable. We're not necessarily looking to calculate the chi-squared between a feature and a target, so the documentation should be more generic.

MaxHalford:
Hi @KOTHA-SRIVIBHU — thanks for this PR! I've taken your branch, rebased it onto current `main` (post the generic `utils.Rolling` change), and applied my own review feedback on top in #1897. I'll close this in favor of that one. Your commits are preserved with original authorship 🙏

PR #1836, opened 2026-04-22
deprecate neural_net module

### LINKED ISSUE #1828: Deprecating `neural_net` module
### Discussed in https://github.com/online-ml/river/discussions/1824

<div type='discussions-op-text'>

<sup>Originally posted by **MaxHalford** April 15, 2026</sup>
Hey @smastelini! I was taking a look at https://github.com/online-ml/river/issues/1460, and it made me a bit ashamed of the `neural_net` module. It implements `learn_many` and `predict_many`, and `learn_one`/`predict_one` call them under the hood. So it means it's quite slow on single instances. It also doesn't support emerging/disappearing features. Both these properties are not in the spirit of the library. I'd be in favour of removing it and letter other packages take care of neural nets for now (i.e. sklearn)

Also, I'd like to move away from relying on pandas (see https://github.com/online-ml/river/issues/1805) because it's not in the spirit of the library either.

I'm not against reintroducing neural nets down the road, but we'd have to do it well, by handling varying features and writing it Rust to get it as fast as possible on single instances.

Let me know if you agree or don't agree!</div>

kulbachcedric:
This PR deprecates the `neural_net` module.

It can be deleted in the upcoming releases

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## neural_net
+
+- Deprecated `river.neural_net`; importing it now emits a `DeprecationWarning` and users are encouraged to use `deep-river` for neural networks. Addresses [#1828](https://github.com/online-ml/river/issues/1828).
+
 ## tree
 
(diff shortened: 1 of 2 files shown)


MaxHalford:
Good idea to use a deprecation warning 👍

PR #1839, opened 2026-04-27
Add RandomTreesEmbedding feature

### LINKED ISSUE #1386: Random trees embeddings
See [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomTreesEmbedding.html#sklearn.ensemble.RandomTreesEmbedding). This uses random trees, which is identical to `HalfSpaceTrees` in the `anomaly` module. It would be nice to reuse that code instead of duplicating the logic. [The Mondrian Kernel](https://arxiv.org/pdf/1606.05241.pdf) is also worth taking into account.

kulbachcedric:
Hi @MaxHalford,
if I understood the RamdomTreeEmbeddings correctly, this PR should implement them.
The crucial lines are the following, where the transforms_one methods returns a dictionary showing which leaf x ends up in for each tree. 
```
features = {}
for tree_id, tree in enumerate(self.trees):
    leaf = tree if isinstance(tree, HSTLeaf) else tree.traverse(x)
    features[(tree_id, self._leaf_indices[tree_id][id(leaf)])] = 1.0
return features 
```

Could you please briefly check it ?

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## feature_extraction
+
+- Added `feature_extraction.RandomTreesEmbedding`, an online random-tree leaf embedding transformer for feeding sparse tree features into downstream models. Addresses [#1386](https://github.com/online-ml/river/issues/1386).
+
 ## neural_net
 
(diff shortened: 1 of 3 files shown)


MaxHalford (commented):
Nice! The one thing that's missing is that the implementation assumes features are scaled between [0, 1]. We need to document in every place where this assumption is made that the user should use a `MinMaxScaler` prior. See what I mean?

PR #1842, opened 2026-04-29
1476 hoeffding races

kulbachcedric:
Added Hoeffding Race Classifier and Regressor

### DIFF
--- a/river/model_selection/hoeffding_race.py
+++ b/river/model_selection/hoeffding_race.py
@@ -1,0 +1,312 @@
+from __future__ import annotations
+
+import abc
+import math
+
+from river import metrics
+from river.model_selection.base import ModelSelectionClassifier, ModelSelectionRegressor
+
+
+class HoeffdingRace(abc.ABC):
+    r"""Hoeffding Race model selection.
+
+    Uses the Hoeffding bound to progressively eliminate underperforming models with
+    statistical confidence. Each model is evaluated on every observation. After each
+    observation, models whose lower bound on error exceeds the upper bound of the best
+    model are eliminated from the race.
+
+    The Hoeffding bound guarantees that the estimated mean of `n` bounded observations
+    is within epsilon of the true mean with confidence `1 - delta`:
+
+    $$\epsilon = \sqrt{\frac{B^2 (\log(2nm) - \log(\Delta))}{n}}$$
+
+    where `B` is the range of the loss, `n` is the number of observations, `m` is the
+    number of models, and `Delta` is the overall confidence parameter.
+
+    Parameters
+    ----------
+    models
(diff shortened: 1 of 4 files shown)


river/model_selection/hoeffding_race.py
```diff
@@ -0,0 +1,313 @@
...
+        The algorithm is correct with probability at least `1 - delta`.
+    loss_range
+        The range of the loss values (max - min). For metrics bounded in [0, 1] like
+        accuracy or zero-one loss, this is 1.0. For unbounded metrics like MAE or MSE,
+        you should set this to a reasonable upper bound on the per-observation loss range.
+
+    """
+
+    models: list[base.Estimator]
+    metric: metrics.base.Metric
```
MaxHalford: Are you confident here? It's not really a pattern we use elsewhere in the codebase, right? I think attributes you declare this way are shared by all instances.

I would be more inclined to simply instantiate a `self._models` and `self._metric` in the `__init__`.

river/model_selection/hoeffding_race.py
```diff
@@ -0,0 +1,313 @@
...
+                    if val - eps > best_val + eps:
+                        continue  # eliminated
+                surviving.append(i)
+
+            self._active = surviving
+
+    @property
+    def n_active_models(self) -> int:
+        """The number of models still in the race."""
+        return len(self._active)
```
MaxHalford: Feels a bit overkill

river/model_selection/sh.py
```diff
@@ -405,5 +405,6 @@ def _pred_func(self, model):
     def predict_proba_one(self, x):
         return self.best_model.predict_proba_one(x)
 
+    @property
```
MaxHalford, line 408: I'm surprised this doesn't break anything

MaxHalford (approved):
Nice! Approved with some comments. Also, don't forget to add an `unreleased.md` entry.

I'd love to benchmark this against the other model selection methods.

river/model_selection/hoeffding_race.py
kulbachcedric: You are right...  the ruff checks failed and I just added it, without thinking about it ...

kulbachcedric:
> Nice! Approved with some comments. Also, don't forget to add an `unreleased.md` entry.
> 
> I'd love to benchmark this against the other model selection methods.

Hi @MaxHalford,
thanks for your review! 
I think it would be better to add an example (or leave the docs as they are) as the algorithm itself is a model selection algorithm and we would need to precise a list of base models to compare all model selection algorithms.

MaxHalford:
Yes I agree, my comment was just a passing thought

PR #1852, opened 2026-05-07
Port AdaptiveWindowing (ADWIN) from Cython to Rust

MaxHalford:
**Stacked on #1851** (port/rolling-auc-to-rust). Once that lands, the merge base will move to `main`.

## Summary
- Reimplements the inner `AdaptiveWindowing` class behind `drift.ADWIN` in Rust under `rust_src/adwin.rs` and exposes it as `river.stats._rust_stats.AdaptiveWindowing`. The Cython sources (`adwin_c.pyx`, `adwin_c.pyi`) are removed; `river/drift/adwin_c.py` becomes a one-line re-export shim so existing imports keep working.
- The Rust port is a near-line port of the original. f64 arithmetic is ordered identically to the Cython, so the streaming Welford-style variance update lines up bit-for-bit.

## Parity
Output is bit-identical to the previous Cython implementation across a 3.8k-step parity fuzz (canonical drift test, constant stream, alternating values, Gaussian noise, custom parameters). Width, total, variance, `n_detections`, and the `drift_detected` flag all match exactly: `mismatches=0 worst_float_diff=0.0`.

## Benchmarks (200k updates)

| config | Cython | Rust | speedup |
|---|---|---|---|
| default (clock=32) | 62 ns/call | 46 ns/call | **1.33×** |
| clock=1 | 644 ns/call | 186 ns/call | **3.46×** |
| max_buckets=10 | 79 ns/call | 49 ns/call | **1.59×** |

`clock=1` is the worst case for the previous implementation because the change-detection loop runs on every update; that's where the Rust port pulls ahead the most.

## Test plan
- [x] `uv run pytest river/drift/test_adwin_regression.py` — 27/27 pass (incl. canonical "drift detected at index 1023")
- [x] `uv run pytest -n auto` — 3773 passed, 42 expected skips, 0 failures
- [x] `cargo test --no-default-features --lib adwin` — 5 Rust unit tests pass
- [x] parity fuzz vs. captured Cython baseline (5 runs, 3.8k steps) — `mismatches=0 worst_float_diff=0.0`


rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+                        if n0 > 0.0 {
+                            let mu0 = u0 / n0;
+                            v0 += self.bucket_list[idx].variance_at(k)
+                                + n0 * n2 * (mu0 - mu2) * (mu0 - mu2) / (n0 + n2);
+                        }
+                        if n1 > 0.0 {
+                            let mu1 = u1 / n1;
+                            v1 -= self.bucket_list[idx].variance_at(k)
+                                + n1 * n2 * (mu1 - mu2) * (mu1 - mu2) / (n1 + n2);
+                        }
```
AdilZouitine: v0 and v1 are accumulated on every inner-loop iteration but never read.

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+                            change_detected = true;
+                            if self.width > 0.0 {
+                                let _ = n0 - self.delete_element();
+                                break 'outer;
+                            }
+                        }
+                    }
+                }
+
+                let _ = (v0, v1); // accumulators are scratch values
```
AdilZouitine: This  a warning-suppression workaround rather than a fix? Is worth either removing.

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+
+                        let delta_mean = (u0 / n0) - (u1 / n1);
+                        if n1 >= self.min_window_length
+                            && n0 >= self.min_window_length
+                            && self.evaluate_cut(n0, n1, delta_mean, self.delta)
+                        {
+                            reduce_width = true;
+                            change_detected = true;
+                            if self.width > 0.0 {
+                                let _ = n0 - self.delete_element();
```
AdilZouitine: This a warning-suppression workaround rather than a fix? Is worth either removing.

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+        self.total -= u;
+        let mu_window = self.total / self.width;
+        let inc = v + n * self.width * (mu - mu_window) * (mu - mu_window) / (n + self.width);
+        self.variance -= inc;
+
+        self.bucket_list[last].remove();
+        self.n_buckets -= 1;
+
+        // Pop the empty trailing bucket so subsequent iterations don't see it.
+        if idx_after - 1 == 0 {
```
AdilZouitine: `idx_after` is captured before `bucket.remove()`, and then the pop condition is `idx_after - 1 == 0`. It forces the reader to mentally undo `remove()`. It also relies on `idx_after >= 1`; if that ever breaks, `idx_after - 1` underflows in debug. We re-borrowing after remove():
```rust
self.bucket_list[last].remove();
self.n_buckets -= 1;
if self.bucket_list[last].current_idx == 0 {
    self.bucket_list.pop();
}
```

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AdaptiveWindowing {
+    delta: f64,
+    clock: i32,
+    max_buckets: usize,
+    min_window_length: f64,
+    grace_period: f64,
```
AdilZouitine, line 84: `min_window_length/grace_period` stored as f64 even though they're conceptually ints. Saves casts inside the hot loop, but a comment explaining the choice would help

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        self.total_width += self.width;
+        if change_detected {
+            self.n_detections += 1;
+        }
+        change_detected
+    }
+
+    fn evaluate_cut(&self, n0: f64, n1: f64, delta_mean: f64, delta: f64) -> bool {
+        let delta_prime = (2.0 * (self.width).ln() / delta).ln();
```
AdilZouitine: (self.width).ln() -> self.width.ln()

rust_src/adwin.rs
```diff
@@ -0,0 +1,392 @@
...
+
+            // The next bucket is not yet full, so we're done.
+            if self.bucket_list[idx + 1].current_idx <= self.max_buckets {
+                break;
+            }
+
+            idx += 1;
+            if idx >= self.bucket_list.len() {
+                break;
+            }
```
AdilZouitine: by the time we reach idx += 1, the push above guarantees `idx + 1 < self.bucket_list.len()`, and nothing else changes the length. The post-increment if idx >= self.bucket_list.len() { break } is unreachable. 
you can drop it no?

AdilZouitine (commented):
Thanks it's nice contribution and we can merge it, I have fews comments, mostly dead code.
We can add also serialization round-trip test.` __getstate__ / __setstate__ / __getnewargs__` are wired up but nothing exercises them. 

One thing worth flagging on layout: `AdaptiveWindowing` ends up at `river.stats._rust_stats`, but `ADWIN` is a drift detector,. And it's not just this PR; `metrics/rolling_pr_auc`, `metrics/rolling_roc_auc`, `metrics/expected_mutual_info` already reach into `river.stats._rust_stats` too. 

I don't think we need to solve this in this PR, but as a follow-up. 
Rename to river._rust (or river._native) and have each Python submodule re-export what it needs.

A bigger question: once the C/C++/Cython -> Rust port is done, we can drop setuptools + Cython.Build + setuptools_rust and move to maturin. Cleaner toolchain, simpler `pyproject.toml`, no `setup.py`. Worth doing in one shot at the end rather than mid-port otherwise we maintain two build paths.

river/drift/test_adwin_regression.py
```diff
@@ -307,3 +309,56 @@ def test_reset_on_drift_detected(self):
         adwin.update(50.0)
         # After reset, width should be 1 (fresh start)
         assert adwin.width == 1
+
+
+# ---- Serialization ----
+
+
+class TestSerialization:
```
AdilZouitine: The test is a class which non consistent with our function based tests

PR #1855, opened 2026-05-08
Port Mondrian tree helpers from Cython to Rust

MaxHalford:
## Summary
- Reimplements `tree.mondrian._mondrian_ops` in Rust at `rust_src/mondrian.rs` and exposes it via `river.stats._rust_stats`. The Cython source (`_mondrian_ops.pyx`) is removed; `mondrian_tree_classifier.py`, `mondrian_tree_regressor.py`, and `mondrian_tree_nodes.py` import the Rust entry points instead.
- Three structural changes shrink per-call cost beyond a line-by-line port:
  - **`_go_upwards` is a single Rust call.** The leaf-to-root walk that recomputes `log_weight_tree` no longer sets up N Python frames per `learn_one` (~360k frames eliminated on a 20k-sample run).
  - **Predict descends + aggregates entirely in Rust.** `predict_proba_one` (classifier) and `predict_one` (regressor) make a single FFI call (`predict_proba_classifier` / `predict_one_regressor`) that walks root → leaf and aggregates upward, removing the Python loop.
  - **Lazy `extensions` dict + hoisted range bounds in `go_downwards_*`.** Most loop iterations descend without splitting; we now compute only the sum (no dict allocation) on the common path, and `range_min` / `range_max` are fetched once per iteration and reused by `update_downwards_*`.
- The existing `stats.Mean` abstraction is preserved end-to-end — the regressor still goes through `_mean.update()` / `_mean.get()` rather than poking at internal attributes.
- Output matches the Cython baseline exactly (Bananas docstring example unchanged at `Accuracy: 70.64%`).

## Benchmarks vs the previous Cython implementation

20k learn+predict iterations on a 10-feature synthetic stream with `MinMaxScaler` upstream:

| model | Cython | Rust | speedup |
|---|---|---|---|
| `MondrianTreeClassifier` | ~32 µs/iter | ~23 µs/iter | **1.28×** |
| `MondrianTreeRegressor`  | ~39 µs/iter | ~31 µs/iter | **1.21×** |

Profiling confirms the gains come from the structural changes above: `_go_upwards` Python time drops from ~73 ms to ~40 ms (single Rust call), the Python predict loops disappear, and `go_downwards_regressor` shaves ~5–6 µs/call by skipping the dict allocation on non-splitting iterations.

## Test plan
- [x] `uv run pytest river/tree/mondrian river/forest river/test_estimators.py -k 'Mondrian or AMF or mondrian or amf'` — **58 passed**
- [x] `uv run mypy river/tree/mondrian` — passes
- [x] Bananas docstring example matches the documented `Accuracy: 70.64%`
- [x] Spot-check: `MondrianTreeRegressor` predictions on a 100-step synthetic stream identical to Cython baseline (sequence + numerical values)
- [x] AMF (`forest.aggregated_mondrian_forest`) tests green — confirms downstream ensemble behaviour is preserved


rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+        let max_v = range_max
+            .get_item(&k)?
+            .ok_or_else(|| PyValueError::new_err("range_max missing key"))?;
+        let mx = as_f64(&max_v)?;
+        if xf > mx {
+            sum += xf - mx;
+        }
+    }
+    Ok(sum)
+}
```
AdilZouitine: The two functions share the entire arithmetic core; only the dict allocation differs. Either:
- Have inner accept an `Option<&PyDict>` for the optional output, or
- Have inner always allocate but cheaply (e.g. with capacity 0) and let sum ignore it.

river/stats/_rust_stats.pyi
```diff
@@ -60,3 +60,49 @@ class RsRollingPRAUC:
     def get(self) -> float: ...
 
 def expected_mutual_info(n_samples: float, a: list[int], b: list[int]) -> float: ...
+
+# Mondrian tree helpers (return values are loose `object`-typed to match the
+# original Cython entry points which Pyright treated as Unknown).
```
AdilZouitine: The real fix is a `Protocol` (or two: `MondrianClassifierNode`, `MondrianRegressorNode`) describing the attributes the Rust code reads `(counts, n_samples, time, parent, children, _mean, weight, log_weight_tree)`. Same goes for dict -> `dict[Hashable, float]`. The signatures are otherwise unchecked.

rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+            )?;
+            let is_leaf: bool = current.getattr(is_leaf_attr)?.extract()?;
+            if is_leaf {
+                return Ok((current.unbind(), new_root, nodes_added));
+            }
+            let feature = current.getattr(feature_attr)?;
+            unsafe {
+                let res = ffi::PyDict_Contains(x.as_ptr(), feature.as_ptr());
+                if res < 0 {
+                    return Err(PyErr::fetch(py));
```
AdilZouitine: PyO3's safe x.contains(&feature)? does the same thing. If this is a measured hot-path optimization, say so in a comment; otherwise drop the unsafe and the raw FFI.

rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+                }
```
AdilZouitine: PyO3's safe x.contains(&feature)? does the same thing. If this is a measured hot-path optimization, say so in a comment; otherwise drop the unsafe and the raw FFI.

rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+    loop {
+        if current.getattr(is_leaf_attr)?.extract::<bool>()? {
+            return Ok(current);
+        }
+        let feature = current.getattr(feature_attr)?;
+        unsafe {
+            let res = ffi::PyDict_Contains(x.as_ptr(), feature.as_ptr());
+            if res < 0 {
+                return Err(PyErr::fetch(py));
+            }
```
AdilZouitine: PyO3's safe x.contains(&feature)? does the same thing. If this is a measured hot-path optimization, say so in a comment; otherwise drop the unsafe and the raw FFI.

rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+                    let children = current.getattr(children_attr)?;
+                    let left_child = children.get_item(0)?;
+                    let child_time: f64 = left_child.getattr(time_attr)?.extract()?;
+                    if candidate < child_time {
+                        split_time = candidate;
+                    }
+                }
+            }
+            range_bounds = Some((range_min, range_max));
+        }
```
AdilZouitine, line 449: the classifier loop fetches `range_min/range_max` lazily (only when `do_split_check`), the regressor look them at the top of every iteration. the classifier pattern is the optimization the file header comment describes. the regressor should match.

rust_src/mondrian.rs
```diff
@@ -0,0 +1,1042 @@
...
+            .get_item(&k)?
+            .ok_or_else(|| PyValueError::new_err("extension key missing"))?;
+        weights.append(v)?;
+    }
+
+    let kwargs = PyDict::new(py);
+    kwargs.set_item(intern!(py, "k"), 1)?;
+    let chosen = rng_choices.call((sorted_keys_list, weights), Some(&kwargs))?;
+    let item = chosen.get_item(0)?;
+    Ok(item)
```
AdilZouitine: weighted_choice re-imports `builtins.sorted` and allocates a Python list every call.
We can do it in rust `(Vec<(Bound<PyAny>, f64)> + sort_by on the f64)`, then build the keys/weights once. That also avoids the per-iteration import `builtins / getattr("sorted")` overhead.

AdilZouitine (commented):
Nice contribution I have few comments

PR #1867, opened 2026-05-12
refactor(rust)!: move native extension to river._river_rust with PyO3 submodules

AdilZouitine:
## What & Why

Reorganises the native extension to match its actual semantics. Today everything lives under `river.stats._rust_stats` even though large parts (`VectorDict`, `AdaptiveWindowing`, the Mondrian helpers) have nothing to do with stats. This PR creates one extension at `river._river_rust` with four PyO3 submodules: `stats`, `drift`, `tree`, `vectordict` — each containing only what belongs there. Python import sites are repointed accordingly.

This is a **breaking change for pickles** generated with prior versions: every `#[pyclass(module = …)]` string changes, and pickles bake those paths in. A conversion script ships in a follow-up PR in the stack.

## How

1. `pyproject.toml`: `module-name = "river._river_rust"` (the only `[tool.maturin]` change). Add `river._river_rust` and `river._river_rust.*` to mypy's `ignore_missing_imports` list (the old `.pyi` no longer applies; see below).
2. `rust_src/pyo3_bindings.rs`:
   - Rename `#[pymodule] fn _rust_stats` → `_river_rust`.
   - Add a `register_submodule` helper that builds a child `PyModule`, pins its `__name__` to the full dotted path, invokes the per-submodule builder, registers the child in `sys.modules`, then attaches it to the parent.
   - Add four registrators (`register_stats`, `register_drift`, `register_tree`, `register_vectordict`); each receives exactly the classes / functions listed in the inventory.
   - Update all 12 `#[pyclass(module = ...)]` strings (11 to `"river._river_rust.stats"`, 1 to `"river._river_rust.drift"` for `AdaptiveWindowing`).
3. `rust_src/vectordict.rs`: update the `VectorDict` `#[pyclass(module = ...)]` to `"river._river_rust.vectordict"`.
4. Python import sites:
   - Pattern A (`from river.stats import _rust_stats`, used in 7 stats files + `expected_mutual_info.py`) → `from river._river_rust import stats as _rust_stats`. Body references unchanged.
   - Pattern B (`from river.stats._rust_stats import X`) repointed per submodule: `river._river_rust.stats` for the two metrics files, `river._river_rust.drift` for `drift/adwin.py` + its regression test, `river._river_rust.tree` for the three Mondrian files (multiple blocks each), `river._river_rust.vectordict` for `utils/vectordict.py`.
5. Drop the orphan `river/stats/_rust_stats.pyi`. It described an import path that no longer exists, and Python's import system would not let us re-shelve a stubs-package directory next to the new `.so` without shadowing the extension. Type stubs for the new submodules can be added in a follow-up. One downstream `# type: ignore[arg-type]` in `river/utils/math.py` (the `_euclidean_distance` call site) is retuned to `[no-any-return]` to match the new untyped signature.
6. Add `river/utils/test_native_submodules.py`: a minimal regression test that asserts each of the four submodules resolves via `importlib.import_module("river._river_rust.<name>")`. The `sys.modules` insertion in `register_submodule` is silent: dropping it in a future refactor would still let `import river._river_rust.stats` succeed (because of `add_submodule`), but break `from river._river_rust.stats import X` in scripts that haven't first touched `river._river_rust`. This test catches that.

The child PyModule's `__name__` must be set to the full dotted path **before** the registrators run; otherwise `m.add_function(...)` captures the short name and pyfunctions end up with `__module__ = "tree"` instead of `"river._river_rust.tree"`, which then diverges from `#[pyclass]` classes (whose module string is fixed via the attribute).

## Testing

- [x] All four submodules import cleanly: `from river._river_rust.{stats,drift,tree,vectordict} import …`
- [x] Negative test: `import sys; sys.modules.pop("river._river_rust", None); from river._river_rust.stats import RsQuantile` still works (the dotted-path import works without the parent being touched first).
- [x] `__module__` strings: every PyO3 class and every pyfunction reports the correct `river._river_rust.<sub>` value.
- [x] `river/utils/test_native_submodules.py` passes.
- [x] Full pytest suite: `uv run pytest -n auto` → 3793 passed.
- [x] Pickle round-trips on the new layout for `Quantile`, `ADWIN`, `VectorDict` (legacy pickles need the migration script that lands in the next PR in the stack).
- [x] Wheel layout: `unzip -l dist/river-*.whl | grep '\\.so$'` shows exactly one match at `river/_river_rust.<abi>.so`; the old `_rust_stats` name is gone.

## Notes for reviewers

- **Diff is large but ~70% mechanical import rewrites.** Suggested review order: `pyproject.toml`/`Cargo.toml`, then `rust_src/pyo3_bindings.rs` (the real work), then `rust_src/vectordict.rs` (one-line attr change), then skim the Python import churn.
- **The `sys.modules` insertion in `register_submodule` is non-optional.** PyO3 doesn't do it automatically. The new `test_native_submodules.py` regression test exists specifically to catch a refactor that drops it.
- **Pinning the child module's `__name__` before populating it** is required to get `pyfunction.__module__` right. Without that step, classes and functions report inconsistent module strings, which breaks pickles for any function-style pyclass member that pickle introspects.
- The pre-existing mismatch in `vectordict.rs` (`#[pyclass(module = "river.utils.vectordict")]` declared but actually accessed via `river.stats._rust_stats`) is corrected: the class genuinely lives at `river._river_rust.vectordict` now. That's why the upcoming pickle migration script has an entry for the legacy `river.utils.vectordict` path.
- **No backwards-compat shim.** A re-export wrapper at `river/stats/_rust_stats.py` was considered and rejected: it would mask the rename in user stack traces and make it harder to find leftover internal imports. Clean break, with a migration script.

BREAKING CHANGE: pickles produced by previous river releases bake the old module strings and no longer load. A migration script lands in the next PR in the stack; see `unreleased.md` for the user-facing note.

---

## Stack

- #1865, PR 1: Cleanup dead Cython/setuptools artifacts
- #1866, PR 2: Switch build backend setuptools-rust to maturin
- **#1867, PR 3: Move extension to `river._river_rust` with PyO3 submodules &larr; you are here**
- #1868, PR 4: Pin Rust toolchain via rust-toolchain.toml and simplify CI
- #1869, PR 5: Pickle migration script + release note
- #1870, PR 6: Bump pyo3 from 0.27.2 to 0.28.x

Parent: #1866 (PR 2). Child: #1868 (PR 4).


MaxHalford:
Agreed! You're right, we should modularize the Rust code too, in preparation for when we ship River as a Rust crate.

PR #1886, opened 2026-06-11
feat(preprocessing): add windowed scaling and warm-start to scalers

MarekWadinger:
## Summary

Two related, atomic extensions to `preprocessing.MinMaxScaler`, `preprocessing.MaxAbsScaler`, and `preprocessing.StandardScaler`.

### 1. `window_size` (all three scalers)

A new `window_size: int | None = None` keyword argument switches the underlying running statistics to their rolling counterparts:

| Scaler | Running (default) | Rolling (`window_size` set) |
|---|---|---|
| `MinMaxScaler` | `stats.Min` / `stats.Max` | `stats.RollingMin` / `stats.RollingMax` |
| `MaxAbsScaler` | `stats.AbsMax` | `stats.RollingAbsMax` |
| `StandardScaler` | Welford `mean` / `var` | `utils.Rolling(stats.Mean())` / `utils.Rolling(stats.Var(ddof=0))` |

The `MinMaxScaler` / `MaxAbsScaler` branching follows the established `anomaly.GaussianScorer` pattern (kwarg dispatch via `functools.partial` in the `defaultdict` factory).

For `StandardScaler`, two extra design points were required:

- **Population variance (`ddof=0`)** is used in windowed mode so that the windowed and running modes agree exactly while `n ≤ window_size`. (River's non-windowed Welford estimator is also population variance.)
- A small picklable `_RollingStatFactory` is used instead of the obvious `functools.partial(utils.Rolling, stats.Mean(), n)`, which would silently share a single `stats.Mean()` instance across every key of the `defaultdict` — a per-feature aliasing bug.

```python
scaler = preprocessing.MinMaxScaler(window_size=100)
scaler = preprocessing.StandardScaler(window_size=100)
```

This is useful when the data distribution drifts and a global running statistic would saturate the scaling range.

#### Windowed `StandardScaler` — known limitations and compromises

- `learn_many` in windowed mode falls back to row-by-row updates, because the existing mini-batch Welford merge formula (`a·old_var + b·new_var + a·b·(old_mean − new_mean)²`) cannot be inverted to support eviction.
- `_from_state` does not seed the rolling window from scalar statistics; for a windowed warm-start, replay recent observations through `learn_one` instead. (This is documented in the classmethod's docstring.)
- `counts` continues to track cumulative observations (not window occupancy) in windowed mode — purely informational, kept for API symmetry with the non-windowed branch.
- `AdaptiveStandardScaler` remains the right choice when an EWMA-style "weight recent observations more" behavior is desired rather than a hard window.

### 2. `_from_state` warm-start (all three scalers)

A new `_from_state` classmethod constructs a scaler with pre-populated running statistics, so a stream can be warm-started from an offline estimate or resumed from a checkpoint without replaying past observations.

```python
scaler = preprocessing.StandardScaler._from_state(
    counts={'x': 100},
    means={'x': 10.0},
    vars={'x': 4.0},
)
scaler.transform_one({'x': 12.0})  # {'x': 1.0}
```

The classmethod naming and shape mirror the existing convention used elsewhere in river — `stats.Var._from_state`, `stats.Cov._from_state`, `proba.Gaussian._from_state`, `covariance.EmpiricalCovariance._from_state`.

For `MinMaxScaler` / `MaxAbsScaler`, `_from_state` composes with `window_size`: each initial value seeds one slot of the rolling window and is eventually evicted as fresh observations arrive.

## Backward compatibility

- Defaults are unchanged across all three classes; zero-argument construction remains behavior-identical to the previous version (regression-tested).
- Every non-windowed code path falls through to the original implementation unmodified — the new branches are gated by `if self.window_size is not None: ... return` at the top of each affected method.
- Pickles produced by the previous version (missing the new `window_size` attribute) are restored transparently via a small `__setstate__` shim on each of the three scalers that defaults missing `window_size` to `None`. Regression-tested.

## Test plan

- [x] `uv run pytest -n auto` — 4677 passed, 50 skipped, 0 failed
- [x] `uv run mypy river` / `uv run prek run --files <changed>` — ruff check, ruff format, mypy all pass
- [x] `test_estimators.py` — global estimator checks pass for the three scalers (including `check_pickling`, `check_clone_is_independent`, `check_predict_one_before_any_learn`)
- [x] 39 unit tests in `river/preprocessing/test_scale.py` covering:
  - warm-start (full keys, partial keys, then-extend)
  - windowed `MinMaxScaler` / `MaxAbsScaler` (eviction, type-of-stat, default-unchanged regression)
  - windowed `StandardScaler` (per-feature independence guard, numpy ddof=0 parity, drift adaptation, `learn_one` ↔ `learn_many` parity, chunked `learn_many` consistency, `transform_many` parity, pipeline integration, `with_std=False`, cumulative counts)
  - `_from_state` + `window_size` composition
  - `clone()` round-trip preserves `window_size`
  - warm-start matches a from-scratch trained scaler
  - warm-start composes with `learn_many` mini-batch path
  - pickle backward-compat: legacy (pre-`window_size`) pickles continue to work for all three scalers
- [x] Updated doctests for all three classes + repr-anchored doctests in `base/base.py`, `compose/pipeline.py`, `model_selection/sh.py`
- [x] Release note added in `docs/releases/unreleased.md`

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -1,3 +1,8 @@
 # Unreleased
 
 * Add `ppc64le` architecture to Linux wheel builds. (@ChidiebereNjoku)
+
+## preprocessing
+
+- Added `window_size` parameter to `preprocessing.MinMaxScaler` and `preprocessing.MaxAbsScaler`. When set, the scaler tracks the min/max (or absolute max) over the last `window_size` observations via `stats.RollingMin` / `stats.RollingMax` / `stats.RollingAbsMax` instead of the running statistic over the entire stream.
+- Added `_from_state` classmethod to `preprocessing.MinMaxScaler`, `preprocessing.MaxAbsScaler`, and `preprocessing.StandardScaler` so a scaler can be warm-started from offline-computed statistics or resumed from a checkpoint without replaying past observations.
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -7,6 +7,10 @@
 - Added `window_size` parameter to `preprocessing.MinMaxScaler` and `preprocessing.MaxAbsScaler`. When set, the scaler tracks the min/max (or absolute max) over the last `window_size` observations via `stats.RollingMin` / `stats.RollingMax` / `stats.RollingAbsMax` instead of the running statistic over the entire stream.
 - Added `_from_state` classmethod to `preprocessing.MinMaxScaler`, `preprocessing.MaxAbsScaler`, and `preprocessing.StandardScaler` so a scaler can be warm-started from offline-computed statistics or resumed from a checkpoint without replaying past observations.
 
+## utils
+
+- `utils.Rolling` and `utils.TimeRolling` now accept a class as their first argument and forward extra keyword arguments to its constructor, e.g. `utils.Rolling(stats.Mean, window_size=3)` or `utils.Rolling(stats.Var, window_size=3, ddof=0)`. This avoids a footgun when using these wrappers as `collections.defaultdict` factories, where the previous instance form silently shared state across keys. Passing a pre-built instance still works but now emits a `DeprecationWarning` and will be removed in a future release.
+
 ## stream
 
 - Added `stream.iter_frame`, a dataframe-agnostic row iterator powered by [Narwhals](https://narwhals-dev.github.io/narwhals/) that works with any eager dataframe (pandas, polars, PyArrow, Modin, cuDF, ...).
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -4,7 +4,7 @@
 
 ## preprocessing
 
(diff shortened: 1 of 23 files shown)


docs/releases/unreleased.md
```diff
@@ -0,0 +1,6 @@
+# Unreleased
+
+## preprocessing
+
+- Added `window_size` parameter to `preprocessing.MinMaxScaler` and `preprocessing.MaxAbsScaler`. When set, the scaler tracks the min/max (or absolute max) over the last `window_size` observations via `stats.RollingMin` / `stats.RollingMax` / `stats.RollingAbsMax` instead of the running statistic over the entire stream.
```
MaxHalford: Don't you want to add `window_size` to `preprocessing.StandardScaler` too?

river/preprocessing/scale.py
```diff
@@ -46,7 +46,6 @@ class Binarizer(base.Transformer):
 
     Examples
     --------
-
```
MaxHalford, line 49: Please keep the newlines

MaxHalford (approved):

MaxHalford:
@MarekWadinger your PR has motivated me to refactor the `Rolling` class, so that in can accept classes instead of instances. This is more idiomatic, and avoids the need for `_RollingStatFactory`. I'm allowing myself to bundle this into your PR. Cheers!

MarekWadinger:
Nice @MaxHalford, thanks for collaboration!

Issue #1881, opened 2026-06-01
Support dataframe agnostic mini-batching


MaxHalford:
> Hey @MaxHalford 👋🏼 I hope it's ok for me to chime in and try my luck in here!
> 
> We briefly interacted in the narwhals discord _long time ago_. Would you consider adopting [narwhals](https://github.com/narwhals-dev/narwhals) to make _any_ dataframe library optional? We greatly extended support since we last spoke (now including pyspark, duckdb and ibis as well) - The one special casing left would be vaex (which we are not planning to support directly anytime soon, unless a community plugin in made).
> 
> If you are curious to see how it would look like, I even [started an effort](https://github.com/FBruzzesi/river/pull/1) after you join the discord 😇 

 _Originally posted by @FBruzzesi in [#1805](https://github.com/online-ml/river/issues/1805#issuecomment-4273920509)_

PR #1890, opened 2026-06-15
feat(stream): Add dataframe-agnostic `iter_frame` via narwhals, deprecate `iter_pandas`/`iter_polars`

FBruzzesi:
Part of #1881. Adds `stream.iter_frame`, a single [Narwhals](https://narwhals-dev.github.io/narwhals/)-powered row iterator that works with any eager backend (pandas, polars, PyArrow, Modin, cuDF, ...).

`iter_pandas`/`iter_polars` become deprecated thin wrappers that delegate to it.

This covers input iteration only for now; dataframe-agnostic mini-batch methods are a follow-up.

**Notes**:

* `iter_vaex` is left untouched: Narwhals only supports vaex at the interchange level, which would force materialization and kill its design.
* Backward compatible: old functions still work, just warn.

**Open questions**:

* Testing: I tried to address only the direct changes. In practice there are not direct tests for `iter_pandas`/`iter_polars`. Should I add direct tests for it, including pyarrow?
* Docs: Anything I need to change in the API? I am a bit lost for how the API reference is generated 🙈

### DIFF
--- a/docs/recipes/feature-extraction.ipynb
+++ b/docs/recipes/feature-extraction.ipynb
@@ -498,7 +498,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 27,
    "metadata": {},
    "outputs": [
     {
--- a/docs/recipes/feature-extraction.ipynb
+++ b/docs/recipes/feature-extraction.ipynb
@@ -498,7 +498,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 27,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -523,7 +523,7 @@
     "\n",
     "rolling_history = []\n",
     "\n",
-    "for x, y in stream.iter_pandas(\n",
+    "for x, y in stream.iter_frame(\n",
     "    df[[\"open\", \"high\", \"low\", \"close\", \"volume\"]],\n",
     "    df[\"target\"],\n",
     "):\n",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,9 +26,6 @@ Repository = "https://github.com/online-ml/river/"
 requires = ["maturin>=1.13,<2"]
 build-backend = "maturin"
 
-[[tool.uv.index]]
-url = "https://pypi.org/"
-
 [tool.maturin]
 module-name = "river._river_rust"
 python-source = "."
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
     "scipy>=1.16,<2",
     "numpy>=2.3.4,<3",
     "altair>=5.0.0",
+    "narwhals>=2.18.1",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ dev = [
     "jupyter>=1.0.0",
     "mike @ git+https://github.com/squidfunk/mike.git",
     "polars>=1.1.0",
+    "pyarrow>=11.0.0",
 ]
 compat = [
     "scikit-learn>=1.5.1,<2",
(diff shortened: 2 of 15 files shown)


docs/recipes/feature-extraction.ipynb
```diff
@@ -498,7 +498,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 27,
+   "execution_count": null,
```
FBruzzesi:
```suggestion
   "execution_count": 27,
```

pyproject.toml
```diff
@@ -25,6 +26,9 @@ Repository = "https://github.com/online-ml/river/"
 requires = ["maturin>=1.13,<2"]
 build-backend = "maturin"
 
+[[tool.uv.index]]
+url = "https://pypi.org/"
+
```
FBruzzesi:
```suggestion
```

MaxHalford (approved):
That looks good, thanks for a nice addition and the proactivity! The docstring looks correct, and anyway I'll audit the docs before releasing the new version. 

Indeed you could add a few tests. The dev dependencies should be cached in CI, so it shouldn't be too much of a slowdown. I'd love to see a DuckDB unit test 🦆

FBruzzesi:
Hey @MaxHalford thanks for your feedback! I will add a test for pyarrow in the next iteration.

## Lazy Backends

Regarding lazy backends, such as DuckDB, we don't support `to_numpy`, matching `polars.LazyFrame` API, as the data is not materialized into memory.

1. In general, we never trigger a collect for a user
2. On top of it, all lazy backends except for polars do not guarantee row order unless explicitly sorted. I would expect that order can impact some learning algorithms (e.g. time series ones)

## Vaex

What's your take on it?

~~After sleeping on it, I was thinking of special casing it within `iter_frame` itself in a lazy way, so that all the dataframe handling is centralized.~~ Nevermind, the spec is quite different.

## `iter_array`

In principle, we could, and maybe should, avoid converting to numpy and passing the results to `iter_array`. 

Addressed in https://github.com/online-ml/river/pull/1890/commits/5c0028c01b81a8ecbb6e6dc974f66188d66a1ad3

MaxHalford:
Thanks for the insights. Generally speaking numpy is not required: what River needs are dicts. So if there are faster ways to go from the frame to the dict, then it's fine for River. For instance there might be ways to go from DuckDB to an iterator of dicts. The sort order is the responsibility of the user in the SQL query, IMO

river/stream/iter_frame.py
```diff
@@ -0,0 +1,137 @@
...
+    # Accept any frame so we can raise an explicit error for the lazy case, rather than letting
+    # Narwhals reject it with a generic message.
+    frame = nw.from_native(X, allow_series=False)
+    if isinstance(frame, nw.LazyFrame):
+        raise TypeError(
+            f"`stream.iter_frame` only supports eager dataframes, but got a lazy frame "
+            f"({type(X).__name__!r}). Iterating rows would force a full materialization, and most "
+            f"lazy backends do not guarantee row order. Collect it first (e.g. via `.collect()`) "
+            f"and pass the resulting eager frame."
+        )
```
FBruzzesi, line 115: @MaxHalford for now I added this dedicated message when detecting a lazyframe. Let me know what you think

MaxHalford: Yes I think that's fine. explicit > magic

MaxHalford (approved):

MaxHalford:
Thank you @FBruzzesi!

FYI I've taken a stab at using Narwhals in [Prince](https://github.com/MaxHalford/prince), but it's really difficult 😅

FBruzzesi:
Thanks for merging @MaxHalford 🙏🏼

> FYI I've taken a stab at using Narwhals in [Prince](https://github.com/MaxHalford/prince), but it's really difficult 😅

Anything specific that is giving you a hard time? Feel free to ping me in a PR/draft/anything 😅

PR #1900, opened 2026-06-19
feat(linear_model): dataframe-agnostic mini-batching via narwhals

FBruzzesi:
# Description

Related to #1881  

Starts the incremental mini-batch migration with `linear_model`.

`LinearRegression` and `LogisticRegression` mini-batch methods (`learn_many`, `predict_many`, `predict_proba_many`) now accept and return any narwhals-supported eager backend (pandas, polars, pyarrow, ...) instead of being pandas-only. 
The input backend is preserved on output, including the pandas index, so existing pandas callers are unaffected.

These methods no longer require pandas to be installed.


river/linear_model/base.py
```diff
@@ -168,32 +171,49 @@ def learn_one(self, x, y, w=1.0) -> None:
...
-        self._y_name = y.name
-        saved = self._enter_learn_mode(set(X))
+        saved = self._enter_learn_mode(set(cols))
         try:
-            self._fit(X, y, w, get_grad=self._eval_gradient_many)
+            # `_fit`'s `get_grad(x, y, w)` reuses its first argument both to compute the gradient
+            # and to drive the weight update. For mini-batches that argument is the column names,
+            # while the feature matrix is bound here via `partial`.
+            get_grad = functools.partial(self._eval_gradient_many, X_np)
+            self._fit(x=cols, y=y_np, w=w_np, get_grad=get_grad)
```
FBruzzesi, line 218: I am not super proud of this "hack"

river/stream/test_iter_frame.py
FBruzzesi, line 1: Changes here are just due to moving the backend fixture (and renaming it) into `conftest.py`

river/utils/dataframe.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def to_native_frame(
+    data: Mapping[Any, NDArray[Any] | Sequence[Any]], *, like: nw.DataFrame[Any]
+) -> IntoDataFrame:
+    """Build a native dataframe matching the backend (and pandas index) of `like`.
+
+    Pandas-like backends accept arbitrary column labels (e.g. the booleans `False`/`True`
+    used by binary classifiers), so they are preserved as-is. Every other backend requires
+    string column names, so labels are stringified for them.
```
FBruzzesi, line 91: Can't really do much about this divergence: pandas keeps arbitrary column labels (the booleans False/True from binary classifiers), polars/pyarrow require string names so labels are stringified.

Doing the same for pandas would be a breaking change, which I wanted to avoid.

So `predict_proba_many` columns are `[False, True]` on pandas-like and `["False","True"]` elsewhere.

river/utils/dataframe.py
```diff
@@ -0,0 +1,96 @@
...
+    like
+        The narwhals dataframe the call received as input. Its backend determines the
+        return type, and its index (pandas only) is carried over to the result.
+    """
+    # TODO(FBruzzesi): narwhals' stub types `name` as `str`, but every backend accepts `None` at runtime.
+    series = nw.new_series(name=name, values=values, backend=nw.get_native_namespace(like))  # type: ignore[arg-type]
+    native = series.to_native()
+    # Carry over the pandas index; no-op for non-pandas backends (maybe_get_index -> None).
+    if (index := nw.maybe_get_index(like)) is not None:
+        native.index = index
```
FBruzzesi, line 80: Index is restored by assigning to the native object rather than `nw.maybe_set_index` for better ergonomics, see https://github.com/narwhals-dev/narwhals/issues/3693

river/linear_model/log_reg.py
```diff
@@ -96,7 +98,19 @@ def predict_proba_one(self, x):
...
-        return pd.DataFrame({False: 1.0 - p, True: p}, index=X.index, copy=False)
+    def predict_proba_many(self, X: IntoDataFrame) -> IntoDataFrame:
+        X_nw = utils.dataframe.into_frame(X)
+        p: NDArray[np.float64] = self.loss.mean_func(
+            self._raw_dot_many(X_nw.to_numpy(), X_nw.columns)
+        )  # logits to probas
+        # pandas keeps the boolean column labels; other backends require string names.
+        return utils.dataframe.to_native_frame({False: 1.0 - p, True: p}, like=X_nw)
+
+    def predict_many(self, X: IntoDataFrame) -> IntoSeries:
```
FBruzzesi, line 109: Overrides `MiniBatchClassifier.predict_many` because the base default uses pandas-only `.idxmax(axis="columns")`. I am assuming multiclass is not possible since `predict_proba` and `preduct_proba_many` were/are using `{False: 1.0 - p, True: p}`

river/linear_model/base.py
MaxHalford: It's ok :)

river/linear_model/log_reg.py
MaxHalford: Correct 👍

river/linear_model/test_glm.py
```diff
@@ -483,3 +497,159 @@ def test_log_reg_sklearn_l1_non_regression():
...
+    model = Model()
+    model.learn_many(frame_backend.frame(data), frame_backend.series(targets))
+
+    assert model.weights.keys() == reference.weights.keys()
+    for key, weight in reference.weights.items():
+        assert math.isclose(model.weights[key], weight, rel_tol=1e-12, abs_tol=1e-12)
+    assert math.isclose(model.intercept, reference.intercept, rel_tol=1e-12, abs_tol=1e-12)
+
+
+def test_learn_many_matches_learn_one(frame_backend: FrameBackend, reg_batch: Columnar) -> None:
```
MaxHalford, line 549: So pytest will automatically run these tests for all available backends?

river/utils/dataframe.py
```diff
@@ -0,0 +1,96 @@
...
+
+River's ``*_many`` methods compute in numpy/scipy but historically took and returned
+``pandas`` objects. These helpers move the pandas dependency to the boundary via
+[narwhals](https://github.com/narwhals-dev/narwhals): the input is wrapped on entry, the
+numpy core is left untouched, and the output is rebuilt with the *same* backend (and pandas
+index) as the input. This lets any narwhals-supported eager backend (pandas, polars,
+pyarrow, ...) flow through unchanged.
+
+See `river/utils/pandas.py` for the pandas-optional import helper.
+"""
```
MaxHalford, line 11: Actually here I'd be in favour of merging this with the `utils/pandas.py` file, and calling it `utils/mini_batch.py`. It's more intentional

river/utils/dataframe.py
```diff
@@ -0,0 +1,96 @@
...
+    string column names, so labels are stringified for them.
+
+    Parameters
+    ----------
+    data
+        A mapping from column label to column values (numpy arrays from the numpy core).
+    like
+        The narwhals dataframe the call received as input. Its backend determines the
+        return type, and its index (pandas only) is carried over to the result.
+    """
```
MaxHalford, line 101:
```suggestion
    like
        The narwhals dataframe the call received as input. Its backend determines the
        return type, and its index (pandas only) is carried over to the result.
        
    """
```

river/conftest.py
```diff
@@ -53,3 +61,42 @@ def _uses_pandas(path: pathlib.Path) -> bool:
...
+
+FRAME_BACKENDS: dict[str, Callable[[], FrameBackend]] = {
+    "pandas": _pandas,
+    "polars": _polars,
+    "pyarrow": _pyarrow,
+}
+
+
+@pytest.fixture(params=list(FRAME_BACKENDS))
+def frame_backend(request: pytest.FixtureRequest) -> FrameBackend:
```
MaxHalford, line 134: Nice

MaxHalford (approved):
This is wonderful :)

river/linear_model/test_glm.py
FBruzzesi: That's correct! As long as the libraries are installed, all the backends will run.

river/conftest.py
FBruzzesi: As a side note, in narwhals we run pandas in 3 dtype backend flavours: None, numpy nullable _and_ pyarrow.

You can see a streamlined version for the implementation I did in fairlearn ([conftest.py](https://github.com/fairlearn/fairlearn/blob/main/test/conftest.py)). Let me know if you are interested in having the same setup

river/utils/dataframe.py
FBruzzesi: As a note: the long term result of this effort is to remove pandas specific helper functions completely.
On top of it, I would keep the name `dataframe` (or similar) so that it's more clear in scope. For instance, we can refactor `iter_frame` via these helpers now - I didn't do it here to keep the PR more contained

FBruzzesi:
Thanks @MaxHalford - Will address your comments later tonight :)

river/conftest.py
MaxHalford: Ok that's pretty cool! Generally speaking the more tests there are, the merrier. What about pandas 2 vs. 3? Is that something you handle too?

FBruzzesi: We tend to test everything we can. We already have full support for pandas 3 i.e. aligning its behavior with polars. In narwhals code, that translates to something like:

```py
impl = self._implementation
if impl.is_pandas() and impl._backend_version() < (3, 0):  # pragma: no cover
    # NOTE: Keep `inplace=True` to avoid making a redundant copy.
    result_native.drop(columns=extra, inplace=True)  # noqa: PD002
    return result_native

return result_native.drop(columns=extra)
```
[source](https://github.com/narwhals-dev/narwhals/blob/71a020412b6ce6b756b70bbde76f69fac74edaea/src/narwhals/_pandas_like/dataframe.py#L635-L640)

river/utils/dataframe.py
```diff
@@ -0,0 +1,109 @@
...
+
+def to_numpy(frame: nw.DataFrame[Any]) -> NDArray[np.float64]:
+    """Extract a `float64` numpy matrix from a narwhals dataframe for the numpy compute core.
+
+    A pandas frame backed by pyarrow (`ArrowDtype`) columns returns an ``object`` array from
+    ``.to_numpy()``, which breaks downstream ufuncs (e.g. ``np.exp`` raises *"loop of ufunc does
+    not support argument 0 of type float"*). Coercing to ``float64`` at the boundary keeps the
+    core backend-agnostic; ``np.asarray`` is a no-op when the frame already yields ``float64``.
+    """
+    return np.asarray(frame.to_numpy(), dtype=np.float64)
```
FBruzzesi, line 43: @MaxHalford the docstrings should do it for the explanation, but it's worth adding that on top of that, the float64 cast is deliberate as the weight vector is float64 (`utils.VectorDict.to_numpy` builds it via `np.array` over Python floats), so every downstream matmul/einsum promotes to float64 regardless of the input dtype. Casting once here can result in even cheaper compute than letting each op re-promote a narrower dtype

PR #1904, opened 2026-06-20
Fix TFIDF mini-batch transforms for DataFrame input

### LINKED ISSUE #1576: `TFIDF.transform_many()` fails on `DataFrame` input
## Versions

**river version**: 0.21.2
**Python version**: 3.11.7
**Operating system**: macOS 14.4

## Describe the bug

The [`TFIDF` feature extractor](https://riverml.xyz/latest/api/feature-extraction/TFIDF/) claims to support both online and mini-batch transformations, but the latter case only works when the transformer doesn't specify the `on` parameter. In other words, batch mode works for `pd.Series` input, but not `pd.Dataframe`.

## Steps/code to reproduce

```python
import pandas as pd
import river.feature_extraction

model = river.feature_extraction.TFIDF()
X = pd.Series(["foo bar bat baz", "foo bar spam eggs"])
for rec in X:
    print(model.transform_one(rec))
# WORKS
# {'foo': 0.5, 'bar': 0.5, 'bat': 0.5, 'baz': 0.5}
# {'foo': 0.5, 'bar': 0.5, 'spam': 0.5, 'eggs': 0.5}
print(model.clone().transform_many(X))
# WORKS
#    foo  bar  bat  baz  spam  eggs
# 0    1    1    1    1     0     0
# 1    1    1    0    0     1     1

model = river.feature_extraction.TFIDF(on="text")
X = pd.DataFrame([{"text": "foo bar bat baz"}, {"text": "foo bar spam eggs"}])
for rec in X.to_dict(orient="records"):
    print(model.transform_one(rec))
# WORKS
# {'foo': 0.5, 'bar': 0.5, 'bat': 0.5, 'baz': 0.5}
# {'foo': 0.5, 'bar': 0.5, 'spam': 0.5, 'eggs': 0.5}
print(model.clone().transform_many(X))
# DOES NOT WORK
```

That last call produces the following traceback:

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[95], line 1
----> 1 print(model.clone().transform_many(X))

File [~/.pyenv/versions/3.11.7/envs/ds/lib/python3.11/site-packages/river/feature_extraction/vectorize.py:349](http://localhost:8888/lab/tree/Desktop/notebooks/~/.pyenv/versions/3.11.7/envs/ds/lib/python3.11/site-packages/river/feature_extraction/vectorize.py#line=348), in BagOfWords.transform_many(self, X)
    347 for d in X:
    348     t: int
--> 349     for t, f in collections.Counter(self.process_text(d)).items():
    350         indices.append(index.setdefault(t, len(index)))
    351         data.append(f)

File [~/.pyenv/versions/3.11.7/envs/ds/lib/python3.11/site-packages/river/feature_extraction/vectorize.py:220](http://localhost:8888/lab/tree/Desktop/notebooks/~/.pyenv/versions/3.11.7/envs/ds/lib/python3.11/site-packages/river/feature_extraction/vectorize.py#line=219), in VectorizerMixin.process_text(self, x)
    218 def process_text(self, x):
    219     for step in self.processing_steps:
--> 220         x = step(x)
    221     return x

TypeError: string indices must be integers, not 'str'
```

BALOGUN-DAVID:
## Summary

- Add `TFIDF.learn_many` so mini-batch learning updates document frequencies.
- Add `TFIDF.transform_many` to return TF-IDF sparse DataFrame output.
- Support `DataFrame` input when vectorizers use `on=...`.
- Add regression tests for Series, DataFrame, non-normalized TF-IDF, and empty documents.

Fixes #1576.

## Tests

- `uv run  pytest river/feature_extraction/test_vectorize.py`
- `uv run  ruff check river/feature_extraction/vectorize.py river/feature_extraction/test_vectorize.py`
- `git diff --check`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,5 +29,5 @@ repos:
           - "--implicit-optional"
         additional_dependencies:
           - "narwhals"
-          - "numpy"
+          - "numpy==2.4.3" # keep in sync with uv.lock; newer stubs use 3.12 `type` syntax mypy rejects at --python-version=3.11
           - "pytest"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,28 +6,26 @@ repos:
       - id: check-json
       - id: check-yaml
 
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version, should be the same as in uv.lock
-    rev: v0.15.8
+  # ruff and mypy run from the uv environment (`uv run`), so their versions and the
+  # dependencies they type-check against (narwhals, numpy, pytest, ...) come from the uv
+  # dev group / uv.lock — no `rev` pins or `additional_dependencies` to keep in sync here.
+  - repo: local
     hooks:
-      # Run the linter.
       - id: ruff-check
+        name: ruff-check
+        entry: uv run ruff check --fix
+        language: system
         types_or: [python, pyi, jupyter]
-        args: [--fix]
-      # Run the formatter.
+        require_serial: true
       - id: ruff-format
+        name: ruff-format
+        entry: uv run ruff format
+        language: system
         types_or: [python, pyi, jupyter]
-
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    # MyPy version, should be the same as in uv.lock
-    rev: v1.19.1
-    hooks:
+        require_serial: true
       - id: mypy
-        args:
-          - "--config-file=pyproject.toml"
-          - "--python-version=3.11"
-          - "--implicit-optional"
-        additional_dependencies:
-          - "narwhals"
-          - "numpy==2.4.3" # keep in sync with uv.lock; newer stubs use 3.12 `type` syntax mypy rejects at --python-version=3.11
-          - "pytest"
+        name: mypy
+        entry: uv run mypy --config-file=pyproject.toml --python-version=3.11 --implicit-optional
+        language: system
+        types_or: [python, pyi]
+        require_serial: true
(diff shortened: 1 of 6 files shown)


MaxHalford (commented):
- Can you confirm you address @bdewilde's comments [here](https://github.com/online-ml/river/issues/1576)? 
- Could you take a look if this mini-batch implementation is aligned with that of sklearn? It would be nice to add more unit tests to check both implementations can match.

BALOGUN-DAVID:
Yes, this now addresses @bdewilde's comments: `TFIDF.learn_many` updates the document-frequency state, and `TFIDF.transform_many` returns TF-IDF weights rather than inheriting BagOfWords count output.

I also added sklearn parity tests for the normalized default case, covering both `Series` input and `DataFrame` input with `on="text"`. The focused vectorizer tests pass locally.

MaxHalford:
Hey @FBruzzesi, what's the Narwhals convention for handling sparse frames? Here we have TF-IDF which should be able to support Narwhals inputs, but I'm not sure what to output. I was thinking of outputting a sparse pandas DataFrame if the input is pandas, and densifying the output if not. Any advice?

FBruzzesi:
Hey @MaxHalford 👋🏼 

> what's the Narwhals convention for handling sparse frames?

That's kind of tricky as pandas is the only backend that supports sparsity. That might be the one thing we _don't test_ as it would be pandas specific.

For context, I need to push the changes for `OneHotEncoder` and I am keeping a special path for pandas as `pd.get_dummies` returns sparse frame, while [`polars.Series.to_dummies`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.to_dummies.html) doesn't. 

> I was thinking of outputting a sparse pandas DataFrame if the input is pandas, and densifying the output if not.

Yep that's the only way I suppose

MaxHalford:
Thanks for the quick answer!

PR #1913, opened 2026-06-23
Speed up the linear models and FeatureHasher

MaxHalford:
## What

Two performance fixes for sparse, high-cardinality workloads (e.g. CTR-style data), plus doc cleanups.

### Linear models: sparse updates
`learn_one` for the generalized linear models (`LinearRegression`, `LogisticRegression`, `Perceptron`, ...) previously cost **O(total features ever seen)** per example, because the masked weight `VectorDict` scanned the full backing dictionary in its dot product and update. It now scans the (small) mask instead when that's the smaller side, so updates cost **O(active features)**.

- Output-preserving: the masked key set is unchanged; only the dot-product summation order shifts (to floating-point roundoff), which already matched what `predict` did.
- On a 100k-row hashed Criteo stream, `LogisticRegression` learn+predict went from ~290 ex/s (degrading as the weight vector grew) to a flat ~10.7k ex/s.

### FeatureHasher: MurmurHash3 in Rust
`preprocessing.FeatureHasher` hashed every token with `blake2s` (a cryptographic hash) plus a hex-string/bigint round-trip — ~40% of a hashed pipeline's time. It now hashes with **MurmurHash3 (x86 32-bit)** in Rust, doing the whole `transform_one` in one native call (~16× faster).

- The Rust hash is verified bit-for-bit identical to scikit-learn's `murmurhash3_32` (no new dependency — sklearn and VW both vendor their own murmur for the same reason).
- Gains an `alternate_sign` parameter (default `True`, matching scikit-learn) and returns a plain `dict`.
- **Behavior change**: hashed feature indices differ from previous versions, and `alternate_sign=True` negates ~half of contributions by default. The docstring previously *claimed* murmurhash + `alternate_sign` but did neither.

### Docs
- Restyled all `unreleased.md` entries to be concise (lead with the change + user-facing effect, drop deep internals) and documented that style in `CLAUDE.md`.
- Dropped stale "ported from Cython" comments in the Rust sources (River no longer uses Cython).
- Fixed the `AdPredictor` docstring's now-outdated implication that plain linear models do a full per-example update.
- Documented the `maturin develop` inner-loop build command in `CONTRIBUTING.md`.

## Verification
- `river/linear_model`, `river/utils`, `river/optim`, `river/facto`, `river/preprocessing` tests + doctests pass (including the exact-value FTRL/SGD doctests, which confirm the linear-model outputs are unchanged).
- `check_estimator(FeatureHasher())` passes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,5 +70,9 @@ All estimators inherit from `base.Estimator` (which inherits from `base.Base`). 
 
 ### Making changes
 
-- When you're done, add a entry to `unreleased.md` if its relevant to end users.
+- When you're done, add an entry to `docs/releases/unreleased.md` if it's relevant to end users, under the section for the relevant module (alphabetical).
+  - Use `-` for list items, not `*`.
+  - Keep entries concise: one to three sentences. Lead with what changed and its user-facing effect.
+  - Include performance numbers (e.g. "~7× faster") and call out behavior changes explicitly, but omit deep implementation mechanics (internal variable names, algebra, marshalling details) — those belong in the code/PR, not the changelog.
+  - Example: "Sped up `learn_one` for the linear models: updates now scale with the number of active features instead of the total number of features ever seen. Outputs are unchanged."
 - Performance matters: if you make a significant change, run a benchmark.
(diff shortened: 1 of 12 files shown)


raphaelsty:
@MaxHalford this is some serious improvements there 😮🙌

smastelini:
wow!!

MaxHalford:
I know it's crazy! I found by benchmarking River against Vowpal Wabbit, and digging into the performance difference

PR #1916, opened 2026-06-24
Add learn_many to BayesianLinearRegression

MaxHalford:
## What

`BayesianLinearRegression` is now a `MiniBatchRegressor` with a `learn_many` method, equivalent to looping `learn_one` over the rows.

## Why it's exact

- **Without smoothing**, the posterior precision and natural-mean updates are purely additive (Bishop 3.50/3.51), so a batch collapses the per-row rank-1 outer products into one Gram matrix `β·XᵀX` (+ `β·Xᵀy`) and a single inverse. Predictions match the loop to ~1e-15.
- **With smoothing**, each row is an EMA step `S ← s·S + (1-s)·β·xxᵀ`, which has a closed form over a batch: decay the prior by `sᴺ` and weight row *k* (oldest first) by `(1-s)·s^(N-1-k)`.

## Dataframe-agnostic

`learn_many`/`predict_many` go through the narwhals boundary like `LinearRegression`/`LogisticRegression` (#1900), so they accept and return any supported eager backend (pandas, polars, pyarrow, ...), preserve the input backend and pandas index, and no longer require `pandas`.

## Performance

`learn_many` is **66–333× faster** than the `learn_one` loop (BLAS-bound Gram matrix + one inverse):

| shape | loop | `learn_many` | speedup |
|---|---|---|---|
| 2000×30 | 14.9 ms | 0.23 ms | 66× |
| 2000×100 | 44.9 ms | 0.39 ms | 115× |
| 10000×50 | 124.8 ms | 0.38 ms | 333× |

## Global equivalence check

Adds `check_learn_many_matches_learn_one` to the estimator-check suite: it trains a model both ways under **emerging/disappearing features** and asserts the predictions agree. Estimators whose `learn_many` is not contractually equivalent declare it in `_unit_test_skips`:

- the gradient-descent linear models (`GLM` → `LinearRegression`, `LogisticRegression`, `Perceptron`): batch mean-gradient ≠ per-row SGD. Pipelines ending in these inherit the skip via `Pipeline._unit_test_skips`.
- naive Bayes (`BaseNB`): `learn_many` consumes sparse count matrices, and its equivalence is covered by the dedicated NB test suite.

Among all mini-batch estimators, `BayesianLinearRegression` is the one that actually runs the new check (and passes).

## Tests

- New backend-parametrized tests in `test_glm.py`: `predict_many` returns the native backend; `learn_many` matches the `learn_one` loop (smoothing on/off); chunked `learn_many` equals a single batch.
- `check_estimator` passes for `BayesianLinearRegression` (now including the mini-batch checks).

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -24,6 +24,7 @@
 - Sped up `learn_one` for the linear models (`LinearRegression`, `LogisticRegression`, `Perceptron`, ...): updates now scale with the number of active features instead of the total number of features ever seen. Outputs are unchanged.
 - Stabilised `BayesianLinearRegression` across BLAS implementations and sped it up (~10-20%) by accumulating an exact natural mean and recovering the posterior mean lazily, instead of propagating it through compounding rank-1 updates (which drifted ~0.6% between macOS Accelerate and Linux OpenBLAS).
 - `linear_model.LinearRegression` and `linear_model.LogisticRegression` mini-batch methods (`learn_many`, `predict_many`, `predict_proba_many`) now accept and return any [narwhals](https://github.com/narwhals-dev/narwhals)-supported eager backend (pandas, polars, pyarrow, ...) instead of being pandas-only. The input backend is preserved on output, including the pandas index. These methods no longer require `pandas` to be installed.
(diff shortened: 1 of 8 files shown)


MaxHalford:
Note: the transformer test-coverage fix that briefly lived here has moved to its own PR, #1918 (it makes the transformer base classes properly abstract, a cleaner root-cause fix than the identity-exclusion workaround). This PR no longer touches `test_estimators.py`. The generalized `check_learn_many_matches_learn_one` here is registered for mini-batch transformers, so once #1918 lands, concrete transformers are automatically covered by it.

Issue #1919, opened 2026-06-24
Migrate all mini-batch (_many) methods to narwhals for dataframe-agnostic support


MaxHalford:
## Motivation

[#1900](https://github.com/online-ml/river/pull/1900) introduced dataframe-agnostic mini-batching via [narwhals](https://github.com/narwhals-dev/narwhals): inputs are wrapped at the method boundary, the numpy compute core stays untouched, and outputs are rebuilt in the caller's native backend (pandas / polars / pyarrow / nullable / arrow-backed pandas). The helpers live in `river/utils/dataframe.py`:

- `into_frame` / `into_series` — wrap native inputs
- `to_numpy` — extract a float64 array for the compute core
- `to_native_frame` / `to_native_series` — rebuild output in the caller's backend, preserving the pandas index

So far only `linear_model` (GLM + `LinearRegression`, `LogisticRegression`, `BayesianLinearRegression`) uses this. Every other mini-batch method still hard-codes pandas (`pd.DataFrame`/`pd.Series` signatures, `.values`, `.columns`, etc.). The goal of this issue is to migrate **all** `_many` methods to narwhals so that any narwhals-supported backend works end to end.

## Scope

Mini-batch methods to migrate (`learn_many`, `predict_many`, `predict_proba_many`, `transform_many`, and friends):

**Base classes (signatures/type hints — do first, they're authoritative):**
- [ ] `base/classifier.py` — `MiniBatchClassifier`
- [ ] `base/regressor.py` — `MiniBatchRegressor`
- [ ] `base/transformer.py` — `MiniBatchTransformer`

**Concrete estimators (still pandas-only):**
- [x] `preprocessing/scale.py` — `StandardScaler` & co. (`.values`/`.columns`) https://github.com/online-ml/river/pull/1932
- [ ] `naive_bayes/base.py` — `BaseNB` (`joint_log_likelihood_many`, `predict_proba_many`, pandas `.subtract`)
- [x] `multiclass/ovr.py` — `OneVsRest` https://github.com/online-ml/river/pull/1933
- [x] `anomaly/lof.py` — `LocalOutlierFactor` https://github.com/online-ml/river/pull/1929
- [x] `anomaly/svm.py` — `OneClassSVM` https://github.com/online-ml/river/pull/1930
- [x] `feature_extraction/vectorize.py` https://github.com/online-ml/river/pull/1904
[...]

PR #1921, opened 2026-06-24
feat(preprocessing): Make `OrdinalEncoder` mini-batching dataframe-agnostic

FBruzzesi:
`OrdinalEncoder.learn_many` and `transform_many` were pandas-only. They now work with any narwhals-supported
eager backend (pandas, polars, pyarrow, ...) and no longer import pandas.

- `learn_many` registers categories via a single `unique(maintain_order=True)` pass per column, so codes are assigned in first-appearance order on every backend (polars' default `unique` does not preserve order), matching `learn_one`.
- `transform_many` encodes via vectorised `replace_strict` (known categories → codes, everything else → `unknown_value`) and folds missing cells into `none_value`. The input backend and shape are preserved on output.

## Behavior change

Missing cells (`None`, `NaN`, `pd.NA`) now encode to `none_value` consistently across backends. Previously the pandas `transform_many` mapped a numeric/object `NaN` to `unknown_value` (only `None` hit `none_value`). This aligns
`transform_many` with `transform_one` and across backends.

---

Related #1881, #1919

### DIFF
--- a/river/preprocessing/ordinal.py
+++ b/river/preprocessing/ordinal.py
@@ -182,8 +182,9 @@ class OrdinalEncoder(base.MiniBatchTransformer):
 
         """
         X_nw = utils.dataframe.into_frame(X)
+        schema = X_nw.schema
         exprs: list[nw.Expr] = []
-        for col in X_nw.columns:
+        for col, dtype in schema.items():
             mapping = self.values[col]
             # `replace_strict` maps each known category to its code and everything else to
             # `unknown_value` in one vectorised pass. An empty mapping (a column never seen during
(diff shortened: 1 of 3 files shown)


river/preprocessing/ordinal.py
```diff
@@ -166,22 +169,72 @@ def learn_one(self, x):
...
+            # Missing cells map to `none_value` separately, since `replace_strict` leaves them
+            # untouched. A numeric `NaN` is missing too, but polars/pyarrow keep it distinct from
+            # null, so it is folded in explicitly (mirroring `learn_many`, which also skips it).
+            # `is_nan` errors on non-numeric dtypes, hence the guard.
+            is_missing = nw.col(col).is_null()
+            if dtype.is_numeric():
+                is_missing = is_missing | nw.col(col).is_nan()
+
+            exprs.append(
+                nw.when(is_missing)  # type:ignore[arg-type]
```
FBruzzesi, line 211: `# type:ignore[arg-type]` is related to https://github.com/narwhals-dev/narwhals/pull/3708

MaxHalford (approved):
Great stuff. FYI I opened https://github.com/online-ml/river/issues/1919 to keep track

PR #1926, opened 2026-06-26
feat(preprocessing): Make `OneHotEncoder` mini-batching dataframe-agnostic

FBruzzesi:
`OneHotEncoder.{learn_many, transform_many}` now accept any narwhals-supported eager backend (pandas, polars, pyarrow, ...) and no longer import pandas.

* As mentioned in #1904 , there is a special path for pandas so that the sparse behavior can be kept as is.
* In [a commit](https://github.com/online-ml/river/tree/8223723a7540dd2a35861d716c8e7295ddb9e917), I am also trying to improve the type hints of attributes involved in `{learn_many, transform_many}`.

---

Related #1919


river/preprocessing/one_hot.py
```diff
@@ -25,8 +28,8 @@ class OneHotEncoder(base.MiniBatchTransformer):
         Categories (unique values) per feature:
             `None` : Determine categories automatically from the training data.
 
-            dict of dicts : Expected categories for each feature. The outer dict maps each feature to its inner dict.
-            The inner dict maps each category to its code.
+            dict of sets : Expected categories for each feature. The outer dict maps each feature
+            to the set of category values to encode.
```
FBruzzesi, line 32: Were these copy-pasted from OrdinalEncoder? 🙈

MaxHalford: Probably! Before AI though :D

river/preprocessing/one_hot.py
```diff
@@ -246,12 +249,22 @@ class OneHotEncoder(base.MiniBatchTransformer):
...
         self.categories = categories
-        self.values: collections.defaultdict | dict | None = None
-        self._zero_dict: dict = {}
+        self.values: (
+            collections.defaultdict[base.typing.FeatureName, set[typing.Hashable]]
+            | dict[base.typing.FeatureName, set[typing.Hashable]]
+        )
+        # `Any` keys (the names are strings) keep `transform_one`'s output assignable to the
+        # base `dict[FeatureName, Any]` return despite `dict`'s invariant key parameter.
+        self._zero_dict: dict[typing.Any, int] = {}
```
MaxHalford, line 267:
```suggestion
        self._zero_dict: dict[typing.Hashable, int] = {}
```

Isn't this more idiomatic?

MaxHalford (approved):

river/preprocessing/one_hot.py
FBruzzesi: I wanted to have the minimal impact on the typing. For full context:

* `_zero_dict` is always created as: `dict[str, int]`. If we do that, then `transform_one` return type will complain. One option is to `return typing.cast("dict[base.typing.FeatureName, int]", oh)`
* If we use `dict[typing.Hashable, int]`, then `min(oh)` will complain. One option here is to change it as  `min(oh, key=str)`. I personally like this option more.

Let me know what you think

PR #1931, opened 2026-06-26
naive_bayes/base.py (_many) methods to narwhals

DagaBhai:
# What does this PR do?

contributing to #1919 

as the approach mentioned
steps 1,2,3,4 are done for the methods in the `BaseNB` (`joint_log_likelihood_many`, `predict_proba_many`, `pandas .subtract`)

## AI writing disclosure

We welcome the use of AI tools to help with contributions. For transparency and to help us improve our review process, please indicate the level of AI involvement in this PR.

- [ ] No AI usage: the PR was written entirely by a human.
- [x] AI-assisted: some parts were suggested or improved by AI, but the PR was written and reviewed by a human.
- [ ] AI-generated: the PR was mostly or fully generated by an AI tool.


DagaBhai:
if there are any changes i got it wrong do lemme know and for the testing i have test for the `predict_proba_many` should i push that too??

MaxHalford:
Ideally I'd like there to be tests to compare against sklearn. Also, you need to support Narwhals in the `learn_many` method. 

Honestly this is the kind of migration task that a coding agent excels at, I wouldn't shy away from using one!

DagaBhai:
> Ideally I'd like there to be tests to compare against sklearn. Also, you need to support Narwhals in the `learn_many` method.

on it
 
> Honestly this is the kind of migration task that a coding agent excels at, I wouldn't shy away from using one!

oh ok got it that would speed things up thank you

DagaBhai:
changes made
`joint_log_likelihood_many` and `learn_many` in gaussian.py
testing - `test_predict_proba_many` and `test_predict_proba_many_matches_sklearn`

river/naive_bayes/test_gaussian.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
```
MaxHalford, line 1: This file should be called `test_gaussian.py`, I would say

river/naive_bayes/test_gaussian.py
```diff
@@ -0,0 +1,70 @@
...
+    skl_model.fit(X, y)
+    skl_probs = pd.DataFrame(
+        skl_model.predict_proba(X),
+        columns=skl_model.classes_,
+    )
+    river_probs.columns = list(skl_probs.columns)
+    assert np.allclose(
+        river_probs.to_numpy(),
+        skl_probs.to_numpy(),
+        atol=1e-2,
```
MaxHalford: That's a high tolerance. Can you check why the tolerance can't be lower?

DagaBhai: `np.max(np.abs(river_probs.to_numpy() - skl_probs.to_numpy()))` using this i got 0.005864465204894698 , 1e-2 is the lowest

river/naive_bayes/test_gaussian.py
DagaBhai: done

river/naive_bayes/test_gaussian.py
MaxHalford: Right, but can you check if this is an algorithmic issue? This can be indicative that the implementation is not exactly correct

DagaBhai: sure

DagaBhai: sklearn's GaussianNB uses population variance (ddof=0) because Gaussian Naive Bayes is based on maximum-likelihood estimation (MLE), where population variance is the correct estimator. Our implementation was using the library's default (ddof=1), which is intended for general statistical estimation. Instead of changing the library-wide default, I fixed it only in GaussianNB so it matches the algorithm while leaving the default behavior unchanged.

MaxHalford: Perfect, thank you, that's much better :)

PR #1935, opened 2026-06-26
perf(utils, preprocessing): Build native frames directly from 2D numpy, simplify `StandardScaler`

FBruzzesi:
* `utils.dataframe.to_native_frame` can now build a native frame directly from a 2D numpy array via `narwhals.from_numpy` (no per-column slicing), and `to_numpy` takes a dtype.
* `StandardScaler.transform_many` drops its pandas-specific fast path for a single backend-agnostic path that infers the compute dtype from the input, so the input float dtype is preserved on every backend (not just pandas). Also fixes a silent transpose when from_numpy builds a polars frame from a square, Fortran-contiguous array.

### DIFF
--- a/river/utils/dataframe.py
+++ b/river/utils/dataframe.py
@@ -40,7 +40,7 @@ __all__ = [
 ]
 
 
-def to_numpy(frame: nw.DataFrame[Any]) -> NDArray[np.float64]:
+def to_numpy(frame: nw.DataFrame[Any], dtype: np.number = np.float64) -> NDArray[np.number]:
     """Extract a `float64` numpy matrix from a narwhals dataframe for the numpy compute core.
 
     A pandas frame backed by pyarrow (`ArrowDtype`) columns returns an ``object`` array from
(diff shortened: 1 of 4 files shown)


river/utils/dataframe.py
```diff
@@ -101,16 +120,37 @@ def to_native_frame(
...
+        # For a square array with column names pinned (as here), it breaks the tie from the
+        # memory layout and reads a Fortran-contiguous array column-major: a silent transpose.
+        # `to_numpy` returns exactly such an F-contiguous array for a polars frame, and
+        # arithmetic on it (e.g. the centering/scaling in `StandardScaler`) keeps that layout.
+        # Forcing C-contiguity pins the read to row-major. Cost: a no-op (returns `data`
+        # unchanged) when `data` is already C-contiguous; otherwise a single O(rows * cols)
+        # memcpy, on the same order as building the frame, so negligible relative to the
+        # conversion itself.
+        frame = nw.from_numpy(np.ascontiguousarray(data), schema=names, backend=impl).to_native()
+    else:
```
FBruzzesi, line 150: https://github.com/narwhals-dev/narwhals/issues/3716

PR #1938, opened 2026-07-02
Add preprocessing.GapEncoder

### LINKED ISSUE #1439: Online GapEncoder
[skrub](https://skrub-data.org/stable/index.html) is a wonderful new project related to scikit-learn. You can see Gaël Varoquaux present it [here](https://www.youtube.com/watch?v=I4Sksd93bZ4). They have a transformer called `GapEncoder`: it's a way to embed fuzzy strings. This could be really powerful online, say for classifying Tweets or Twitch messages, where typos are aplenty.

We already have a way to do online TD-IDF/count vectorization. But we don't have Gamma-Poisson matrix factorization. It is doable online though. Once we have it, we could assemble the two into a nice GapEncoder class. See paper [here](https://inria.hal.science/hal-02171256v4).

This is related to #1412. Indeed, maybe this works well without Gamma-Poisson matrix factorization. For instance, we could use `decomposition.LDA`, which we already have.

Fazel94:
Closes #1439.

This adds an online `GapEncoder` to `preprocessing`. It's the streaming version of skrub's `GapEncoder`: each string is turned into character n-gram counts and factorized into a few latent topics with a Gamma-Poisson model, learning one sample at a time. Fuzzy variants of the same string ("london", "London, UK", "Lomdon") share most of their n-grams, so they end up on the same topic — which is what makes it useful for messy categorical text like hand-typed city names or chat messages.

Following the discussion on the issue:

- The n-gram vocabulary grows online, the same way `preprocessing.LDA` grows its word vocabulary (unbounded for now, no `maximum_size_vocabulary` cap).
- Topics update through `A`/`B` accumulators with a `rho` forgetting factor.
- `transform_one` is read-only: it only touches n-grams already seen and never mutates the model.

There's a runnable example in the docstring (the fuzzy city-name case @MaxHalford asked about) that shows typo'd inputs landing on the right topic. It passes `check_estimator`, and I added behavioral tests covering the clustering, transform purity, vocabulary growth, seeding, and the `on`/`strip_accents`/`lowercase` options.

I'll follow up with a fuller worked example under `docs/examples` in a separate PR.

### DIFF
--- a/river/preprocessing/gap_encoder.py
+++ b/river/preprocessing/gap_encoder.py
@@ -1,0 +1,237 @@
+from __future__ import annotations
+
+import collections
+
+import numpy as np
+
+from river import base
+from river.feature_extraction.vectorize import strip_accents_unicode
+
+__all__ = ["GapEncoder"]
+
+
+def _char_ngrams(text: str, ngram_range: tuple[int, int]) -> collections.Counter:
+    """Counts of contiguous character n-grams (spaces included).
+
+    Like sklearn's `CountVectorizer(analyzer='char')`, except whitespace runs are not collapsed.
+
+    """
+    counts: collections.Counter = collections.Counter()
+    for n in range(ngram_range[0], ngram_range[1] + 1):
+        for i in range(len(text) - n + 1):
+            counts[text[i : i + n]] += 1
+    return counts
+
+
+class GapEncoder(base.Transformer):
+    """Online Gamma-Poisson encoder for fuzzy string categories.
+
+    This is an online version of [skrub's `GapEncoder`](https://skrub-data.org/stable/reference/generated/skrub.GapEncoder.html),
+    learning one string at a time instead of in batches. Each string is turned into its character
+    n-gram counts `v`, and we look for a small set of latent topics that explain those counts:
+    `v ≈ h @ W`, where `W` holds the topic/n-gram weights and `h` says how strongly each topic is
+    activated by the string. The counts are treated as Poisson with a Gamma prior on `h`, and both
+    `h` and `W` are refined with multiplicative updates as data comes in.
+
+    The point is fuzzy matching. Strings that mean the same thing usually share most of their
+    n-grams ("london", "London, UK", "Lomdon"), so they light up the same topics even when they
+    have no exact token in common. That makes the encoder handy for messy categorical text:
+    hand-typed city names, job titles, chat messages full of typos, and so on.
+
+    The n-gram vocabulary is not fixed ahead of time; it grows as new strings show up, the same
+    way `preprocessing.LDA` grows its word vocabulary. Topics are kept up to date through two
+    accumulators `A` and `B`, with a forgetting factor `rho` that lets old observations decay.
+    `transform_one` is read-only: it only looks at n-grams it has already seen and never changes
+    the model.
+
+    Parameters
+    ----------
+    n_components
+        Number of latent topics.
+    on
+        The name of the feature that contains the text to encode. If `None`, then each
+        `learn_one` and `transform_one` should treat `x` as a `str` and not as a `dict`.
+    strip_accents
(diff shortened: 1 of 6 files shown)


MaxHalford:
Thank you! I am going on a break for a couple of weeks, so I will check your PRSs when I'm back.

Fazel94:
I wish you a fun holiday. @MaxHalford

MaxHalford:
I like it, it looks like a very good start! Could you provide some throughput benchmarks to get a sense of how fast/slow this is?

Fazel94:
Thank you,  
I will send you the benchmarks at Wednesday.

Fazel94:
## Benchmark

### Data
10 city names(London, Berlin, ...) corrupted (by deleting a letter, swaping two, doubling one) probably adding a suffix(", UK", "city", ...) and sometimes lowercased, eg "londo, uk", "London city".

4000/1000 train/test.

### Models
Skrub vs this implementation,  Both `n_components=10`.

### Speed and Quality
| | train | transform | nearest-neighbour accuracy |
|---|---|---|---|
| river GapEncoder (online, one pass) | 4,758 rows/s | 9,239 rows/s | 0.990 |
| skrub GapEncoder (batch, 5 iters, k-means++ init) | 4,761 rows/s | 16,786 rows/s | 0.997 |
| skrub GapEncoder (batch, 5 iters, random init) | 4,718 rows/s | 14,924 rows/s | 0.993 |


### Analysis
Same training speed, transform is half the speed due to using `transform_one` on each new example vs batching in Skrub.  

Setting `rho` to 0.999 make the accuracy 0.990 but a `rho` of 0.95 leads to 0.947 accuracy, It is due to `rho` in skrub in batch and in current implementation applied per example, that causes forgetting value 0.95 to be very harsh.

Fazel94:
I can calculate `rho` to be rescaled to match Skrub in a given `batch_size`, I call it `half_life` it would be `self.decay = 0.5 ** (1 / half_life)` defaults to 1000(effectively `rho=0.993`), putting to `float('inf')` makes the model to never forget.

`rho` doesn't fit here since it is applied per-batch and here we process per example and thus  setting a value to similar Skrub leads to over-forgetting.

river/preprocessing/gap_encoder.py
```diff
@@ -0,0 +1,243 @@
...
+    training:
+
+    >>> for x in ["London, UK", "Lndon", "Paris, France", "Pariss"]:
+    ...     topics = enc.transform_one(x)
+    ...     print(x, max(topics, key=topics.get), {k: round(v, 3) for k, v in topics.items()})
+    London, UK 1 {0: 0.053, 1: 12.047}
+    Lndon 1 {0: 0.05, 1: 3.05}
+    Paris, France 0 {0: 16.547, 1: 0.053}
+    Pariss 0 {0: 4.549, 1: 0.051}
+
```
MaxHalford: One last thing: could you add an example that puts a `GapEncoder` in a classification pipeline?

MaxHalford (approved):
Great stuff! It's great work, and a useful addition. Just asked for one thing in the docstring.

river/preprocessing/gap_encoder.py
Fazel94: Thank you for your kindness, will do that, next week.

MaxHalford:
@Fazel94 I have an ask: could you add a benchmark to CodSpeed? I would like us to do that systematically from now on, when adding new estimators.

Fazel94:
Of course, I'll be on it the next week.

MaxHalford:
The failing unit test is flaky and unrelated, I'm ok merging. Good work!

Fazel94:
Do you want me to take a look at the flaky unit test for another pr?

MaxHalford:
I guess the next one who sees it can fix it :)

PR #1945, opened 2026-07-07
fix remaining deprecated abstractproperty decorator

mugulmd:
`abc.abstractproperty` is deprecated since Python 3.3, in favor of stacked decorators:
```
@property
@abstractmethod
def ...
```
This pattern is enforced everywhere in the codebase except one tiny omission, this PR fixes it.


MaxHalford:
Thanks! Easy enough to merge from my phone :)

Issue #1944, opened 2026-07-06
Strict typing all modules


mugulmd:
## Description

This is an issue to keep track of the strict typing status of each module.

Strict typing has the potential to bring great quality of life for both maintainers and library users: a wide range of bugs can be caught earlier, stronger IDE integrations, clearer API contracts and documentation, etc.

Strict typing all of River's modules can be achieved incrementally, with one PR per module, and with the corresponding updates in the last `pyproject.toml` section:
```
[[tool.mypy.overrides]]
# Disable strict mode for all non fully-typed modules
...
```

If any strict typing update ends up changing something in the API, this should be clearly mentioned in the PR description.

---

## Status

- [X] _river_rust
- [ ] active
- [ ] anomaly
- [ ] bandit
- [X] base
- [ ] checks
- [ ] cluster
- [ ] compat
- [ ] compose
- [x] conf
- [ ] covariance
- [ ] datasets
- [ ] drift
- [ ] ensemble
- [ ] evaluate
- [ ] facto
- [ ] feature_extraction
- [ ] feature_selection
- [ ] forest
- [ ] imblearn
- [ ] linear_model
- [ ] metrics
- [X] misc
- [ ] model_selection
- [ ] multiclass
- [ ] multioutput
- [ ] naive_bayes
- [ ] neighbors
- [x] optim
- [ ] preprocessing
- [ ] proba
- [ ] reco
- [ ] rules
- [ ] sketch
- [x] stats
- [ ] stream
- [ ] time_series
- [ ] tree
- [X] utils

e10e3:
Hello @mugulmd,
This is not the first time that someone proposes to contribute on this topic, and I'm happy to see new people join the effort!

You can see my initial proposal, which is similar to yours, in discussion #1592; I also made a few PRs to complete the first modules (e.g. #1596, #1719, #1719).

The task before us is still great, but hopefully by starting from the foundations and working bit by bit we'll cover all the library.

I checked few boxes in your list for the modules that are type-complete.

MaxHalford:
Thank you @e10e3 if you can coordinate with @mugulmd!

While we're at it, would it worth considering other tools than mypy? I am personally quite fond of ty, as it seems more opinionated and strict out of the box. Also a bit faster, but to be honest negligible at our scale. Any thoughts on the tooling?

I trust the two of you :)

mugulmd:
Hey @e10e3 I'm glad to see you here!
[...]

PR #1950, opened 2026-07-12
Strict typing: stubs for rust bindings

mugulmd:
## Description

> This PR does not impact the library API nor its runtime behavior in any way: it only adds type hints for the internal `_river_rust` package.

Related issue:
https://github.com/online-ml/river/issues/1944

Relevant PyO3 documentation pages:
- https://pyo3.rs/main/python-typing-hints
- https://pyo3.rs/main/type-stub

Typing hints for the `_river_rust` package are essential to enable full strict typing of the library. In addition they provide an improved experience for maintainers, as they won't need to dive into Rust code to figure out the input/output types of the functions coming from the binding.

As stated in the PyO3 docs, the best solution to achieve this is to write and maintain stub files manually (their automatic generation by PyO3 is still under development).

This PR adds stub files matching the current state of `_river_rust`, and provides a simple test to ensure that the stubs are kept in sync with the actual bindings.

## Future work

A non-negligible portion of type annotations in the stubs are defined as `_typeshed.Incomplete` (docs: https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs), which serves as a placeholder until more accurate typing is provided.

One compelling example is the `VectorDict` class, where we could use the type parameters `KT` and `KV` for accurately typing the methods.

## AI disclaimer

A coding agent was used to generate the very 1st draft code (stubs of `_river_rust.stats` and sync test).
The AI-generated code has been thoroughly reviewed, edited where corrections were needed, and complemented manually with everything else in this PR.


MaxHalford:
Cheers.

PR #1951, opened 2026-07-16
Fix Metrics collection dropping the sample weight

chuenchen309:
## Problem

A `metrics.base.Metrics` collection — the object you get from `metric_a + metric_b` — accepts a sample weight `w` in `update`, but never forwards it to its child metrics. The weight is silently treated as `1`.

`Metrics.revert` *does* pass `w` to the children, so the two methods are asymmetric. Two consequences:

```python
from river import metrics

coll = metrics.MAE() + metrics.MSE()
coll.update(0, 0, w=1)
coll.update(0, 10, w=99)
coll[0].get()          # 5.0   ❌  (weight dropped)

standalone = metrics.MAE()
standalone.update(0, 0, w=1)
standalone.update(0, 10, w=99)
standalone.get()       # 9.9   ✅  (weight honored)
```

1. A weighted metric inside a collection reports the wrong value.
2. Because `update` adds weight `1` while `revert` removes the real weight, an `update(w)` / `revert(w)` pair no longer cancels — it can raise `ValueError("Cannot go below 0")` from the underlying running statistic.

Note `Metrics.works_with_weights` already advertises weight support (`all(child.works_with_weights ...)`), so the collection claims a capability its `update` didn't honor.

## Fix

Forward `w` in the three `update` call sites, exactly mirroring `revert`.

## Test plan

Added `test_metrics_collection_forwards_sample_weight` to `river/metrics/test_metrics.py`: a weighted collection now matches a standalone metric (`9.9`), and `update`/`revert` cancel back to empty state.

- On `main` the new test fails (`assert 5.0 == 9.9`); with this change it passes.
- `pytest river/metrics/` → 274 passed (including doctests).
- `ruff check` / `ruff format --check` clean on both files.
- Added a `## metrics` entry to `docs/releases/unreleased.md`.

---
Authored with assistance from Claude Code (Anthropic); the bug, fix, and test were reviewed and verified locally by the contributor.


MaxHalford:
Great catch!

PR #1956, opened 2026-07-20
Fix CohenKappa ignoring sample weights

chuenchen309:
`CohenKappa.get()` divides the observed agreement (`p0`) and the two expected-agreement estimators by `self.cm.n_samples` — the raw, *unweighted* observation count — while the numerators (`total_true_positives`, `sum_row[c]`, `sum_col[c]`) are all *weighted* sums. When any sample weight differs from 1, the denominator no longer matches the numerator and the returned Kappa is wrong. With unit weights `n_samples == total_weight`, which is why the existing tests never caught it.

`CohenKappa` inherits `works_with_weights == True` (the base default) and is driven through the documented `update(y_true, y_pred, w=...)` weight entry point, so it *advertises* weight support while computing the wrong number. The sibling metrics that genuinely can't handle weights (`mutual_info`, `vbeta`, `rand`, `fowlkes_mallows`) all override `works_with_weights = False`; and `Accuracy.get()` — which the in-line comment says CohenKappa's `p0` is "same as" — already divides by `total_weight`.

Fix: divide the three agreement terms in `get()` by `self.cm.total_weight` instead of `self.cm.n_samples`.

Verification (re-runnable from the diff):

- New test `test_cohen_kappa_supports_sample_weight` asserts equality with `sklearn.metrics.cohen_kappa_score(..., sample_weight=...)` and that the weighted result differs from the unweighted one (proving the weight is actually consumed). It is red before the change and green after.
- Full `river/metrics/` suite + doctests: 278 passed.
- Oracle cross-checked against `sklearn` on the fixed case, the unweighted case, and 500 random weighted fuzz cases (0 mismatches; degenerate single-label windows yield `nan` in both and are excluded).

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro, wrote the tests, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -77,6 +77,7 @@
 
 - Fixed `metrics.base.Metrics` (a metrics collection, built via `metric_a + metric_b`) dropping the sample weight `w`: `update` now forwards `w` to each child metric, so weighted metrics report correct values inside a collection and `update`/`revert` cancel exactly. Previously `revert` applied the weight but `update` ignored it.
 - Fixed `metrics.BalancedAccuracy` deflating its score when a label appears in the predictions but never as a ground-truth label. Such a class has no support, so its recall is undefined and must be excluded from the per-class average; previously it was counted as `0` and inflated the denominator, disagreeing with `sklearn.metrics.balanced_accuracy_score`. `BalancedAccuracy` is now covered by the scikit-learn equivalence test.
+- Fixed bug where `metrics.CohenKappa` did not take into account sample weights.
 
 ## multiclass
 
--- a/river/metrics/kappa.py
+++ b/river/metrics/kappa.py
@@ -50,15 +50,15 @@ class CohenKappa(metrics.base.MultiClassMetric):
 
     def get(self):
         try:
-            p0 = self.cm.total_true_positives / self.cm.n_samples  # same as accuracy
+            p0 = self.cm.total_true_positives / self.cm.total_weight  # same as accuracy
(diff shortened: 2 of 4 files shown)


MaxHalford:
Hey there @chuenchen309! May I ask you to open issues instead of immediately proceed to a fix? I appreciate you discovering these bugs, but I would like to dissociate the bug report from the correction. Would that be acceptable for you?

chuenchen309:
Yes, that's completely acceptable — and sorry both for the delay in answering and for the shape of the run that prompted it.

Your reasoning makes sense to me: with the report and the correction in one place, you can't weigh whether something is a bug independently of whether you like the patch, and a batch of them arriving at once makes that worse rather than better. I'll open an issue first from here on.

For this one the description and the fix are already in the same place, so I don't want to manufacture noise by splitting it after the fact — tell me which you'd prefer:

- I close this PR and open an issue with just the behaviour, leaving the fix to whoever picks it up; or
- it stays as a PR and you judge it as-is.

Either is fine by me. It has also gone conflicting with `main` since July, so it needs a rebase before it's mergeable in any case — happy to do that if you want to keep it.

*AI-assisted with agentic coding tools; all changes were reviewed and tested by me.*

MaxHalford:
Yes no worries, my advice was for future PRs. We can iterate on this one and merge it.

PR #1957, opened 2026-07-20
Strict typing: stats sub-package

mugulmd:
## Description

Related issue:
https://github.com/online-ml/river/issues/1944

Implement strict typing for the `river.stats` package as it only depends on already strictly typed packages.

Since statistics can ingest and produce various types of data, generics have been used in base classes to ensure proper typing (with sensible defaults to avoid burdening developers).

### Easier to ask for permission than forgiveness

One notable API change has been introduced: in order to fix several inconsistencies in the code we propose that the `get()` method of statistics does not return `None` when not enough samples were ingested, but instead raises a custom `NotEnoughSamples` exception.

We also provide a `get_or_none()` method for conveniency, so that callers depending on the "none" behavior can easily adapt without adding `try / except` blocks in their own code.


river/base/errors.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+
+class RiverError(Exception):
```
e10e3: Using an exception is a nice way of dealing with the problem of uninitialised statistics.

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
+from . import errors
 
+__all__ = ["Bivariate", "Link", "RollingUnivariate", "Statistic", "Univariate"]
 
-class Statistic(abc.ABC, base.Base):
+# The type of the values a statistic produces. Defaults to float so that existing
+# annotations such as `stats.base.Statistic` keep meaning "a float statistic".
+R = TypeVar("R", default=float)
+# Another type for produced values, used for chaining statistics.
+S = TypeVar("S", default=float)
```
e10e3: The TypeVars should be private because they shouldn't be used outside the module.

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
+# The type of the values a statistic observes.
+X = TypeVar("X", default=float)
+# Another type of the values a (bivariate) statistic observes.
+Y = TypeVar("Y", default=float)
```
e10e3: Do the letters for the TypeVars have a meaning? Common practice is to use `T` to mean "type" or and abbreviation like `KT`/`VT` in dictionaries for "key/value type", so I'm wondering what they stand for here.

Using more explicit names would make it easier to understand what they mean and how the types flow. Otherwise, when you have and expression that reads Link[X,R,S], how do you know what you're supported to set in each field?

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
+
+
+class Statistic(abc.ABC, base.Base, Generic[R]):
```
e10e3: I'm not convinced about making the Statistics generic. Do you have an example of the problem it solves?

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
             value = None
         fmt_value = None if value is None else f"{value:{self._fmt}}".rstrip("0")
         return f"{self.__class__.__name__}: {fmt_value}"
 
-    def __str__(self):
+    def __str__(self) -> str:
         return repr(self)
 
-    def __gt__(self, other):
+    def __gt__(self: Statistic[float], other: Statistic[float]) -> bool:
```
e10e3: Because Statistic is generic, what about the case where `R` is not a float? They can't be compared?

Also nit: the `self` parameter is usually not annotated because its type is known, but I guess this doesn't do harm.

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self.__class__.__name__.lower()
 
-    def __or__(self, other):
-        from .link import Link
-
+    def __or__(self, other: Univariate[R, S]) -> Univariate[X, S]:
```
e10e3: If the method returns a `Link` object, shouldn't it be annotated as such?

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
         return Link(left=self, right=other)
 
 
-class RollingUnivariate(Univariate):
+class Link(Univariate[X, S], Generic[X, R, S]):
```
e10e3: What is the reasoning behind moving `Link` from its file to here? Is there an advantage?

river/stats/base.py
```diff
@@ -1,72 +1,171 @@
...
 
-__all__ = ["Univariate", "Bivariate"]
+from . import errors
 
+__all__ = ["Bivariate", "Link", "RollingUnivariate", "Statistic", "Univariate"]
 
-class Statistic(abc.ABC, base.Base):
+# The type of the values a statistic produces. Defaults to float so that existing
+# annotations such as `stats.base.Statistic` keep meaning "a float statistic".
+R = TypeVar("R", default=float)
```
e10e3: The `default` keyword was introduced in Python 3.13, it will fail in Python 3.11 and 3.12.

river/stats/chi_squared.py
```diff
@@ -55,24 +55,24 @@ class ChiSquared(stats.base.Bivariate):
     """
 
     def __init__(self) -> None:
-        self.counts: collections.defaultdict[typing.Any, collections.Counter] = (
+        self.counts: collections.defaultdict[Hashable, collections.Counter[Hashable]] = (
```
e10e3, line 58: Mapping types have an invariant key type (this includes all dicts and dict-based classes like `defaultdict` and `Counter`). This means that only exactly that type is allowed as a key, and all its subclasses are forbidden.

In this case, because `Hashable` is not a concrete  type, the type checker will cause errors if you try using any object as a key, such a a string.

AFAIK, if you want more than one type as keys, you should use `Any`.

river/stats/cov.py
```diff
@@ -66,50 +66,52 @@ class Cov(stats.base.Bivariate):
 
     """
 
-    def __init__(self, ddof=1):
-        self.ddof = ddof
-        self.mean_x = stats.Mean()
-        self.mean_y = stats.Mean()
-        self.cov = 0
+    def __init__(self, ddof: float = 1) -> None:
```
e10e3: Shouldn't the degree of freedom (`ddof`) be an integer? Unless we want to do [funky maths](https://en.wikipedia.org/wiki/Fractional_calculus).

river/stats/kolmogorov_smirnov.py
```diff
@@ -26,20 +27,20 @@ class _Node:
...
-        self.priority = _random()
-        self.size = 1
-        self.height = 1
-        self.lazy = 0
-        self.max_value = value
-        self.min_value = value
-        self.left = None
-        self.right = None
+    def __init__(self, key: typing.Any, value: float = 0) -> None:
+        self.key: typing.Any = key
```
e10e3: Since the class `_Node` is private and is only used here, can't we be more specific on the key type, and use the actual type used in the implementation?
```suggestion
        self.key: tuple[float, int] = key
```

river/stats/kolmogorov_smirnov.py
```diff
@@ -286,13 +294,13 @@ def get(self):
             raise ValueError(f"Unknown statistic {self.statistic}, expected one of: ks, kuiper")
 
     @staticmethod
-    def _ca(p_value):
-        return (-0.5 * math.log(p_value)) ** 0.5
+    def _ca(p_value: float) -> float:
+        return float((-0.5 * math.log(p_value)) ** 0.5)
```
e10e3: Should we use math.sqrt()? It's faster and has the correct typing (it takes a float and give a float, whereas **0.5 can return complex numbers).
```suggestion
        return math.sqrt(-0.5 * math.log(p_value))
```

river/stats/mode.py
```diff
@@ -114,15 +113,16 @@ class RollingMode(stats.base.RollingUnivariate):
...
+    def __init__(self, window_size: int) -> None:
+        self.window: collections.deque[Hashable] = collections.deque(maxlen=window_size)
+        self.counts: collections.defaultdict[Hashable, int] = collections.defaultdict(int)
+        self.window_size_value: int = window_size
 
     @property
-    def window_size(self):
-        return self.window.maxlen
+    def window_size(self) -> int:
+        return self.window_size_value
```
e10e3: What is the advantage of storing this value in a variable over using `window.maxlen`?

river/stats/test_chi_squared.py
```diff
@@ -141,7 +141,9 @@ def test_chi_squared_rolling():
 
     for x, y in data[:4]:
         chi2.update(x, y)
-    assert math.isclose(chi2.get(), 4.0)
+    # `utils.Rolling` proxies attribute access via `__getattr__ -> object`, so mypy cannot
+    # know that `get` is callable here.
+    assert math.isclose(chi2.get(), 4.0)  # type: ignore[operator]
```
e10e3: Because the stats is wrapped in `Rolling`, you can access the base object though `.obj`
```suggestion
    assert math.isclose(chi2.obj.get(), 4.0)
```

river/stats/test_chi_squared.py
```diff
@@ -156,10 +158,10 @@ def test_chi_squared_rolling():
     #      + (0 - 1.0)^2 / 1.0 + (2 - 1.0)^2 / 1.0
     #      + (1 - 0.5)^2 / 0.5 + (0 - 0.5)^2 / 0.5
     #      = 0.5 + 0.5 + 1.0 + 1.0 + 0.5 + 0.5 = 4.0
-    assert math.isclose(chi2.get(), 4.0)
+    assert math.isclose(chi2.get(), 4.0)  # type: ignore[operator]
```
e10e3: Same
```suggestion
    assert math.isclose(chi2.obj.get(), 4.0)
```

river/stats/test_stats.py
```diff
@@ -39,7 +43,7 @@ def load_stats():
 
 
 @pytest.mark.parametrize("stat", load_stats(), ids=lambda stat: stat.__class__.__name__)
-def test_pickling(stat):
+def test_pickling(stat: typing.Any) -> None:
```
e10e3: You can add a more specific type to most of these tests in the file, this would help ensure the annotations in the module are correct.
```suggestion
def test_pickling(stat: stats.base.Statistic) -> None:
```

river/stats/test_stats.py
```diff
@@ -80,7 +86,7 @@ def test_repr_with_no_updates(stat):
         (stats.Var(), functools.partial(np.var, ddof=1)),
     ],
 )
-def test_univariate(stat, func):
+def test_univariate(stat: typing.Any, func: typing.Callable[..., typing.Any]) -> None:
```
e10e3: Another example:
```suggestion
def test_univariate(stat: stats.base.Univariate, func: typing.Callable[[Iterable[float]], float]) -> None:
```

e10e3 (commented):
Good job!

I like your proposition of using an Exception in `get` when the stat is not sufficiently initialised. This said, I'm a bit torn because it means changing the API and requires introducing a secondary `get_or_none` that gets rather widely used.

I'm not sold on making the base classes generic. I don't see what it adds, do you have an example where it makes a difference? I'll concede that it's purely internal, it doesn't make a difference to end users.

I see that you added all annotations in the `__init__` methods, including the optional ones. I tend to only add the required ones, but your approach doesn't harm.

You should also add a changelog note, including the breaking changes.

Also, in your cover letter you say "we", who is this uncredited contributor?

mugulmd:
Hey @e10e3 thanks for the review!

I'll tackle code comments one-by-one later when I get some time to update the code as well.

> I like your proposition of using an Exception in `get` when the stat is not sufficiently initialised. This said, I'm a bit torn because it means changing the API and requires introducing a secondary `get_or_none` that gets rather widely used.

Completely agreed: raising custom errors for ill-defined situations is a common and efficient pattern in Python (at least in my experience) and it fits well here, however I'm not a big fan of changing the API in a *typing* PR (which should ideally not affect API nor runtime behavior).
I'm guessing this is not the only place where a custom error would be useful, so what we can do instead is the following: go back to `get() -> float | None`, continue working on typing without affecting the API, and prepare a follow-up work to introduce these custom errors (we will get a better picture of the amount of work involved once we start typing other packages I think).
Wdyt?

> I'm not sold on making the base classes generic. I don't see what it adds, do you have an example where it makes a difference? I'll concede that it's purely internal, it doesn't make a difference to end users.

This was a way to properly handle the fact that various statistics can ingest / produce various data types: `Count` ingests nothing and produces `int`, `Mean` ingests and produces `float` data, `Entropy` ingests any kind of hashable data, `Mode` ingests and produces hashable data, etc.
If you have another (simpler) mechanism for that then I am definitely interested :)

> I see that you added all annotations in the `__init__` methods, including the optional ones. I tend to only add the required ones, but your approach doesn't harm.

For some reason it looked like `mypy` needed these annotations, which is why I added them, but I didn't investigate much further so I can double check if they are actually required or not.

> Also, in your cover letter you say "we", who is this uncredited contributor?

Ahaha there is nobody else, just me, myself and I, this is a just a habit: I consider PRs to be a group effort (at least the code writer and the reviewer, and in a larger sense all people involved in the tool) and the PR description to be a message addressed from this group to any future visitor -> hence the *we* :)

river/stats/base.py
mugulmd: To solve a circular import, since `Link` depends on `Univariate` :)
The old code was dynamically importing `Link` in the method where it is used, which is a quite dangerous pattern in my experience, so I preferred to have `Link` in here (which somehow makes sense because it is the only place where it is used -> the API is quite focused on chaining things with `|` instead of explicitly declaring chaining objects).

river/stats/base.py
mugulmd: I had no idea, thanks! I'll definitely change this one then, do you know if there is any "equivalent" mechanism before Python 3.13?

river/stats/cov.py
mugulmd: Well spotted thanks :)

river/stats/mode.py
mugulmd: `window.maxlen` has type `int | None` (because the queue can be unbounded) and there is no "infinite integer" in Python to easily replace the `None` value.
However since we explicitly construct the queue with a `maxlen` I agree casting `window.maxlen` to `int` is acceptable in this situation :)

river/stats/base.py
mugulmd: Simplfied and renamed more explicitly :)

river/stats/base.py
mugulmd: Not all statistics ingest and produce the same types: most of the time it's float in -> float out, but it can also be integers (e.g `Count`) and anything hashable (e.g `Entropy` and `Mode`).
I have simplified the use of generics to only keep type vars for the "return type" (the values produced by `get()`): using `Any` for the inputs of `update(...)` works fine and can be specialized in subclasses, but this is not true for the return type of `get()` where using `Any` would make use loose a lot of typing accuracy throughout the codebase.

river/stats/base.py
mugulmd: IMO this is ill-defined interface and probably should not exist, but since it is there and part of the API I decided to keep it there and just use `cast(...)` for now.

river/stats/base.py
mugulmd: I used `type_extensions` for backward compatibility :)

river/stats/chi_squared.py
mugulmd: I don't think that's true, invariance governs *assignability of the container*, not what you're allowed to put in it:
```python
counter: collections.Counter[Hashable] = collections.Counter()
counter["a"] += 1        # fine
counter[42] += 1         # fine
counter[("a", 1)] += 1   # fine

str_keyed: dict[str, int] = {"a": 1}
wants_hashable: dict[Hashable, int] = str_keyed   # error: incompatible types
```
I think this is working fine, and `Hashable` is more accurate than `Any`, so if it works I definitely vote we keep it in :)
However if I missed (or misunderstood) anything don't hesitate to correct me of course 😅

mugulmd:
@e10e3 I'm all good for a re-review when you get some spare time 😄

river/stats/chi_squared.py
e10e3: Yes, you're right, the invariance actually only covers the assignment dicts and related stuff like function parameters.
I thought the incompatibility was larger and extended to updating the dict as well. Good to know, I can be less paranoid.

river/stats/kolmogorov_smirnov.py
```diff
@@ -286,13 +295,13 @@ def get(self):
...
 
-    def _test_ks_threshold(self, ca):
+    def _test_ks_threshold(self, ca: float) -> bool:
         """
         Test whether the reference and sliding window follows the same or different probability distribution.
         This test will return `True` if we **reject** the null hypothesis that
         the two windows follow the same distribution.
         """
-        return self.get() > ca * (2 * self.n_samples / self.n_samples**2) ** 0.5
+        return self.get() > ca * float((2 * self.n_samples / self.n_samples**2) ** 0.5)
```
e10e3: Also a case where `math.sqrt` can be used (sorry, I did not flag all the applicable spots in the first review)

river/stats/pearson.py
```diff
@@ -65,33 +65,33 @@ class PearsonCorr(stats.base.Bivariate):
...
 
-    def __init__(self, ddof=1):
+    def __init__(self, ddof: int = 1) -> None:
         self.var_x = stats.Var(ddof=ddof)
         self.var_y = stats.Var(ddof=ddof)
         self.cov_xy = stats.Cov(ddof=ddof)
 
     @property
-    def ddof(self):
+    def ddof(self) -> float:
```
e10e3: ddof is an int now

river/stats/sem.py
```diff
@@ -56,8 +56,9 @@ class SEM(var.Var):
 
     """
 
-    def get(self):
+    # TODO: drop the ignore once get() raises stats.NotEnoughSamples instead of returning None
+    def get(self) -> float | None:  # type: ignore[override]
         try:
-            return (super().get() / self.mean.n) ** 0.5
+            return float((super().get() / self.mean.n) ** 0.5)
```
e10e3: Also a case where `math.sqrt` can be used

river/stats/test_chi_squared.py
```diff
@@ -141,7 +141,9 @@ def test_chi_squared_rolling():
 
     for x, y in data[:4]:
         chi2.update(x, y)
-    assert math.isclose(chi2.get(), 4.0)
+    # `utils.Rolling` proxies attribute access via `__getattr__ -> object`, so mypy cannot
+    # know that `get` is callable here.
```
e10e3: I believe this comment is not accurate anymore now that we access the stat though `.obj`

river/stats/base.py
```diff
@@ -1,72 +1,166 @@
...
+        return cast(float, self.get()) > cast(float, other.get())
 
 
-class Univariate(Statistic):
+class Univariate(Statistic[_OutT]):
     """A univariate statistic measures a property of a variable."""
 
     @abc.abstractmethod
-    def update(self, x: numbers.Number) -> None:
+    def update(self, x: Any) -> None:
```
e10e3: I believe `float` is correct here, the other input types are always supertypes of it.
```suggestion
    def update(self, x: float) -> None:
```

e10e3 (changes requested):
We're in a pretty good shape now.

My comments are only small remarks.

The biggest thing is about the input type of `Univariate.update`, which is also an answer to your remark about generic stats.

> This was a way to properly handle the fact that various statistics can ingest / produce various data types: Count ingests nothing and produces int, Mean ingests and produces float data, Entropy ingests any kind of hashable data, Mode ingests and produces hashable data, etc.
If you have another (simpler) mechanism for that then I am definitely interested :)

The input in the stats is always a `float` or a supertype of `float`: `Count` ingests `Any`, `Mode` ingests `Hashable`, etc. This respects Liskov's substitution principle: daughter classes can have wider input than the mother.
The place you are right about is the output type, since by default daughters can only be stricter that the mother. Making the base class generic on the output solves the problem.

Putting `Any` for the input of `Univariate.update` is not wrong, but we can make it more precise without entering in conflict with the substitution principle.
The only problem is in `Link` where there is the possibility of feeding a non-float output to a float-only stat, but we can probably put an ignore on it because it would trigger a runtime error anyway.

Don't forget to add a changelog note to `unreleased.md` :)

(You didn't really need to remove the extra annotations by the way, they didn't interfere with anything, I made the remark to save you some work next time.)

river/stats/kolmogorov_smirnov.py
mugulmd: It's my fault I should have grep'ed, I got lazy 😅

river/stats/test_chi_squared.py
mugulmd: Good spot thanks!

river/stats/base.py
mugulmd: It's slightly better in this place but also slightly worse in `Link` (because `Link` *must* accept `Any` as input and therefore you end up with an awkward `# type: ignore[arg-type]` comment) but I agree it's not the end of the world so I'm going with your solution 👍 

Overall I think using generics for the input type as well would solve the problem properly even though it's more verbose, but we can definitely work on that later -> the vast majority of typing work in `stats` in done, and it will unlock other sub-packages, let's not over-focus on tiny details :)

Issue #1914, opened 2026-06-23
Drift module roadmap: broaden detector coverage + build an evaluation harness (6-month focus)


MaxHalford:
## Summary

Proposal to make `river.drift` a focused area of development for the next ~6 months: broaden detector coverage across the three main families (online statistical CPD, supervised/error-stream, and unsupervised/data drift), and — just as importantly — build a **drift-detector evaluation harness** so we can measure and defend the module's quality.

The trigger is a recent empirical study that benchmarks River head-to-head against other tools, and reveals concrete gaps.

## Motivation: an external benchmark shows where we fall short

Besbes et al., *"Exploring Statistical Change Point Detection Techniques for Performance Anomaly Detection at Mozilla"* (2026, [arXiv:2606.18377](https://arxiv.org/abs/2606.18377)) benchmarks **25 change-point/drift detection methods + 15 ensemble strategies** on a new ground-truth dataset of 174 Mozilla performance time series (released on [Zenodo, doi:10.5281/zenodo.20382322](https://doi.org/10.5281/zenodo.20382322)), built by extending [TCPDBench](https://github.com/alan-turing-institute/TCPDBench).

River is one of the libraries they use. The telling detail is in their method table:

- **From `river`:** ADWIN, KSWIN, Page-Hinkley, DummyDriftDetector
- **Implemented "Custom" by the authors:** CUSUM, EWMA, Shewhart, SPRT
- **Pulled from other libraries:** BOCPD (`ocp`), Cramér-von Mises windowing (`alibi-detect`)

In other words: the authors reached for River first, then had to leave it for **six** of the online detectors they wanted. Findings worth noting:

- **BOCPD** (Bayesian Online Change Point Detection) was the **best online method out of the box** (F1 0.633, recall 0.803) — and it's absent from the whole River ecosystem.
- River's ADWIN/KSWIN/Page-Hinkley scored low on their retrospective CPD task (default F1 ~0.27–0.29). This is partly an evaluation-regime artefact (single hyperparameter grid; these are streaming detectors tuned for reaction speed, not retrospective F1) — but it underlines that **we have no harness of our own to characterize our detectors' behavior**, so we can't contextualize or defend results like these.

## Current state of `river.drift`

- **Online, continuous input:** `ADWIN`, `KSWIN`, `PageHinkley` (+ `DummyDriftDetector`, `NoDrift`)
[...]

PR #1963, opened 2026-07-27
feat(drift): evaluation metrics for drift detectors

jevwithwind:
Hi, as discussed in #1914, this is a prototype evaluation tool for drift detection at 'river/drift/evaluate.py'. It's only a draft and needs help on the API side.

I used claude code to write the code and tests, while I directed it and reviewed its output before putting it here.

It's my first time contributing here and please let me know what you make of this.

### DIFF
--- a/river/drift/__init__.py
+++ b/river/drift/__init__.py
@@ -9,7 +9,7 @@ the true positives while keeping the number of false positives to a minimum.
 
 from __future__ import annotations
 
-from . import binary, datasets
+from . import binary, datasets, evaluate
 from .adwin import ADWIN
 from .dummy import DummyDriftDetector
 from .kswin import KSWIN
@@ -20,6 +20,7 @@ from .retrain import DriftRetrainingClassifier
 __all__ = [
     "binary",
     "datasets",
+    "evaluate",
     "ADWIN",
     "DriftRetrainingClassifier",
     "DummyDriftDetector",
--- a/river/drift/evaluate.py
+++ b/river/drift/evaluate.py
@@ -1,0 +1,498 @@
+from __future__ import annotations
+
+import collections
+import math
+import typing
+
+__all__ = [
+    "Matching",
+    "detection_delay",
+    "episode_recall",
+    "f1",
+    "false_alarm_rate",
+    "match",
+    "mean_time_between_false_alarms",
+    "mean_time_ratio",
+    "missed_detection_rate",
+    "normalized_detection_time",
+    "precision",
+    "recall",
+    "report",
+    "score",
+]
+
+Drift = int | tuple[int, int]
+
+Matching = collections.namedtuple("Matching", ["hits", "false_alarms"])
+
+
+def _windows(drifts: list[Drift], delta_max: int, delta_pre: int) -> list[tuple[int, int, int]]:
+    if delta_max <= 0:
+        raise ValueError("delta_max must be strictly positive")
+    if delta_pre < 0:
+        raise ValueError("delta_pre must be positive")
+
+    windows = []
+    for drift in drifts:
+        start, end = drift if isinstance(drift, tuple) else (drift, drift)
+        if end < start:
+            raise ValueError(f"drift {drift} ends before it starts")
+        windows.append((start, start - delta_pre, end + delta_max))
+    return windows
+
+
+def match(
+    alarms: typing.Iterable[int],
+    drifts: typing.Iterable[Drift],
+    *,
+    delta_max: int,
+    delta_pre: int = 0,
+) -> Matching:
+    """Assign each alarm to a drift, or to the false alarm pile.
+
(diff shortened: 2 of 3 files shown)


jevwithwind:
For metrics definitions, I used:
1. Cerqueira et al. 2026 (preprint) since two of the authors are on the River team and it implements the protocol in CapyMOA, so claude matched it.
2. Bifet 2017 (doi 10.1007/978-3-319-59060-8_64), which is where the four metrics in #1914 come from.
3. Lukats et al. (doi 10.1007/s41060-024-00620-y), recommended by @e10e3.

jevwithwind:
To continue, I want to know: does this belong in river/drift/ or river/evaluate/?
And should delta_max should keep having no default following Cerqueira et al. 2026?
Also any clue on how to proceed with the API side.

e10e3:
Thank you for the contribution!

If all additions here are metrics, maybe they should go to the `metrics` module, or perhaps one of its submodule?
If you chose this option, I shall note that all metrics (and much of River's API) are implemented as classes, whereas yours are functions. To integrate them in the `metrics`, you will need to turn them into classes.

`metrics` already has metrics with the same name as yours (F1, Precision, Recall…), is there no way to reuse them?

The [article from Cerqueira _et al._](https://arxiv.org/abs/2606.07789) that you mention says that $\delta_\mathrm{pre}$ should be 0 only for synthetic streams because of how they're constructed, so 0 is not a good value for all possible streams. If you _need_ to have defaults for $\delta_\mathrm{pre}$ and $\delta_\mathrm{max}$, you could consider what other papers use. [Bifet's paper](https://link.springer.com/chapter/10.1007/978-3-319-59060-8_64) appears to use $\delta_\mathrm{pre} = 0$ and $\delta_\mathrm{max} = +\infty$ (this shows that both papers have differing evaluation methodologies).

You implementation relies on inputing `Iterable`s of results. Does this work well with River's iterative behaviour? We don't want to need to store all of a model's predictions to evaluate it. The current metrics work the following may:
```py
for x, y in dataset:
    y_pred = model.predict_one(x)
    metric.update(y, y_pred)
   model.learn_one(x, y)
```
Where, at every step, we update the metric with the model's output, so we only need to store the aggregate and not the full output series of the model. This also allows us to get a reading of the metric at any time without additional computation.
This way of  working is central to River, and we want to preserve it.

jevwithwind:
Thank you for the reply! Yes, metrics should be in classes not functions, and putting them in `metrics` makes sense. *δpre* default to 0 should be removed too.

Regarding the `Iterable`s, understood, the code is fixable to fit the existing frame. The drift schedule is known in advance so each step can be classified as it arrives.

Regarding the existing metrics, F1, precision, recall can be reused but episode recall counts windows not alarms. NDT and the rest are not classification metrics. To reuse them, I would add a matching step (alarm stream -> drift schedule -> TP/FP/FN under a tolerance window) before reusing the existing arithmetic.

Claude also made a mistake with Bifet's paper on how it counts drifts. I will start with the rework shortly.

mateenali66:
the metrics need a chance level inside the same object, not computed alongside them by whoever reads the results.

I ran the four metrics on this branch against detectors that ignore the data, posted on the roadmap issue at https://github.com/online-ml/river/issues/1914#issuecomment-5409322994. a detector that alarms at every 100th index scores on all four. the numbers read as performance and are a property of the alarm rate.

the reason it matters here rather than later: precision and recall against a known drift schedule both move with how many alarms a detector emits, so with `delta_max` fixed, two detectors at different alarm rates are not comparable on the raw numbers. a harness exists to make them comparable, so the floor belongs in the output.

on e10e3's point about classes rather than functions, that helps here. if each metric is a class it can expose the chance level for the schedule it was given, and the report can carry score and floor together instead of the floor being an extra thing to remember.

related, `check_roc_auc` in the estimator checks currently passes for a detector returning a constant score, which I opened separately at https://github.com/online-ml/river/issues/1992. same shape of problem, different surface.

I can send the closed form for the uniform-alarm baseline and a couple of degenerate fixtures, either as review comments here or as a follow-up PR on top of this branch. not looking to take this over, it is your PR.

nikolas-sapa:
Independent confirmation of @mateenali66's finding — I ran the evaluate module from this PR head on a different generator (Gaussian, abrupt shifts, seed fixed) and see the same failure mode: an "alarm every 100 steps" detector scores F1 0.049 at 5 drifts but 0.413 at 50 drifts on the same 20k samples — the 8.4x gap is drift density, nothing else. The row that convinced me: at 50 drifts with small shifts, ADWIN at defaults scores F1 0.000 while always-fire scores 0.403.

I'd also +1 scoring the same metrics on DummyDriftDetector inside report() and printing it next to the headline numbers — it costs nothing (river already ships it) and would make a below-baseline result visible instead of inferred. Happy to share the ~80-line repro script if useful.

PR #1965, opened 2026-07-28
feat(compat): dataframe-agnostic scikit-learn wrappers via narwhals

FBruzzesi:
# Description

This PR is following two streams of work: adoption of narwhals (#1919 ) and strict typing (#1944) .

## Narwhals adoptiong

Routes the mini-batch methods of `SKL2RiverRegressor` and `SKL2RiverClassifier` through narwhals, so `learn_many` / `predict_many` / `predict_proba_many` accept any eager dataframe (pandas, polars, pyarrow, ...) and return output in the same backend. Previously they hard-required pandas via a runtime `import_pandas()`.

`_align_df` becomes `_align_frame`: it reorders a batch's columns to the order seen in the first batch (scikit-learn identifies features by position), and now raises a `ValueError` naming the missing features instead of surfacing a pandas `KeyError`. 

## Strict typing

I allowed myself some creative freedom. Both classifier and regressor wrappers and the conversion helper are now fully typed: `convert_sklearn_to_river` gains overloads, so `classes=None` narrows to `SKL2RiverRegressor` and `classes=[...]` to `SKL2RiverClassifier`; two `Protocol`s name the duck-typed slice of the scikit-learn interface the wrappers actually rely on.

Some big effort was done also to strict type the test suite

## Breaking changes

* Passing `classes=` when converting a regressor now raises `ValueError`; it was silently ignored before.
* `predict_proba_one` returns plain `float` probabilities rather than numpy scalars.

### DIFF
--- a/river/compat/test_sklearn.py
+++ b/river/compat/test_sklearn.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+import typing
+
+import narwhals.stable.v2 as nw
+import numpy as np
 import pandas as pd
 import pytest
 from sklearn import datasets as sk_datasets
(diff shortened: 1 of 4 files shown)


FBruzzesi:
I can't directly tag other reviewers, but I would like to loop @e10e3 in to double check the typing part 🙏🏼

river/compat/test_sklearn.py
```diff
@@ -1,34 +1,76 @@
...
-            linear_model.LogisticRegression(),
-            preprocessing.StandardScaler(),
-            cluster.KMeans(n_clusters=3, seed=42),
-        ]
+        linear_model.LinearRegression(),
+        linear_model.LogisticRegression(),
+        preprocessing.StandardScaler(),
+        cluster.KMeans(n_clusters=3, seed=42),
     ],
+    ids=str,
```
MaxHalford, line 64: Nice I didn't know about this

MaxHalford (approved):

e10e3 (approved):
The type annotations look good to me.

Issue #1331, opened 2023-10-03
score_one method modifies anomaly.LocalOutlierFactor internal state unintentionally?


MarekWadinger:
## Versions

**river version**: development
**Python version**: Python 3.10.1
**Operating system**: macOS Sonoma 14.0 (23A344)

## Describe the bug

Hello 👋

I found out that `score_one` method of `anomaly.LocalOutlierFactor` changes internal state of the detector.

If this is not intentional I'd like to assist in resolving this issue.  I tried to trace back the error and found out that the problem is connected to `expand_objects` function, which creates only a references to original variables. Following modification to lines 349-369 in `anomaly.LocalOutlierFactor` [HERE](https://github.com/online-ml/river/blob/1a32e0d04eecbf79d8977c1a419d5b336d82bc77/river/anomaly/lof.py#L349), helped me to preserve the state over `score_one` calls.

```python
        (
            nm,
            x_list_copy,
            neighborhoods,
            rev_neighborhoods,
            k_dist,
            reach_dist,
            dist_dict,
            local_reach,
            lof,
        ) = expand_objects(
            self.x_scores,
            x_list_copy,
            self.neighborhoods,
            self.rev_neighborhoods,
            self.k_dist,
            self.reach_dist,
            self.dist_dict,
            self.local_reach,
            self.lof,
        )
```
to 
```python
        import copy
        (
            nm,
            x_list_copy,
            neighborhoods,
            rev_neighborhoods,
            k_dist,
            reach_dist,
            dist_dict,
            local_reach,
            lof,
        ) = expand_objects(
            self.x_scores,
            x_list_copy,
            self.neighborhoods.copy(),
            self.rev_neighborhoods.copy(),
            self.k_dist.copy(),
            copy.deepcopy(self.reach_dist),
            copy.deepcopy(self.dist_dict),
            self.local_reach.copy(),
            self.lof.copy(),
        )
```

Please, let me know if that makes sense, and I'd be happy to elaborate on any related issues or comments.

Thank you 🙏

## Steps/code to reproduce

```python
from river import anomaly

lof = anomaly.LocalOutlierFactor()

X = [{"a": 1, "b": 1}, {"a": 1, "b": 1}]
for x in X:
    lof.learn_one(x)

print(
    lof.x_list,
    lof.neighborhoods,
    lof.rev_neighborhoods,
    lof.k_dist,
    lof.reach_dist,
    lof.dist_dict,
    lof.local_reach,
    lof.lof, sep="\n"
[...]

PR #1970, opened 2026-07-30
fix(anomaly): stop PredictiveAnomalyDetection.score_one from mutating the model

JayeshSuryavanshi:
## What

`anomaly.PredictiveAnomalyDetection.score_one` updated the dynamic MAE/variance statistics that back its anomaly threshold. Scoring a point therefore changed the detector: repeated `score_one` calls on the same point returned different values, and scoring without learning silently shifted the threshold. This is the same class of bug fixed for `LocalOutlierFactor` (#1331, #1929), where the convention is that `score_one` must be side-effect free — `PredictiveAnomalyDetection` was the only remaining `anomaly` detector whose `score_one` mutated state.

## Fix

The threshold statistics are now maintained by `learn_one`, which computes the prediction error *before* the predictive model learns each observation. `score_one` only reads them, so it no longer mutates the model. A small private `_predict` helper is shared by both methods.

**Behavior is preserved for the documented score-then-learn loop:** the full 144-score sequence over the docstring's `AirlinePassengers` example is byte-for-byte identical before and after, and the doctest output is unchanged (`0.05329236123455621`).

## Tests

Added `river/anomaly/test_pad.py`: it asserts `score_one` leaves `dynamic_mae` / `dynamic_se_variance` and a full pickle snapshot untouched (so it is idempotent), and that `learn_one` now maintains the threshold statistics. Both tests fail on `main` and pass with this change; the 14 automated estimator checks for `PredictiveAnomalyDetection` still pass. `ruff check`, `ruff format --check`, and `mypy --strict` are clean.


river/anomaly/test_pad.py
```diff
@@ -0,0 +1,52 @@
...
+
+    probe_x, probe_y = {"x": 30.0}, 999.0
+    scores = [model.score_one(probe_x, probe_y) for _ in range(5)]
+
+    # Scoring left every statistic untouched...
+    assert model.dynamic_mae.n == mae_n
+    assert model.dynamic_se_variance.n == var_n
+    assert pickle.dumps(model) == snapshot
+    # ...so it is idempotent.
+    assert len(set(scores)) == 1
```
MaxHalford: Actually here I'd prefer it we added a global test in the `checks` to make sure `score_one` does not mutate any anomaly detector whatsoever

JayeshSuryavanshi: Good call — done in 79e9202. Added `checks.anomaly.check_score_one_does_not_mutate`: it learns a stream, snapshots the model (via `pickle.dumps`), calls `score_one`, and asserts the state is unchanged. It's registered for every anomaly detector — both `AnomalyDetector` and `SupervisedAnomalyDetector` — and passes all of them (`GaussianScorer`, `HalfSpaceTrees`, `LODA`, `LocalOutlierFactor`, `OneClassSVM`, `StandardAbsoluteDeviation`, `PredictiveAnomalyDetection`); reverting the PAD fix makes it fail, so it genuinely guards the whole class going forward.

One structural note: I yield it directly with its own dataset rather than appending it to the generic `dataset_checks`, because supervised anomaly detectors (currently just `PredictiveAnomalyDetection`) aren't wired into those — the common checks call `predict_one`, which `score_one`-based detectors don't implement. Wiring supervised anomaly detectors into the generic checks felt like a bigger, separate change, so I kept this check self-contained; happy to do that as a follow-up PR if you'd like.

I also dropped the now-redundant per-detector mutation test and kept the `learn_one`-maintains-the-threshold-statistics test in `test_pad.py`.

JayeshSuryavanshi:
Hi @MaxHalford, can you please review it when you get time :)

river/checks/__init__.py
```diff
@@ -233,6 +233,19 @@ def yield_checks(model: Estimator) -> typing.Iterator[typing.Callable]:
...
+    # supervised anomaly detectors are not wired into the generic
+    # (predict_one-based) dataset_checks, but their score_one must be
+    # side-effect free too.
+    if isinstance(model, (AnomalyDetector, SupervisedAnomalyDetector)):
+        from river import datasets as _anomaly_datasets
+
+        yield _wrapped_partial(
+            anomaly.check_score_one_does_not_mutate,
+            dataset=list(_anomaly_datasets.CreditCard().take(500)),
+        )
```
MaxHalford: You can simplify here. First of all the import datasets is weird. Just use the one that's already imported. Also no need for the comment, it's too verbose. Finally, you can simplify and run both the ROC AUC test and the new test for all anomaly detectors, including supervised ones.

river/checks/anomaly.py
```diff
@@ -20,3 +20,35 @@ def check_roc_auc(anomaly_detector, dataset):
...
+            anomaly_detector.learn_one(x)
+
+    x, y = first
+    snapshot = pickle.dumps(anomaly_detector)
+    if supervised:
+        anomaly_detector.score_one(x, y)
+    else:
+        anomaly_detector.score_one(x)
+
+    assert pickle.dumps(anomaly_detector) == snapshot, f"score_one mutated {anomaly_detector!r}"
```
MaxHalford: Couldn't you test this at each `learn_one` step? That would be stricter (and therefore a better test)

MaxHalford (commented):
Getting there.

river/checks/__init__.py
JayeshSuryavanshi: Done in 6262189. Both `check_roc_auc` and `check_score_one_does_not_mutate` now run for every anomaly detector, supervised ones included, and I dropped the verbose comment.

On the import: I kept it function-local (like `_yield_datasets` already does) because a module-level `from river import datasets` risks a circular import, but I dropped the alias. I also couldn't fold the two anomaly checks into the generic `dataset_checks` for supervised detectors, since those go through `predict_one`, which `SupervisedAnomalyDetector`s don't implement, so both checks are yielded here over a shared `CreditCard` stream. Happy to restructure if you had a cleaner shape in mind.

river/checks/anomaly.py
JayeshSuryavanshi: Good call, done in 6262189. It now snapshots and asserts no mutation at each step of the stream, interleaved with `learn_one`.

One wrinkle the stricter version surfaced: PAD's predictive-model pipeline lazily memoises `_inference_transformers` / `_last_step_cached` on the first predict, which a naive per-step pickle flags (that is caching, not learning). I prime it with a single warm-up score before the loop so the check targets real state changes. It runs on all 7 anomaly detectors including PAD, and a short prefix over the stream keeps the per-step pickling fast.

PR #1971, opened 2026-07-31
docs: add linear quantile regression recipe

### LINKED ISSUE #1423: Quantile regression
[See scikit-learn](https://scikit-learn.org/stable/modules/linear_model.html#quantile-regression)

We already have `losses.Quantile`.

Siege-Cobby:
Addresses #1423.

This replaces the standalone `QuantileRegressor` implementation with a documentation recipe using River's existing quantile-loss support.

The recipe shows how to perform online linear quantile regression with:

`linear_model.LinearRegression(loss=optim.losses.Quantile(...))`

It demonstrates conditional quantiles at multiple levels on a synthetic stream with known quantiles, and clarifies the distinction between fitted quantiles and calibrated prediction intervals.

It also documents the model-specific tree/leaf quantile approach conceptually: a tree partitions the feature space into leaves, and a quantile-aware leaf can estimate quantiles from the target distribution observed in that region.

Finally, it demonstrates `conf.RegressionJackknife` as River's existing estimator-agnostic interval approach, while noting that its intervals are marginal rather than feature-conditional.

No new estimator or public API is introduced. A generic feature-conditional interval or quantile method remains a separate problem.

### DIFF
--- a/docs/recipes/.pages
+++ b/docs/recipes/.pages
@@ -5,6 +5,7 @@ nav:
     - model-evaluation.md
     - pipelines.md
     - feature-extraction.md
+    - linear-quantile-regression.md
     - saving-loading.md
     - mini-batching.md
     - active-learning.md
--- a/docs/recipes/linear-quantile-regression.ipynb
+++ b/docs/recipes/linear-quantile-regression.ipynb
@@ -4,14 +4,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Linear quantile regression"
+    "# Quantile regression"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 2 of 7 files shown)


Siege-Cobby:
I’ve updated this PR following the discussion in #1423:

https://github.com/online-ml/river/issues/1423#issuecomment-5151352326

The PR is now recipe-only and covers both approaches discussed there:

1. linear quantile regression using River’s existing quantile loss
2. the tree/leaf-based quantile approach as a model-specific alternative

It also demonstrates conf.RegressionJackknife as the current estimator-agnostic interval approach, while clarifying that its intervals are marginal rather than feature-conditional.

No new estimator or public API is introduced here. I’m leaving the generic feature-conditional solution as a separate follow-up problem.

The PR is updated and ready for review.

Siege-Cobby:
Hi @MaxHalford just a gentle follow-up on this PR. I’ve incorporated the earlier feedback and updated it to follow the recipe-first approach discussed in #1423. It should now be ready for review whenever you get a chance. Thanks!

PR #1973, opened 2026-07-31
Fix duplicating API index on the website

e10e3:
Every time the docs were re-generated, the API index was re-created and _appended_ to the previous one, creating needless duplication. The duplication was also visible [on the website](https://riverml.xyz/latest/api/overview/).

The docstring parsing script intended to replace the previous index, but a typo in the regex prevented it from selecting the complete relevant YAML tree. Instead, it only selected the first three lines and replaced them with the new index.

Over time, this created over 7000 duplicate lines in `mkdocs.yml`.

The pattern `(?:    - .*\n)*` matches any number of lines that start with four spaces and a dash. This is a good start, but because the YAML structure is nested, the indent changes starting from the fourth line. So, at the fourth line, the pattern stops matching and does not select anything more for replacement.

Here are the first lines of the relevant part of the file:
```yaml
  - API:
    - api/overview.md
    - active:
      - api/active/EntropySampler.md
```

The fix is to have a flexible indentation. By adding `(?: )*` before the dash, we allow any further level of indentation to ensure we match the nested levels: now we selected all lines with four or more leading spaces and a dash. Because there is a minimum of four spaces of indentation, we can't match outside the targeted tree.

Regression introduced in 95e5fe51.


MaxHalford:
Good catch!

Issue #1389, opened 2023-11-01
Theta forecasting model


MaxHalford:
- https://github.com/Nixtla/statsforecast#theta-family
- https://www.e3s-conferences.org/articles/e3sconf/pdf/2019/10/e3sconf_pe2019_01004.pdf

First question: can it be done online?
Second question: can we reuse seasonality and trend logic we've already implemented for SNARIMAX?

ChidiebereNjoku:
Hi @MaxHalford, I took a look at the two questions you posed for bringing Theta to River:

**1. Can it be done online?**  
Yes, absolutely. The standard Theta model decomposes into a Simple Exponential Smoothing (SES) component and a linear trend component. SES is already an inherently online algorithm, and the OLS trend line can be updated in $O(1)$ time per point using standard running sums ($n$, $\sum t$, $\sum y$, $\sum ty$, and $\sum t^2$).

**2. Can we reuse the seasonality/trend logic from SNARIMAX?**  
Not directly. SNARIMAX uses integer differencing via its `Differencer` to handle trend and stationarity, which is a fundamentally different mechanism than Theta’s explicit trend line decomposition. Structurally, `HoltWinters` is a much closer pattern to follow for managing incremental state components, though we'd use it as a design guide rather than importing directly from it.

**Proposed Scope**  
To keep the initial PR clean and bounded, I’d like to start with the standard unseasonal Theta model (SES + linear drift) first, and leave automatic deseasonalization for a follow-up iteration. 

Does this scope sound right to you before I start building?

MaxHalford:
It sounds right, but I could have figured out all of this myself using Claude/Codex :)

ChidiebereNjoku:
Fair😄.  Just wanted to align on scope before writing code. I'll get started on the PR for unseasonal Theta.

MaxHalford:
No worries. But please be aware of the new [coding agent rules](https://github.com/online-ml/river/blob/main/CONTRIBUTING.md#rules-for-using-coding-agents), and try to avoid copy/pasting LLM output. Please prefer using your own voice. [Don't be a meat proxy](https://gruhn.me/blog/2026-08-03/) :)

ChidiebereNjoku:
Thanks Max, Noted 
 
PR is up (#1977) and all checks are green. Please review whenever you're ready :)

PR #1977, opened 2026-08-04
Add Theta forecasting model (unseasonal)

ChidiebereNjoku:
Adds the unseasonal Theta model to `river.time_series`, following the scope agreed in #1389.

## Approach
The forecast averages two online components:
* **Simple Exponential Smoothing (SES)** level for the flat baseline.
* **OLS linear trend** on $(t, y)$, maintained in $O(1)$ time per point via running sums ($n$, $\sum t$, $\sum y$, $\sum ty$, $\sum t^2$) and extrapolated at half weight.

Averaging the regression line with the SES level yields half the slope per step, matching the drift produced by the original Theta model once its theta-lines are combined *(Assimakopoulos & Nikolopoulos, 2000; Hyndman & Billah, 2003)*.

## Response to Issue Questions
* **Online execution:** Yes, both components update in $O(1)$ per observation.
* **SNARIMAX logic reuse:** No, its `Differencer` relies on a different mechanism. `HoltWinters` served as the structural reference for incremental state management.

## Scope
Seasonality handling is deliberately left out of this PR and will be added in a follow-up once this base model lands.

## Testing
* Module doctests pass locally.
* Existing `time_series` test suite passes.

Addresses #1389


river/time_series/theta.py
```diff
@@ -0,0 +1,76 @@
...
+    ...     model.learn_one(y)
+    >>> [round(f, 3) for f in model.forecast(horizon=3)]
+    [4.613, 5.113, 5.613]
+
+    References
+    ----------
+    Assimakopoulos, V., & Nikolopoulos, K. (2000). The theta model: a
+    decomposition approach to forecasting. International Journal of
+    Forecasting, 16(4), 521-530.
+    """
```
MaxHalford:
```suggestion
    Assimakopoulos, V., & Nikolopoulos, K. (2000). The theta model: a
    decomposition approach to forecasting. International Journal of
    Forecasting, 16(4), 521-530.
    
    """
```

river/time_series/theta.py
```diff
@@ -0,0 +1,76 @@
...
+      original Theta model ends up with once its theta-lines are combined.
+
+    Parameters
+    ----------
+    alpha
+        Smoothing parameter for the SES level.
+
+    Examples
+    --------
+    >>> from river import time_series
```
MaxHalford, line 24:
```suggestion
    Examples
    --------
    
    >>> from river import time_series
```

river/time_series/theta.py
```diff
@@ -0,0 +1,76 @@
...
+    >>> model = time_series.Theta(alpha=0.3)
+    >>> for y in [1, 2, 3, 4, 5]:
+    ...     model.learn_one(y)
+    >>> [round(f, 3) for f in model.forecast(horizon=3)]
+    [4.613, 5.113, 5.613]
```
MaxHalford: How do you know whether this is correct or not? Do you have a reference implementation to compare against?

ChidiebereNjoku:
The doctest was hand-computed at first (SES recursion plus the exact OLS fit on a linear series), which only proves the code matches my math. Cross-checking the method against Hyndman & Billah (2003) actually caught a mistake, my first version averaged the SES level with the full trend line, intercept included, which shifts the forecast. 

Canonical form is SES level plus half the OLS slope as drift, so it's now `level + h * slope / 2` and the doctest is updated.

ChidiebereNjoku:
And to answer your second question, I haven't done a direct numeric comparison against `statsforecast` for this PR, their standard Theta implementation also relies on this exact SES + half-slope equivalence proven by Hyndman. Since my focus was strictly on the unseasonal core we agreed upon ;)

PR #1979, opened 2026-08-05
feat(typing): strict typing stream file readers (`iter_csv`, `iter_arff` and `iter_libsvm`)

FBruzzesi:
# Description

Makes the three file-based stream readers and their shared `open_filepath` helper type-complete, and opts them into strict mypy, with aliases factored out into a new `river.stream.typing` module.

Along the way it fixes a file-handle leak in `iter_libsvm`, which never closed the file it opened because `should_close` was assigned `False` twice.

Adds parametrized tests for all three readers plus `open_filepath`, sharing one 5-case compression matrix (plain / gzip / zip, inferred and explicit) and asserting each reader closes what it opens while leaving caller-supplied buffers alone. **Note**: tests were scaffolded/initially generated by Claude Opus 5.

Finally, added a `[[tool.mypy.overrides]]` entry in pyproject.toml to track files that are type complete. I am sure there are others to add - I simply added what I worked on here and past PRs related to strict typing.

---

Related #1944 

---

Unwanted opinion for reviewing:

* I think the split view is a bit more digestible for a review in this case 😇
* 65%+ of the changes are either tests or non codebase related

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -287,30 +287,3 @@ allow_untyped_calls = true
 allow_incomplete_defs = true
 allow_untyped_defs = true
 warn_return_any = false
-
-[[tool.mypy.overrides]]
-# Opt fully-typed modules back into strict mode, one by one, as part of the incremental typing
-# effort tracked in https://github.com/online-ml/river/issues/1944. mypy applies the most
-# specific pattern, so these bare module names win over the package wildcards above. Keep the
-# list alphabetical, and only add a module once `uv run mypy river` is clean with it listed.
-module = [
-    "river.compact.sklearn_to_river",
-    "river.stream.iter_arff",
-    "river.stream.iter_csv",
-    "river.stream.iter_libsvm",
-    "river.stream.typing",
-    "river.stream.utils",
-    "tests.stream.conftest",
-    "tests.stream.test_iter_arff",
-    "tests.stream.test_iter_csv",
-    "tests.stream.test_iter_libsvm",
-    "tests.stream.test_utils",
-]
-warn_unused_ignores = true
-check_untyped_defs = true
-disallow_subclassing_any = true
-disallow_any_generics = true
-disallow_untyped_calls = true
-disallow_incomplete_defs = true
-disallow_untyped_defs = true
-warn_return_any = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -287,3 +287,30 @@ allow_untyped_calls = true
 allow_incomplete_defs = true
 allow_untyped_defs = true
 warn_return_any = false
+
+[[tool.mypy.overrides]]
+# Opt fully-typed modules back into strict mode, one by one, as part of the incremental typing
+# effort tracked in https://github.com/online-ml/river/issues/1944. mypy applies the most
+# specific pattern, so these bare module names win over the package wildcards above. Keep the
+# list alphabetical, and only add a module once `uv run mypy river` is clean with it listed.
+module = [
+    "river.compact.sklearn_to_river",
+    "river.stream.iter_arff",
+    "river.stream.iter_csv",
+    "river.stream.iter_libsvm",
+    "river.stream.typing",
+    "river.stream.utils",
+    "tests.stream.conftest",
+    "tests.stream.test_iter_arff",
+    "tests.stream.test_iter_csv",
+    "tests.stream.test_iter_libsvm",
+    "tests.stream.test_utils",
+]
+warn_unused_ignores = true
+check_untyped_defs = true
+disallow_subclassing_any = true
+disallow_any_generics = true
+disallow_untyped_calls = true
+disallow_incomplete_defs = true
+disallow_untyped_defs = true
+warn_return_any = true
--- a/river/stream/iter_libsvm.py
+++ b/river/stream/iter_libsvm.py
@@ -1,12 +1,20 @@
 from __future__ import annotations
 
-from river import base
+import os
+import typing
 
-from . import utils
+from river import base
+from river.stream import utils
+
+if typing.TYPE_CHECKING:
+    from river.base.typing import FeatureName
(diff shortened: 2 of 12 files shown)


FBruzzesi:
Gentle ping to @e10e3 and @mugulmd as you are putting a lot of effort in the typing overhaul 🙏🏼

pyproject.toml
```diff
@@ -287,3 +287,30 @@ allow_untyped_calls = true
 allow_incomplete_defs = true
 allow_untyped_defs = true
 warn_return_any = false
+
+[[tool.mypy.overrides]]
```
mugulmd: In the other strict-typing PRs we have opted for a simpler approach: you could just remove the `river.stream.*` from the list above. The end goal is to make that list empty :)

river/stream/typing.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+import os
+import typing
+
+if typing.TYPE_CHECKING:
```
mugulmd: I'm not sure I understand, why do we need to restrict all this to `TYPE_CHECKING` time only?
This is usually done to avoid very specific import problems, but here everything seems pretty standard and simple to me.

river/stream/utils.py
```diff
@@ -1,30 +1,57 @@
...
-        _, ext = os.path.splitext(filepath_or_buffer)
-        compression = {".gz": "gzip", ".zip": "zip"}.get(ext)
+    from river.stream.typing import Compression, FilePath, ResolvedCompression
 
-    def open_zipfile(path):
-        with zipfile.ZipFile(path, "r") as zf:
-            f = zf.open(zf.namelist()[0], "r")
-            f = io.TextIOWrapper(f)
-            return f
+EXT_TO_COMPRESSION: typing.Final[Mapping[str, ResolvedCompression]] = {
```
mugulmd: a good old `dict[str, ResolvedCompression]` would work just fine as well, much more simple :)

river/stream/iter_libsvm.py
```diff
@@ -49,16 +57,12 @@ def iter_libsvm(
     """
 
     # If a file is not opened, then we open it
-    buffer = filepath_or_buffer
-    should_close = False
-    if not hasattr(buffer, "read"):
+    if isinstance(filepath_or_buffer, (str, os.PathLike)):
```
mugulmd, line 60: I *think* you can plainly replace `(str, os.PathLike)` with `FilePath`

pyproject.toml
FBruzzesi: Wouldn't removing it entirely _right now_ make mypy complain for all the other modules in the `stream` that are not type safe/complete?

river/stream/typing.py
FBruzzesi: Yes it is pretty standard and simply, my point was mostly that these objects are not needed at _actual_ runtime anyway. I usually run ruff with most rules activated, and there are a few that would suggest to move these imports under a TYPE_CHECKING block (see [ruff flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc))

river/stream/iter_libsvm.py
FBruzzesi: That's a good use case to move `FilePath` outside the `if TYPE_CHECKING` block 😉

FBruzzesi: I just double checked this and it leads to the following error:

> Second argument to "isinstance" must be a class or tuple of classes
>  Generic type with type arguments not allowed for instance or class checks

river/stream/utils.py
FBruzzesi: Addresses in https://github.com/online-ml/river/pull/1979/commits/815fdfd0a814f28f739f42b92fe7b7b08b69d1b4

pyproject.toml
FBruzzesi: Rollbacked the pyproject edits in https://github.com/online-ml/river/pull/1979/commits/6f4af83aee28633b8b0d288844d6dbff538f6e0a

We will need to keep `river.stream.*` until entire module is type safe

river/stream/typing.py
FBruzzesi: Moved the definitions outside of the TYPE_CHECKING guard in https://github.com/online-ml/river/pull/1979/commits/03a97368b3acded77b76fa6d49a4efde8fd32d90, yet kept the import from other files within such guard

pyproject.toml
mugulmd: Oh my bad I thought you had handled typing for the entire `stream` package 😅 (I'm still a bit of a newbie on this project ahaha)

Would you mind checking how much more work that would be to get there? At first glance it seems like there's only `cache.py` and `tweet_stream.py` left (but I could be wrong, it needs a proper `mypy` run to be sure).

river/stream/iter_libsvm.py
mugulmd: "Generic type with type arguments"? `str` and `os.PathLike` aren't generic, are they? That's definitely a confusing error message 😅

pyproject.toml
FBruzzesi: No worries! Commenting out `river.stream.*` leads to a few more issues: `tweet_stream`, `twitch_chat_stream`, `shuffling`, `iter_sql`, `iter_array`, `cache`, `qa` (and probably `iter_sklearn`, but I have a patch locally already)

river/stream/iter_libsvm.py
FBruzzesi: Yes `os.PathLike` is a generic, in `stream.typing` it's used as `FilePath = str | PathLike[str]`

mugulmd (approved):
LGTM, thanks @FBruzzesi for the changes (and the reactivity!) 🙂 

Since I'm rather new to this project I don't think my approval should be enough to merge though @MaxHalford @smastelini

PR #1980, opened 2026-08-05
detect import cycles with lint-imports + break the 2 module-level cycles

mugulmd:
As mentioned in https://github.com/online-ml/river/issues/1944 we currently have some circular dependencies in the module imports and we would like to break these cycles.

---

### import-linter

We introduce a new dev tool (and add it to the pre-commit config) called [import-linter](https://import-linter.readthedocs.io/en/stable/) to easily detect these import cycles using its `lint-imports` CLI.
This should prevent re-introducing new cycles in the future.

---

### broken cycles

In this PR we fix the 2 main import cycles, however it is important to note that many **function-local imports** are also creating cycles.
Solving all these cycles could quickly become a large refactoring work, which we leave for future (progressive) work.

The 2 cycles that we break here are:
- `anomaly` -> `preprocessing` -> `feature_extraction`
  - we move some of `anomaly`'s code to `tree` where it most likely belong, and make `feature_extraction` import `tree` instead
- `time_series` -> `metrics` -> `anomaly`
  - `metrics` only needs the interface, and this interface lives in the implementation package `anomaly`, so we move the interface up into `river.base`.

---

> [!NOTE]
> *AI disclaimer*: most of the changes in this PR have been written with an LLM-based coding agent.
> I have personally reviewed / corrected these changes before submitting this PR.


river/anomaly/__init__.py
```diff
@@ -25,8 +24,6 @@
 from .svm import OneClassSVM
 
 __all__ = [
-    "base",
-    "AnomalyDetector",
```
MaxHalford, line 29: Good idea. I remember hesitating at the time 😅

pyproject.toml
```diff
@@ -287,3 +288,45 @@ allow_untyped_calls = true
...
+type = "acyclic_siblings"
+ancestors = ["river"]
+# TODO: we ignore all these function-local imports for now
+# as refactoring them would require a lot of work.
+# import-linter has no "ignore function-local" setting
+# so we need to list them manually, which is brittle.
+# Long-term solution would be to prevent function-local imports
+# (enable ruff PLC0415) and solve all the cyclic dependencies
+# they create.
+ignore_imports = [
```
MaxHalford, line 307: I guess a lot of these are due to `_unit_test_params`. I agree it's totally okay to skip them for now. I guess we could think of different ways to separate our code. Good idea to bring this new tool anyhow 👍

MaxHalford (approved):
I like it! Better than before.

mugulmd:
Hey @MaxHalford thanks for the review! Are we waiting for anything specific before merging this PR?

MaxHalford:
No all good! Thanks

PR #1984, opened 2026-08-11
Strict typing: optim sub-package

mugulmd:
## Description

Implement strict typing for the `river.optim` package as it only depends on already strictly typed packages.

Some methods (see `losses.py`) accept `Any` as input but internally check for various specific types at runtime using `isinstance` and then routes differently depending in the detected type: this works fine but doesn't provide much typing info to the user, so we added `@typing.overload`s. Note that this does not alter runtime behaviour.

> [!NOTE]
> *AI disclaimer*: most of the changes in this PR have been written with an LLM-based coding agent.
> I have personally reviewed / corrected these changes before submitting this PR.

### DIFF
--- a/river/optim/ada_bound.py
+++ b/river/optim/ada_bound.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import collections
 import math
 
-from river import optim, utils
+from river import base, optim, utils
 from river.optim.base import DictLike
 
 __all__ = ["AdaBound"]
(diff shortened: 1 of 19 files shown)


river/optim/ada_bound.py
```diff
@@ -54,7 +54,15 @@ class AdaBound(optim.base.Optimizer):
 
     """
 
-    def __init__(self, lr=1e-3, beta_1=0.9, beta_2=0.999, eps=1e-8, gamma=1e-3, final_lr=0.1):
+    def __init__(
+        self,
+        lr: int | float | optim.base.Scheduler = 1e-3,
```
MaxHalford, line 59: Is `int` necessary?

MaxHalford (approved):

river/optim/ada_bound.py
mugulmd: This comes inheriting `Optimizer` defined in `base.py`, which already has this annotation on `main`: https://github.com/online-ml/river/blob/0b8596f930dfc2513c8ede019a8a4d20379fb47c/river/optim/base.py#L76.
Also, `Optimize` uses `int` for an `isinstance` check, and since these typing PRs aren't supposed to change runtime behavior I preferred to leave it that way :)

Issue #1424, opened 2023-11-01
Savitzky-Golay filter


MaxHalford:
See [Wikipedia](https://www.wikiwand.com/en/Savitzky%E2%80%93Golay_filter). The winners of the Kaggle Ashrae competition [used it](https://www.kaggle.com/c/ashrae-energy-prediction/discussion/124709). I checked and it can be computed online. `scipy` [has a function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.savgol_coeffs.html) to obtain the convolution coefficients.

It seems that there is a direct connection between the Savitzky-Golay filter and [LOESS regression](https://www.wikiwand.com/en/Local_regression).

eugene-kamenev:
Found interesting paper on this topic: [Why and How Savitzky–Golay Filters Should Be Replaced](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9026279)
Source codes for filters are available under Supplementary Materials section.

PR #1986, opened 2026-08-12
feat(stats): add Savitzky-Golay smoothing filter

Shizoqua:
Adds `stats.SavitzkyGolay`, a rolling Savitzky-Golay smoothing filter as requested in #1424.

Savitzky-Golay fits a polynomial of a given degree to a sliding window of observations by least squares and returns the smoothed value of the most recent point via the precomputed `scipy.signal.savgol_coeffs` dot product. Because the polynomial is evaluated at the last point of the window, the filter is causal and compatible with online learning.

```python
>>> from river import stats
>>> stat = stats.SavitzkyGolay(window_size=5, polyorder=2)
```

- Coefficients computed once in `__init__`, applied over a fixed-size `deque`
- `get()` returns `None` until the window has enough observations
- `window_size`/`polyorder` follow scipy's constraint (`polyorder < window_size`, enforced by scipy)
- Registered in `stats/__init__.py` and the docs nav
- Release-note entry and CodSpeed benchmark included

Verification:
- `pytest tests/stats` — 175 passed (incl. doctests)
- Matches a manual least-squares polynomial fit to the same window; reproduces a degree-≤2 polynomial exactly
- `ruff check` / `ruff format --check` clean

DCO sign-off included.


river/stats/savitzky_golay.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, window_size: int = 5, polyorder: int = 2) -> None:
+        self.window_size_value = window_size
+        self.polyorder = polyorder
+        self._coeffs = savgol_coeffs(
+            window_length=window_size,
+            polyorder=polyorder,
+            pos=window_size - 1,
+            use="dot",
+        )
```
MaxHalford, line 105: Is this a list or a numpy array? I think it should be the former

MaxHalford (commented):
Overall looks good, but I would like to see a use case so users can understand when to use this

Shizoqua:
Thanks @MaxHalford! I added a use case to the docstring: a noisy temperature probe drifting upward, where smoothing reveals the underlying trend so it can be fed to a model or a threshold. New commit 7121884 includes the example and the doctest output.

e10e3:
A commenter in the issue you reference mentioned the article [_Why and how Savitzky–Golay filters should be replaced_](https://doi.org/10.1021/acsmeasuresciau.1c00054), which argues that one should replace the SG filter with other, more stable and less noisy, filters like sinc or Whittaker–Henderson. Is it not better to implement these alternative filters rather than Savitzky–Golay?

Also, did you read the [Contributing guide](https://github.com/online-ml/river/blob/main/CONTRIBUTING.md#rules-for-using-coding-agents)? Your PR cover letter makes me think you did not, but I could be wrong.

Shizoqua:
Hi @MaxHalford, I've rebased onto the latest `upstream/main` and the branch is now clean with all checks green. The use-case example we discussed (noisy temperature probe trend) is still in the docstring. Could you re-review when you have time?

MaxHalford:
@Shizoqua please answer @e10e3's comment :)

river/stats/savitzky_golay.py
Shizoqua: Good catch, it was a numpy array. It's now stored as a plain list (`.tolist()`) in e765fdfe.

Shizoqua:
Thanks for the pointer, @e10e3. The paper's critique is fair, especially for the causal (endpoint) evaluation used here, which is where SG suppresses noise least. Whittaker–Henderson solves a global penalized least-squares problem over the whole series, so it doesn't map naturally onto River's one-sample `update`. The modified-sinc kernel from the paper is a fixed FIR kernel, though, so it could be implemented the same way as this class. I'd suggest keeping SG as a familiar baseline, with a docstring note linking the paper and its caveats, and I'm happy to follow up with a sinc-kernel smoother in a separate PR if that's useful. @MaxHalford, what do you prefer?

(Also rebased onto the latest `main`; the coefficients are now stored as a list per your earlier comment.)

PR #1993, opened 2026-08-31
feat: strict typing `stream.iter_array`

FBruzzesi:
Scope as per title: add strict typing for `stream.iter_array` module/function.

This is the most tricky typing PR I did so far in river as the runtime behavior allows for _a lot_ of cases.

---

* Related #1944
* Tests cases were mainly generated with Claude Opus 5.0


FBruzzesi:
Gentle ping to @e10e3 and @mugulmd for reviewing 🙏🏼

river/stream/iter_array.py
```diff
@@ -2,20 +2,109 @@
...
 from river import base
 
+if typing.TYPE_CHECKING:
+    from collections.abc import Callable, Iterator, Sequence
 
+Target = typing.TypeVar("Target", bound=base.typing.Target)
+"""The type of a single target value, i.e. of a row of a 1D array of targets."""
+
+Array: typing.TypeAlias = "np.ndarray | Sequence[typing.Any]"
+"""A numpy array or a plain Python sequence."""
```
e10e3, line 18: This comment is redundant.

river/stream/iter_array.py
```diff
@@ -2,20 +2,109 @@
...
+
+Features: typing.TypeAlias = "dict[base.typing.FeatureName, typing.Any]"
+"""One row of features, labeled, which is what an estimator takes as its `x`."""
+
+
+def _passthrough(row: str) -> str:
```
e10e3, line 24: This function is used exactly once. Why not define it at its usage location?

e10e3 (changes requested):
Thank you for the effort, that's indeed a lot of overloads for that function.

I'm surprised by the quantity of changes that you propose. Usually, typing PRs only add types, they aim at not changing the behaviour. If they make changes, it's as little as possible, only to improve the quality of the type annotations.

Hence, why did you need to add tests in a typing PR?
Why did you need to change the whole implementation of iter_array?

FBruzzesi:
Hey @e10e3 thanks for your feedback. I agree with you that I can split this PR into 3 atomic ones:

* Add tests
* Add typing
* Internal refactor; although something might be needed to end up with a type safe return variables

For context, I sometimes find hard to _fully_ understand what a function/method input might be, therefore my process is to ask an LLM to understand the function behavior and generate a test suite with all the cases. Once I have those, it's much easier for me to add the missing type information. That's also the rationale behind the sequence above

river/stream/iter_array.py
FBruzzesi: I tend to avoid defining functions within functions or use lambda functions at all. It's probably more of a personal preference than anything else, so I am ok with whatever style it's preferred here

FBruzzesi:
@e10e3 I will close this one for now and follow up with more atomic PRs. Sorry for the headache and thanks for your feedback 🙏🏼

e10e3:
No worries

PR #1996, opened 2026-09-02
feat: Ranking metrics

sinpe13:
Hello,

PR for the precision@k, recall@k and f1@k metrics ranking metrics. I am new to the codebase and pretty new to open source in general too; so feedback would be welcome before moving into the other ranking metrics (it is all my code as my main goal is learning and contributing, only using AI to ask questions about codebase/my code etc).

Some remarks: 
- https://github.com/online-ml/river/blob/88ca4633f5f5b0db5a8d3c30d91b70fe8eff9933/river/metrics/base.py#L316 has no super init in there, so I needed to create a BinaryRankAtKMetric class to initialize the upstream classes. Maybe worth considering the super init for the MeanMetric class.
- I used the sorted list structure returned by https://github.com/online-ml/river/blob/88ca4633f5f5b0db5a8d3c30d91b70fe8eff9933/river/reco/base.py#L74 as the y_pred format. 

Discussion created for PR as requested by the automatic message: https://github.com/online-ml/river/discussions/1990


MaxHalford (commented):
I love the idea of these metrics, but you have to test them against sklearn. There is an already a whole unit suite that allows testing River metrics against their sklearn equivalent, please hook into it. 

Thanks for working on this!

sinpe13:
> I love the idea of these metrics, but you have to test them against sklearn. There is an already a whole unit suite that allows testing River metrics against their sklearn equivalent, please hook into it.
> 
> Thanks for working on this!

I have seen that it has been done on other metrics. The reason I did not do it is that sklearn does not have the @k version of these metrics so nothing to compare against.

PR #1998, opened 2026-09-03
fix: handle zero vectors in Normalizer 

### LINKED ISSUE #1997: Normalizer.transform_one fails on zero vectors
**river version**: 0.26.1
**Python version**: 3.13
**Apple MacOS**

##  The Bug

This issue is related to `Normalizer.transform_one`.

Currently, `transform_one` works correctly when the input vector contains non-zero values, for example:

x = {"x": 0.3, "y": 0.4}

However, when the input vector contains only zero values:

x = {"x": 0.0, "y": 0.0}

the normalization results in a division by zero and raises:

ZeroDivisionError: float division by zero

Since a zero vector can be a valid input, I think `Normalizer` should handle this case instead of raising an exception.

## Code

normalizer = preprocessing.Normalizer()

x = {"x": 0.0, "y": 0.0}

normalizer.transform_one(x)

## Expected behavior

I would expect the zero vector to be returned unchanged:

{"x": 0.0, "y": 0.0}

rather than raising a ZeroDivisionError.

**File Path** - river/preprocessing/scale.py

I would be happy to submit a PR for this if this behavior is considered a bug.

RoopakBRK:
Fixes #1997 

Fixed the issue related to Normalizer.transform_one

Previously it was giving ZeroDivisionError. This change handles zero vectors safely instead.

File Changed - river/preprocessing/scale.py


MaxHalford:
Can you add an entry to unreleased.md? 🙏

RoopakBRK:
Added the preprocessing.Normalizer fix entry to unreleased.md as requested. Thanks!

Issue #1612, opened 2024-09-09
Global Split Criteria for Decision Trees


danielnowakassis:
In #1610, we noted that at each time a split attempt occurs in Decision Trees, an instance of the Split Criterion Object is created. 

This is not necessary and a globally instantiated Split Criterion Object would enhance the memory usage of LAST.

In river/tree/nodes/last_nodes.py

```python
# change this in future PR's by acessing the tree parameter in the leaf
self.split_criterion = (
    split_criterion  # if None, the change detector will have binary inputs
)
```  

```python
def learn_one(self, x, y, *, w=1, tree=None):
        self.update_stats(y, w)
        if self.is_active():
            if self.split_criterion is None:
                mc_pred = self.prediction(x)
                detector_input = max(mc_pred, key=mc_pred.get) != y
                self.change_detector.update(detector_input)
            else:
                detector_input = self.split_criterion.current_merit(self.stats)
                self.change_detector.update(detector_input)
            self.update_splitters(x, y, w, tree.nominal_attributes)
``` 

would become :

```python
def learn_one(self, x, y, *, w=1, tree=None):
        self.update_stats(y, w)
        if self.is_active():
            if tree.track_error:
                mc_pred = self.prediction(x)
                detector_input = max(mc_pred, key=mc_pred.get) != y
                self.change_detector.update(detector_input)
            else:
                detector_input = tree.current_merit(self.stats)
                self.change_detector.update(detector_input)
            self.update_splitters(x, y, w, tree.nominal_attributes)
```

PR #2000, opened 2026-09-03
refactor(tree): share LAST split criterion

Siege-Cobby:
Addresses #1612.

This refactors `LASTClassifier` so distribution-tracking leaves no longer retain a separate split-criterion instance.

Before this change, when `track_error=False`, each LAST leaf stored its own persistent Gini or Information Gain criterion. A tree with N leaves therefore retained N monitoring criteria.

This PR moves that monitoring criterion to the tree level:

- one lazily initialized criterion is retained per LAST tree
- leaves access the tree-owned criterion through `_change_detector_merit`
- `track_error=True` behavior is unchanged
- `track_error=False` behavior is unchanged
- split evaluation still creates its own temporary criterion, preserving the existing split-attempt lifecycle
- Hellinger restrictions remain unchanged

Tests cover:
- error-tracking detector inputs
- Gini and Information Gain distribution tracking
- reuse of the single tree-owned criterion
- absence of per-leaf `split_criterion`
- deterministic predictions and tree structure
- Hellinger behavior

No public API changes are introduced.

### DIFF
--- a/river/tree/last_classifier.py
+++ b/river/tree/last_classifier.py
@@ -146,6 +146,7 @@ class LASTClassifier(HoeffdingTreeClassifier, base.Classifier):
         )
         self.change_detector = change_detector if change_detector is not None else drift.ADWIN()
         self.track_error = track_error
+        self._change_detector_split_criterion = None
 
         # To keep track of the observed classes
         self.classes: set = set()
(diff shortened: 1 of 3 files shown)


MaxHalford:
Hello @smastelini 👋

Would you mind taking this one?

smastelini:
Hi there! Yeap, I can check it out   😃 

I will review it soon!

smastelini (changes requested):
Hi @Siege-Cobby! Thank you so much for the contribution. 

I agree with the functional changes and they are in agreement with comments left in the original  LAST's PR. Good catch!

I do think some of the tests are overkill. Most of them say more about the underlying drift detector than the tree per se. 

I prefer to keep the ones that really track an important characteristic related to the LAST trees and must be ensured always. 

Moreover, please add an entry in the release notes to help us keep track of the changes :)

PR #2003, opened 2026-09-08
Strict typing: linear_model sub-package

mugulmd:
## Description

Implement strict typing for the `river.linear_model` package as it only depends on already strictly typed packages.

Some `type: ignore` comments were added to handle special cases, for instance when a function relies on another external function with not-precise-enough typing information.

### DIFF
--- a/river/linear_model/softmax.py
+++ b/river/linear_model/softmax.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import copy
 import functools
+import typing
 
 from river import base, optim, utils
 
@@ -61,8 +62,8 @@ class SoftmaxRegression(base.Classifier):
         self,
         optimizer: optim.base.Optimizer | None = None,
         loss: optim.losses.MultiClassLoss | None = None,
-        l2=0,
-    ):
+        l2: float = 0,
+    ) -> None:
         if optimizer is None:
             optimizer = optim.SGD(0.01)
         new_optimizer = functools.partial(copy.deepcopy, optimizer)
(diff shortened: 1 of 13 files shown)


MaxHalford:
> One notable example is BayesianLinearRegression.predict_one which can either output a simple float or a full probability distribution, depending on the with_dist flag, which does not correspond to the common interface of regression models.

Here I think it makes sense to introduce a new `predict_dist_one` method 👍

mugulmd:
> Here I think it makes sense to introduce a new `predict_dist_one` method 👍

Agreed, and done :)

river/linear_model/softmax.py
```diff
@@ -93,7 +96,9 @@ def learn_one(self, x, y):
...
             self.weights[label] = self.optimizers[label].step(w=weights, g=gradient)
 
-    def predict_proba_one(self, x):
+    def predict_proba_one(
+        self, x: dict[base.typing.FeatureName, typing.Any], **kwargs: typing.Any
+    ) -> dict[base.typing.ClfTarget, float]:
         return utils.math.softmax(
             {label: utils.math.dot(weights, x) for label, weights in self.weights.items()}
-        )
+        )  # type: ignore[return-value]
```
MaxHalford, line 104: I assume you'll remove this one later?

MaxHalford (approved):

river/linear_model/softmax.py
mugulmd: Yep the issue is that `utils.math.softmax` returns a `MutableMapping[Any, float]` which doesn't map properly to a `dict[base.typing.ClfTarget, float]`. Fixing this would require a bit of care and uniformity throughout the codebase, so I'm just keeping it in the corner of my mind for now 😅

PR #2007, opened 2026-09-17
Strict typing: neighbors sub-package

mugulmd:
## Description

Implement strict typing for the `river.neighbors` package as it only depends on already strictly typed packages.

### Notes

In `SWINN` the `_data` field was typed as sequence of `Vertex | None` however the `None` value was only used to handle the "vertex replacement" occurring when adding a new vertex on a full sliding window. Using this `None` intermediate value does not bring anything to the table, since the vertex being replaced is being isolated first.
Therefore, in order to simplify the typing of the `SWINN` class and make the code more readable we made the `_data` field a sequence of `Vertex` only.

### Next steps

Currently all items considered for nearest neighbors queries are simply typed as `Any`, which makes things simple considering the potential variety of item types to handle. However using generics instead would provide much more precise typing information. This would mean changing quite a lot in the package's interface, so it is left for later.


river/neighbors/ann/swinn.py
```diff
@@ -341,7 +344,8 @@ def append(self, item: typing.Any, **kwargs: typing.Any) -> None:
 
         # Assign the closest neighbors to the new item
         if len(self) == self.maxlen:
-            neighbors, dists = self._search(node.item, self.graph_k, exclude={node.uuid})
+            # neighbors, dists = self._search(node.item, self.graph_k, exclude={node.uuid})
+            neighbors, dists = self._search(node.item, self.graph_k)
```
MaxHalford, line 348: Isn't this a logic change?

mugulmd: My bad it's leftover code from some experiments, no idea how it ended up there 😅

river/utils/math.py
```diff
@@ -37,6 +37,7 @@
     "softmax",
     "woodbury_matrix",
     "log_sum_2_exp",
+    "_euclidean_distance",
```
MaxHalford: Not a fan of exporting this. Can you suggest an alternative?

MaxHalford (approved):
Approving with one ask. Cheers!

river/utils/math.py
mugulmd: Good spot, this looked quite bad indeed! Found an alternative that's a bit more consistent with the rest of the code :)

PR #2008, opened 2026-09-17
fix(cluster): DenStream offline expansion and micro-cluster keys

### LINKED ISSUE #2006: DenStream: predict_one only expands clusters one hop
## Versions

**river version**: 0.26.1, same on `main`
**Python version**: 3.14.3
**Operating system**: macOS 15

## Describe the bug

In `predict_one` the neighbors of a neighbor are added to the seed queue only if they are already labeled, so the expansion stops after one hop:

https://github.com/online-ml/river/blob/6ff56d7cc7e11936ce827fd1526cc9ee62fb8235/river/cluster/denstream.py#L384

```python
if labels[neighbor_neighbor] is not None:
    seed_queue.append(neighbor_neighbor)
```

It should be `is None`. With three p-micro-clusters A, B, C where A reaches B and B reaches C, DBSCAN should give one cluster. Here it gives two.

This is point 3 of #1555, which I don't think was ever picked up. I ran into it while going through the class for #2005.

## Steps/code to reproduce

Three micro-clusters on a line, 0.25 apart, `epsilon = 0.4`. A-B and B-C are directly density-reachable, A-C is not. The micro-clusters are built with `DenStreamMicroCluster` directly, so only the offline phase runs.

```python
import random
from river import cluster
from river.cluster.denstream import DenStreamMicroCluster

random.seed(0)
ds = cluster.DenStream(decaying_factor=0.01, beta=1.0, mu=1.0005, epsilon=0.4, n_samples_init=3)
ds.initialized = True
ds.timestamp = 0

mcs = []
for cx in (0.0, 0.25, 0.5):
    mc = None
    for _ in range(200):
        p = {0: random.gauss(cx, 0.15), 1: random.gauss(0.0, 0.15)}
        if mc is None:
            mc = DenStreamMicroCluster(x=p, timestamp=0, decaying_factor=ds.decaying_factor)
        else:
            mc.insert(p, 0)
    mcs.append(mc)
ds.p_micro_clusters = {i: mc for i, mc in enumerate(mcs)}

for i, j in [(0, 1), (0, 2), (1, 2)]:
    print("ABC"[i], "-", "ABC"[j], ds._is_directly_density_reachable(mcs[i], mcs[j]))
ds.predict_one({0: 0.0, 1: 0.0})
print("n_clusters =", ds.n_clusters, "(expected 1)")
```

Output:

```
A - B True
A - C False
B - C True
n_clusters = 2 (expected 1)
```

With `is None` on that line it gives 1.

Point 2 of #1555 (new keys from `len(self.p_micro_clusters)` right after a `del`, lines 223 and 231) seems to be still around too. I can put both in a PR if you want.

alfoadd:
Closes #2006.

In `predict_one` a cluster only grew to the direct neighbors of the micro-cluster it started from: the neighbors of a neighbor were queued only when they already had a label, the check was backwards. So two micro-clusters that touch only through a third one in the middle ended up in different clusters.

What I did:

- flipped that `is not None` to `is None`, nothing else in the loop
- a new micro-cluster now gets `max(keys) + 1` as key instead of `len(dict)`, in the three places where that happens (point 2 of #1555). After a deletion `len` could give out a key that was still in use, and the micro-cluster that had it was silently replaced.
- added two tests in `tests/cluster/test_denstream.py`: three micro-clusters in a line end up in one cluster, and a micro-cluster promoted after a deletion doesn't replace another one. Both fail on `main`.
- updated the release notes

Open to changes, as always.


MaxHalford (approved):
Great stuff.

PR #2016, opened 2026-09-21
feat(typing): strict typing for `drift`

nikolas-sapa:
Continues the strict-typing effort (#1944): river.drift.* is removed from the non-strict
mypy overrides, and the whole sub-package now passes under the global strict
configuration.

Adds annotations and return types across ADWIN, KSWIN, PageHinkley, NoDrift,
DummyDriftDetector, DDM, EDDM, FHDDM, HDDMA, HDDMW, SampleInfo,
DriftRetrainingClassifier and the drift.datasets subpackage (21 files). All 547 project
files are mypy-strict clean; 38/38 drift tests, doctests and the docs test pass.

Three items are behavior-relevant and worth explicit review:
1. DriftRetrainingClassifier now passes a bool (was int) as the error indicator, matching
   BinaryDriftDetector.update(x: bool).
2. DDM's running minimums now initialize to float('inf') instead of None — identical
   observed behavior, since the values are always assigned before any read.
3. The type checker surfaced a latent bug in drift/datasets/base.py:
   _annotations_aggregated reads a nonexistent self._annotations attribute (it has no
   callers today, so it would have AttributeError'd on first use), and its intersection
   branch keyed on int 0 where annotator keys are strings. Both fixed — but please
   confirm the intent; if the method is dead code we can drop it instead.

Tests and type annotations were mainly generated with a coding agent (as in #1991);
commit message and this description are mine.

Refs #1944


MaxHalford (approved):
Cheers! I can't spot anything wrong.

mugulmd:
Thanks @nikolas-sapa for participating in this typing effort, `river.drift` was very central and not so easy to tackle!

